Issue #28, opened 2022-07-27
add a show/hide bounding box button


toloudis:
Use Case
volume-viewer supports hiding/showing a bounding box around the volume being displayed. But there is no control for it in the react front end.

Solution
Add a button in the top set of controls before the Channels, next to the hide/show axes button for now.  App component also has a prop so that bounding box can be initialized on or off when the volume is first loaded.

Acceptance criteria
A button that toggles the bounding box on the 3d canvas.
Button should say "Show Bounds" when the box is NOT visible, and "Hide Bounds" when the box IS visible.
App should be initializable with a boolean prop available/testable in the test app in the public/ directory.

frasercl:
Resolved by #33

frasercl:
Resolved by #33

PR #29, opened 2022-07-28
Feature/axis visibility toggle button

### LINKED ISSUE #27: Add a button to show/hide the axis indicator
## Use Case

volume-viewer supports hiding/showing an axis indicator in the corner of the display. But there is no control for it in the react front end.

## Solution

Add a button in the top set of controls before the Channels, on a new row

## Acceptance criteria

A button that toggles the axis indicators on the 3d canvas.
Button should say "Show Axes" when the axes are NOT visible, and "Hide Axes" when the axes ARE visible.

frasercl:
Resolves #27 by:
- Adding a "Show Axes"/"Hide Axes" button to the control panel
- Adding the `viewerConfig.showAxes` prop to `App`
- Modifying some styles in `ControlPanel` to allow a third button row (#28 will likely go here as well)

Also contains dependency updates and addition of `cross-env` package to get `build` script working on Windows.

image


toloudis:
Since this is the first small part of a bigger project, I'm thinking we can create a "ui-overhaul" branch and merge these PRs into that branch as incremental updates so we don't have to worry about accidentally publishing a release with partial changes.   And then since we will have reviewed each of the PRs, at the end we can do a quicker review before merging the whole ui branch into main for a release.  
There may also be some good ways to split the bigger project into releasable parts, which has not been decided yet.   (For example, top toolbar only)

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1077,6 +1091,7 @@ export default class App extends React.Component<AppProps, AppState> {
             mode={userSelections[MODE]}
             imageType={userSelections.imageType}
             autorotate={userSelections[AUTO_ROTATE]}
+            showAxes={userSelections.showAxes}
```
toloudis, line 1094: for naming consistency, maybe this should be `userSelections[SHOW_AXES]` ?

toloudis:
For anyone else reviewing, bear in mind that the layout/CSS is something of a placeholder, as we intend to move this button later.  It was decided that we should just do some basic placement to make the functionality work.   The viewer is unlikely to be released with the button in that location.

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
meganrm: We do both in this block. The convention of using constants like this is a hold over from Javascript I think. With typescript you get better warnings and autocomplete using the dot notation

frasercl: There is nonetheless a `SHOW_AXES` constant used everywhere but here, so for consistency with the rest of this change this looks like a good catch to me.

toloudis: I agree with Megan that we could start removing those string constants as we go, in favor of explicit names knowing that we have typing.

Issue #32, opened 2022-07-28
refactor: get rid of symbolic constants as object property names


toloudis:
## Use Case

A lot of places in the code do:
`userSettings[MY_SETTING]` where `MY_SETTING="mySetting"` as an imported constant.

## Solution

Just use `userSettings.mySetting` and make sure there is a typescript definition for the `userSettings` (which I believe there is)

This is just a refactoring/code maintenance task which might touch enough code that it could stand to be its own PR.

## Acceptance Criteria

no functionality changes / regressions should occur.
Several constants will be removed and code will be smaller and simpler and easier to read.

toloudis:
what remains to be decided is :  there is still a big switch statement that also uses the same symbolic constants.  Consider changing them to use an Enum with integer values but the same symbolic names, or at least collect the values up into their own special container so they are encapsulated and well organized in the code.  We can consider more elegant ways of dealing with the switch statement.

PR #33, opened 2022-07-28
Feature/bounding box toggle button

### LINKED ISSUE #28: add a show/hide bounding box button

frasercl:
Resolves #28
- adds a toggle button to show/hide the bounding box
- adds the prop `showBoundingBox` to the `App` component

Per #32, does NOT add a symbolic constant for property key name (e.g. `SHOW_BOUNDS = "showBounds"`).

image

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -100,7 +100,7 @@ const defaultProps: AppProps = {
   },
   viewerConfig: {
     showAxes: false,
-    showBounds: false,
+    showBoundingBox: false,
     view: "3D", // "XY", "XZ", "YZ"
     mode: "default", // "pathtrace", "maxprojection"
     maskAlpha: ALPHA_MASK_SLIDER_3D_DEFAULT[0],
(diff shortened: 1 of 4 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -960,6 +968,11 @@ export default class App extends React.Component<AppProps, AppState> {
     this.handleChangeToImage(SHOW_AXES, showAxes);
   }
 
+  changeBoundingBoxShowing(showBoundingBox) {
+    this.setUserSelectionsInState({ showBoundingBox });
+    this.handleChangeToImage("showBoundingBox", showBoundingBox);
```
toloudis, line 973: ah, now this part is awkward.  We are using this string only as a key into the switch statement in handleChangeToImage. (which really is more like a map of string to functioncall, which could be done differently than a switch statement).    So the question in my mind is do we just make calls on view3D from here directly, or try to keep it abstracted in some way.   I feel like it would not be so bad to just make the call directly on view3d.

frasercl: Looks like a benefit of `handleChangeToImage` is it puts the check that `view3d` and `image` are defined in one place (line 767) rather than spread wherever a call to `view3d` is made - are there cases in which either would be missing outside startup?

toloudis: I see your point. It's not a huge deal but it doesn't look good when it's a quoted string like that.  After #32 they are all going to become quoted strings in there.  There might be a cleaner way (as part of implementing #32) to make it an object with properties like:
```
{
  showBoundingBox: ()=>view3d.setShowBoundingBox(...),
}
```
That's very bad pseudo code and leaving out important details but at least it doesn't use string literals.

toloudis (approved):

PR #36, opened 2022-07-29
Feature/reset camera button

### LINKED ISSUE #30: Add a button to reset camera
## Use Case

Users unfamiliar with 3d navigation by mouse in a 2d window can sometimes get lost, translate/zoom/rotate to a position that they can't get back from.   This is also a useful convenience even for users who are more experienced.

## Solution

A single command button to put the view back to its initial orientation

## Acceptance Criteria

A button called "Reset Camera" that resets the view to its initial orientation.  (I believe this functionality is already present in the viewer and just needs the button in the UI.). Button will eventually end up in our WIP toolbar design but for now should go above the Channels in the left hand panel.

frasercl:
Adds a button to reset the camera. Resolves #30.

image


toloudis (approved):
LGTM!

PR #41, opened 2022-08-04
Feature/background and bounding box color customization

### LINKED ISSUE #31: Add a color picker to select color for Background and Bounding Box
## Use Case

The viewer provides options to change color of the bounding box and the viewer background but there is no user interface to do so.

## Solution

Add a section after Global Rendering Adjustments called "Customization" with a row for a color picker swatch for Background Color and another row for Bounding Box Color

## Acceptance Criteria

- Color pickers are functional and change color of background and bounding box lines, respectively.  Alpha is not important for either of the two colors so it would be good if the color picker were configured to not show alpha.
- Since bounding box can be turned on/off from top toolbar, in-panel color setting for bounding box should be disabled when bounding box button is off, which text that states, "bounding box turned off".

## Notes

We have a particular color picker library used in other parts of the app which is very configurable.  Look at how the channel colors are changed as a starting point.  It may be possible to use the identical color picker layout but without the alpha part of it.

## Design Specs

frasercl:
Resolves #31:
- Adds a new "Customize" section with color pickers for background and bounding box color, which update dynamically
- Adds the props `backgroundColor` and `boundingBoxColor` to the `App` component
- Bounding box color picker is aware of whether the bounding box is visible, and will display a message if it is not

Modifies the `ColorPicker` component:
- Adds the prop `disableAlpha` to hide the alpha slider (which just passes through to the same prop on `SketchPicker`)
- Because the Customize section is so close to the bottom of its container, color pickers would usually appear off the bottom of the screen and be inaccessible. Therefore, `ColorPicker` is now aware of its position in screen space, and will open above the swatch if it does not have enough space to open below (see second screenshot). This change affects channel color pickers as well.
  - If this feature stays, `ColorPicker` should probably be able to receive a container to measure against, as it is still possible for color pickers to open off the bottom edge of their container if there is technically enough room on screen.

Misc:
- Removes the `reactcss` dependency (still required by `react-color`)
- Bumps `volume-viewer` version to get `view3d.setBackgroundColor` (see allen-cell-animated/volume-viewer#54)
- Changes the type of `channelSettings.color` in `UserSelectionState`, which I believe was mistyped

Screenshots from Cell Feature Explorer:
image
image

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -37,6 +37,8 @@ import {
   SHOW_AXES,
   MAX_PROJECT,
   PATH_TRACE,
+  BACKGROUND_COLOR,
+  BOUNDING_BOX_COLOR,
   LUT_MIN_PERCENTILE,
   LUT_MAX_PERCENTILE,
   COLORIZE_ENABLED,
@@ -54,9 +56,9 @@ import {
   brightnessSliderToImageValue,
   alphaSliderToImageValue,
 } from "../../shared/utils/sliderValuesToImageValues";
+import { colorArrayToFloatArray, rgbObjectToArray } from "../../shared/utils/colorObjectArrayConverting";
 
 import "./styles.css";
-import { colorArrayToFloatArray, rgbObjectToArray } from "../../shared/utils/colorObjectArrayConverting";
 
 const ViewMode = enums.viewMode.mainMapping;
 const { Sider, Content } = Layout;
@@ -104,8 +106,8 @@ const defaultProps: AppProps = {
     showBoundingBox: false,
     view: "3D", // "XY", "XZ", "YZ"
     mode: "default", // "pathtrace", "maxprojection"
-    backgroundColor: [0, 0, 0],
-    boundingBoxColor: [255, 255, 255],
+    backgroundColor: BACKGROUND_COLOR as [number, number, number],
+    boundingBoxColor: BOUNDING_BOX_COLOR as [number, number, number],
     maskAlpha: ALPHA_MASK_SLIDER_3D_DEFAULT[0],
     brightness: BRIGHTNESS_SLIDER_LEVEL_DEFAULT[0],
     density: DENSITY_SLIDER_LEVEL_DEFAULT[0],
@@ -181,8 +183,8 @@ export default class App extends React.Component<AppProps, AppState> {
         [AUTO_ROTATE]: false,
         [SHOW_AXES]: props.viewerConfig.showAxes,
         showBoundingBox: props.viewerConfig.showBoundingBox,
-        boundingBoxColor: props.viewerConfig.boundingBoxColor || [255, 255, 255],
-        backgroundColor: props.viewerConfig.backgroundColor || [0, 0, 0],
+        boundingBoxColor: props.viewerConfig.boundingBoxColor || BOUNDING_BOX_COLOR,
+        backgroundColor: props.viewerConfig.backgroundColor || BACKGROUND_COLOR,
         [MAX_PROJECT]: maxproject,
         [PATH_TRACE]: pathtrace,
         [ALPHA_MASK_SLIDER_LEVEL]: [props.viewerConfig.maskAlpha] || ALPHA_MASK_SLIDER_3D_DEFAULT,
@@ -280,6 +282,10 @@ export default class App extends React.Component<AppProps, AppState> {
   }
 
   onView3DCreated(view3d) {
+    const { userSelections } = this.state;
+    view3d.setBackgroundColor(colorArrayToFloatArray(userSelections.backgroundColor));
+    view3d.setShowAxis(userSelections[SHOW_AXES]);
+
     this.setState({ view3d });
   }
 
@@ -443,10 +449,8 @@ export default class App extends React.Component<AppProps, AppState> {
     view3d.setGamma(aimg, imageValues.min, imageValues.scale, imageValues.max);
     // update current camera mode to make sure the image gets the update
     view3d.setCameraMode(enums.viewMode.VIEW_MODE_ENUM_TO_LABEL_MAP.get(userSelections.mode));
-    view3d.setShowAxis(userSelections[SHOW_AXES]);
     view3d.setShowBoundingBox(aimg, userSelections.showBoundingBox);
     view3d.setBoundingBoxColor(aimg, colorArrayToFloatArray(userSelections.boundingBoxColor));
-    view3d.setBackgroundColor(colorArrayToFloatArray(userSelections.backgroundColor));
     // tell view that things have changed for this image
     view3d.updateActiveChannels(aimg);
   }
@@ -726,10 +730,8 @@ export default class App extends React.Component<AppProps, AppState> {
     view3d.setGamma(aimg, imageValues.min, imageValues.scale, imageValues.max);
     // update current camera mode to make sure the image gets the update
     view3d.setCameraMode(enums.viewMode.VIEW_MODE_ENUM_TO_LABEL_MAP.get(userSelections.mode));
-    view3d.setShowAxis(userSelections[SHOW_AXES]);
     view3d.setShowBoundingBox(aimg, userSelections.showBoundingBox);
-    view3d.setBackgroundColor(colorArrayToFloatArray(userSelections.backgroundColor));
-    view3d.setBoundingBoxColor(colorArrayToFloatArray(userSelections.boundingBoxColor));
+    view3d.setBoundingBoxColor(aimg, colorArrayToFloatArray(userSelections.boundingBoxColor));
     // tell view that things have changed for this image
     view3d.updateActiveChannels(aimg);
 
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -231,6 +231,8 @@ export default class App extends React.Component<AppProps, AppState> {
     this.changeAxisShowing = this.changeAxisShowing.bind(this);
     this.changeBoundingBoxShowing = this.changeBoundingBoxShowing.bind(this);
     this.resetCamera = this.resetCamera.bind(this);
+    this.changeBackgroundColor = this.changeBackgroundColor.bind(this);
+    this.changeBoundingBoxColor = this.changeBoundingBoxColor.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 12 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -56,6 +56,7 @@ import {
 } from "../../shared/utils/sliderValuesToImageValues";
 
 import "./styles.css";
+import { colorArrayToFloatArray, rgbObjectToArray } from "../../shared/utils/colorObjectArrayConverting";
```
toloudis, line 59: move this up to be next to the other utils import

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -438,6 +445,8 @@ export default class App extends React.Component<AppProps, AppState> {
     view3d.setCameraMode(enums.viewMode.VIEW_MODE_ENUM_TO_LABEL_MAP.get(userSelections.mode));
     view3d.setShowAxis(userSelections[SHOW_AXES]);
     view3d.setShowBoundingBox(aimg, userSelections.showBoundingBox);
+    view3d.setBoundingBoxColor(aimg, colorArrayToFloatArray(userSelections.boundingBoxColor));
+    view3d.setBackgroundColor(colorArrayToFloatArray(userSelections.backgroundColor));
```
toloudis, line 449: since this (bg color) doesn't depend on aimg, it probably should not be called here.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -719,6 +728,8 @@ export default class App extends React.Component<AppProps, AppState> {
     view3d.setCameraMode(enums.viewMode.VIEW_MODE_ENUM_TO_LABEL_MAP.get(userSelections.mode));
     view3d.setShowAxis(userSelections[SHOW_AXES]);
     view3d.setShowBoundingBox(aimg, userSelections.showBoundingBox);
+    view3d.setBackgroundColor(colorArrayToFloatArray(userSelections.backgroundColor));
+    view3d.setBoundingBoxColor(colorArrayToFloatArray(userSelections.boundingBoxColor));
```
toloudis, line 732: setBoundingBoxColor needs to pass in aimg ?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -719,6 +728,8 @@ export default class App extends React.Component<AppProps, AppState> {
     view3d.setCameraMode(enums.viewMode.VIEW_MODE_ENUM_TO_LABEL_MAP.get(userSelections.mode));
     view3d.setShowAxis(userSelections[SHOW_AXES]);
     view3d.setShowBoundingBox(aimg, userSelections.showBoundingBox);
+    view3d.setBackgroundColor(colorArrayToFloatArray(userSelections.backgroundColor));
```
toloudis, line 731: same comment as above, bg color probably should be initialized prior to image load time

src/aics-image-viewer/components/ColorPicker.jsx
```diff
@@ -58,45 +68,55 @@ class ColorPicker extends React.Component {
...
-          zIndex: '9999',
-        },
-        cover: {
-          position: 'fixed',
-          top: '0px',
-          right: '0px',
-          bottom: '0px',
-          left: '0px'
-        }
+    let styles = {
```
toloudis, line 71: should still be const?

src/aics-image-viewer/components/App/index.tsx
toloudis: suggestion: initialize bg color in CellViewerCanvasWrapper/index.jsx, in componentWillReceiveProps (pass bg color in as a prop to this component).   It may also need to be set in componentDidMount also...

src/aics-image-viewer/components/App/index.tsx
frasercl: looks like just about all the setters in volume-viewer with a `volume` argument (the ones that require `aimg`) leave that argument unused and instead check `this.image`, potentially to future-proof against the viewer eventually supporting multiple images at a time? if so, `setBoundingBoxColor` will need to receive a volume to know which of multiple bounding boxes to recolor.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -103,6 +104,8 @@ const defaultProps: AppProps = {
     showBoundingBox: false,
     view: "3D", // "XY", "XZ", "YZ"
     mode: "default", // "pathtrace", "maxprojection"
+    backgroundColor: [0, 0, 0],
+    boundingBoxColor: [255, 255, 255],
```
meganrm, line 108: These seem like they should be in constants so you can have a default you use other places

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -178,6 +181,8 @@ export default class App extends React.Component<AppProps, AppState> {
         [AUTO_ROTATE]: false,
         [SHOW_AXES]: props.viewerConfig.showAxes,
         showBoundingBox: props.viewerConfig.showBoundingBox,
+        boundingBoxColor: props.viewerConfig.boundingBoxColor || [255, 255, 255],
+        backgroundColor: props.viewerConfig.backgroundColor || [0, 0, 0],
```
meganrm, line 185: Here's where you could use those default constants

src/aics-image-viewer/components/ControlPanel/index.jsx
```diff
@@ -223,11 +224,11 @@ export default class ControlPanel extends React.Component {
...
-          title={[
-            this.renderAxesButton(),
-            this.renderBoundingBoxButton(),
-            this.renderResetCameraButton()
-          ]}
+          title={<>
+            {this.renderAxesButton()}
+            {this.renderBoundingBoxButton()}
+            {this.renderResetCameraButton()}
+          </>}
```
meganrm, line 231: what's the reason for this change?

src/aics-image-viewer/components/App/index.tsx
toloudis: yes, setBoundingBoxColor should get an aimg passed in for sure.

lynwilhelm:
@frasercl  I would consider these couple things:
- Investigate if the bottom line in panel can be removed
- See if the 1px stroke around the color rectangles can be implemented globally without too much effort

Great work:)

src/aics-image-viewer/components/ControlPanel/index.jsx
frasercl: React was complaining that this array of components did not have indexes; fragments don't have that requirement. I'm pretty sure this isn't the best practice (I believe fragments are meant to be used pretty much exclusively as the top-level container in JSX, to allow a component's `render` function to return multiple elements) but these buttons are soon to be moved out of the control panel anyways and into the new toolbar.

src/aics-image-viewer/components/App/index.tsx
frasercl: In this case `setShowAxis` should probably move as well. Looks like everything else that's here and doesn't accept `aimg` as an argument still internally depends on the image having been initialized (looking at `updateExposure`, `setCameraMode`).

toloudis: basically agree, but my thinking was that the axes wont be indicative of much without an image there

frasercl: makes sense, though in an editor configured to display axes by default, having the axes appear with no image at least indicates that the viewer has loaded correctly and any problem is with image loading.

src/aics-image-viewer/components/ColorPicker.jsx
frasercl: was adding a property to the style based on whether the color picker should open above or below swatch, which I agree was weird. just changed to be more consistent with other single-file components, with an all-caps `const STYLES` at the very bottom.

toloudis: Adding a property shouldn't have changed constness.   Would be surprised if you got some tool-based indication that you had to change it.  Unless you were reassigning the `styles` variable itself.

frasercl: I was conditionally adding either one property or another after declaration. You're right that with the way const works in JS it wouldn't have mattered, but in personal projects I tend to change the keyword to indicate whether any property of an element changes, and I hadn't reconsidered that habit until just now. Regardless, style-wise I feel better about the changes I just made.

src/aics-image-viewer/components/ColorPicker.jsx
```diff
@@ -58,48 +68,56 @@ class ColorPicker extends React.Component {
...
-          position: 'fixed',
-          top: '0px',
-          right: '0px',
-          bottom: '0px',
-          left: '0px'
-        }
-      }
-    });
+
+    let popoverDirectionStyle = {};
```
toloudis, line 72: const. 
could even be 
```
const popoverDirectionStyle = this.state.openAboveSwatch ? {bottom:"25px"} : {top:"1px"};
```
I suspect, because of the single quotes around the px values, that you need to run `npm run lint`, `npm run typeCheck`, and run prettier (formatter) to do the basic formatting checks.
We should probably see if there's a way to ensure autoformatting on commit or something.  I have "format on save" enabled for the project in vscode settings.

toloudis: Actually I see that the gh-actions did run the lint but I'm surprised it didn't warn about let vs const. I guess volume-viewer warns about this but website-3d-cell-viewer doesn't, in the eslint config ( "prefer-const": ["warn"] ).
I would advocate for adding this but in a separate maintenance pr because it might trigger a whole bunch of warnings.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -103,6 +106,8 @@ const defaultProps: AppProps = {
     showBoundingBox: false,
     view: "3D", // "XY", "XZ", "YZ"
     mode: "default", // "pathtrace", "maxprojection"
+    backgroundColor: BACKGROUND_COLOR as [number, number, number],
+    boundingBoxColor: BOUNDING_BOX_COLOR as [number, number, number],
```
toloudis, line 110: i suggest moving the [number,number,number] typing over to where the variable is defined rather than here.  you might have to rename the file to be a ts rather than a js file?

toloudis: we can be gradually walking this codebase from js to ts overall

toloudis (approved):
Looking good.  I feel like the bg color should be passed to the CellViewerCanvasWrapper as a prop but it's not a deal breaker to do it this way at all.
* The typescript [number,number,number] is the only thing I would still change to make this code cleaner.

src/aics-image-viewer/components/ColorPicker.jsx
frasercl: Yep, your way is better on this. Don't know how I missed it in the first place. And eslint throws up ~50 warnings when told to care about const, so definitely an issue for another time.

PR #52, opened 2022-08-11
Feature/autorotate prop

### LINKED ISSUE #35: autorotate mode on startup
## Use Case

we want to be able to initialize the viewer having the image rotating about the vertical axis

## Solution

Move the "autorotate" or "turntable" toggle button out of the Options area and into the toolbar where the other buttons are.
Add a prop to the app to let it start up in turntable mode (default to false)

frasercl:
Resolves #35 by adding the `viewerConfig.autorotate` prop and moving the autorotate button to a new (extremely temporary) home in the control panel.

Also:
 - replaces some (not all) legacy component lifecycles that cause warnings in development mode, per [this post by the React devs](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html).
 - contains a bunch of formatting changes since `formatOnSave` is on now.

Screenshot:
image


src/aics-image-viewer/components/CellViewerCanvasWrapper/index.jsx
```diff
@@ -1,74 +1,72 @@
...
+    if (!this.view3D) {
+      this.view3D = new View3d(this.view3dviewerRef.current);
+      this.props.onView3DCreated(this.view3D);
+      this.view3D.setAutoRotate(this.props.autorotate);
+    }
+  }
+
+  componentDidUpdate(prevProps, _prevState) {
+    if (this.view3D && prevProps.mode && prevProps.mode !== this.props.mode) {
+      this.view3D.setCameraMode(viewMode.VIEW_MODE_ENUM_TO_LABEL_MAP.get(this.props.mode));
```
frasercl, line 34: Worth talking generally about what component should be responsible for propagating which changes to `view3d` in this PR, since I chose not to move `backgroundColor` into this component in #41, piggybacked off of `autorotate` already being in this component here, and having more clarity on this will be useful in #32.

I'm a bit concerned about the added complexity of letting this component be responsible for calls to `view3d` that don't require access to `image`. In order to keep this component fully responsible for calling `setAutoRotate`, I put the initial call in `componentDidMount` here (potentially a non-obvious place for it). The other method that this component appears to manage, `setCameraMode`, is not initially called here, but in the base `App` component, and is currently being called twice every time `ControlPanel` calls the update handler: once by `App` in `handleChangeToImage`, then a second time here when this component receives new props. I think it makes some sense to move responsibility for calling `view3d` methods to handlers in `App` to mitigate similar double-call issues in the future and reduce complexity a bit, even if it doesn't make the most semantic sense.

toloudis: I see your point.  I can see it both ways.  Imagine if this React component "CellViewerCanvasWrapper" were coming from the `volume-viewer` library.    Shouldn't it have props including `backgroundColor` and `autoRotate`?
On the other hand, if the CellViewerCanvasWrapper component exposes the view3d, then it doesn't need any other means of accessing the volume-viewer functionality.
I suppose right now I would slightly lean to agreeing with you -- that the view3d is already being cached up at the app level and all calls made in it from that layer.

toloudis (approved):
General comment:  code reformatting adds too much noise to PRs and makes them harder to read.  But code formatting is important.  Maybe a "format everything" PR that doesn't change any functionality...
Also, we should be gradually transitioning this whole codebase to 100% Typescript.   I'm not sure whether that should have the same philosophy: all or nothing, or piecemeal folded in with other code changes?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.jsx
frasercl: I definitely see tradeoffs for doing it both ways - it's very nice at the app level to be able to just call a state change and have `CellViewerCanvasWrapper` propagate that to `view3d` on its own. I will add as well that I spent a while debugging some strange behavior after modifying `App` to turn off `autorotate` in this manner when switching to 2d view modes, which turned out to be due to the fact that `view3d` maintains separate control states for 2d and 3d modes, and `componentDidUpdate` only gets called to sync that change after `App` has already changed modes and `view3d` has already swapped control states. So while convenient, I'd still say updating `view3d` in component lifecycles isn't quite worth the footguns.

Issue #42, opened 2022-08-05
update antd version


toloudis:
## Use Case

We are at least 1 major version behind antd and the toolchain is starting to pass it by

## Solution

Update antd, fix everything that breaks.  Make sure it doesn't cause a chain reaction of having to update fundamental things like react etc.

toloudis:
given some of what we have learned about antd, this may have to be rethought - embedding components using antd version M in a version N app is probably a disaster.

toloudis:
antd 5 has some intriguing developments ("no more global styles"?) which might be very worthwhile.  TODO: nbvv global style hack might be removable, test with multiple viewer instances in one page, etc.

PR #54, opened 2022-08-22
Feature/toolbar

### LINKED ISSUE #37: create toolbar component
## Use Case

We want to move several of the buttons at the top of the control panel into a horizontal toolbar that can be placed at the top of the 3d canvas.

## Solution

- Create a React component that renders the complete set of buttons per @lynwilhelm design spec. 
- We can use placeholder icons for those buttons that require icons.
- Some buttons will need to convert from radio button groups to dropdowns.
- We want the toolbar to float over the 3d canvas (same relative "top")  but it should be its own component so that if we decided to, we could do layout such that it is vertically above the canvas.

## Design Specs
Follow specs in this design XD file for all styling guidance

frasercl:
Resolves #37 by moving viewer controls out of `ControlPanel` and into a new `Toolbar` component centered within the viewport along the top edge.

Also:
 - Adds a new optional prop `onControlPanelToggle` to the base `App` component (helps CFE keep viewer title centered; see below)
 - Makes control panel responsive: the control panel will close if the window loads or resizes below a threshold width (currently 900px)
 - Removes a couple more [legacy component lifecycles](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html).
 - Much lighter-weight `ControlPanel` component is now a function component; old `AutoRotateButton` component is removed.

image

Note in this screenshot:
1. Most icons are placeholders pending final designs from Thao
2. This screenshot includes changes currently on the `feature/toolbar` branch of Cell Feature Explorer which moves the viewer title out of the space now occupied by the toolbar into a centered position in the unused space above the viewer, which will be PR'ed once these changes are finalized


package.json
```diff
@@ -59,6 +59,8 @@
     "@babel/preset-typescript": "^7.14.5",
     "@babel/runtime": "^7.14.6",
     "@types/jest": "^27.4.0",
+    "@types/react": "17.0.2",
+    "@types/react-dom": "17.0.2",
```
toloudis, line 63: minor concern that these are pinned at version 17 ( and no `^` )but we are still on react 16?  I am assuming that nothing's wrong here but sometimes these things need to be in sync.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -947,7 +961,7 @@ export default class App extends React.Component<AppProps, AppState> {
       };
       // if path trace was enabled in 3D turn it off when switching to 2D.
       if (userSelections[PATH_TRACE]) {
-        this.changeRenderingAlgorithm("volume");
+        this.onChangeRenderingAlgorithm("volume");
```
toloudis, line 964: There must be a symbolic constant for `"volume"` here (I hope).  I see PATH_TRACE and MAX_PROJECT used elsewhere.

toloudis: An enum would be an appropriate way to logically encapsulate all 3 modes as a type, too.

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+    autoRotateButton: boolean;
+    viewModeRadioButtons: boolean;
+    fovCellSwitchControls: boolean;
+    resetCameraButton: boolean;
+    showAxesButton: boolean;
+    showBoundingBoxButton: boolean;
+  };
+}
+
+function DownloadButton({ fovDownloadHref, cellDownloadHref, hasFov }: DownloadButtonProps) {
```
toloudis, line 56: I suggest moving DownloadButton into its own file within this directory. I know it's very short but it's sort of right on the edge of worth moving out.  Also it's good to aim for one component per file

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+        </Menu.Item>
+      </Menu>
+    );
+    return (
+      <Dropdown overlay={menu} placement="bottomRight" trigger={["click"]}>
+        <Button className="btn-borderless" icon="download" />
+      </Dropdown>
+    );
+  } else {
+    return <Button className="btn-borderless" icon="download" href={cellDownloadHref} />;
```
toloudis, line 78: maybe if fov href and cell href are both empty, we don't render a download button?

toloudis (approved):

package.json
frasercl: Pinned per [this stackoverflow response](https://stackoverflow.com/questions/71791347/npm-package-cannot-be-used-as-a-jsx-component-type-errors/71828113#71828113). I was experiencing similar type errors despite making no changes, apparently as a result of dependencies changing. (potentially better to add these in `resolutions` as described?) If I recall correctly, picking a version that started with 16 was giving me different errors. May be worth revisiting regardless, potentially as part of #42.

toloudis: interesting, they also talk about node/npm versions in the discussion there.   If nothing is broken downstream then I'm happy but it looks weird.  I wish we could put comments in json files.

frasercl: worked for me just now to match types version with react version, including ^ - perhaps I got package-lock in a state where it will hold down the correct version?

PR #55, opened 2022-08-24
Feature/toolbar, color customization, added viewer controls

### LINKED ISSUE #27: Add a button to show/hide the axis indicator

### LINKED ISSUE #28: add a show/hide bounding box button

### LINKED ISSUE #30: Add a button to reset camera

### LINKED ISSUE #31: Add a color picker to select color for Background and Bounding Box

### LINKED ISSUE #35: autorotate mode on startup

### LINKED ISSUE #37: create toolbar component

frasercl:
Contains the cumulative changes of #29, #33, #36, #41, #52, and #54.

Highlights:
- Adds color customization section to control panel, for picking the colors of the background and bounding box
- Adds new controls to show/hide axes and bounding box and reset the camera
- Moves these and other viewer controls out of the control panel and into a new toolbar component
- Adds new props to `App` for initial settings of these new controls, render config for the toolbar, etc.

Resolves #27, resolves #28, resolves #30, resolves #31, resolves #35, resolves #37.


toloudis (approved):
LGTM

Issue #57, opened 2022-08-24
full conversion to typescript


toloudis:
## Use Case

Typescript is the preferred js flavor at AICS, for many reasons.  The code is partially converted but linting has been relaxed due to many many functions missing return types and argument types.

## Solution

Convert all file extensions from js,jsx to ts, tsx.  Add linter rules to warn or error on public methods that don't have argument types or return types.  Suggestion: use simularium-viewer linting rules for consistency. 

Then fix all errors.  This may require some real code changes if types have been abused too much.  Possibly can be done in multiple passes, converting only a few files at a time, for review-ability.

PR #58, opened 2022-08-26
Feature/dry state on init

toloudis:
While debugging some code, I made a few DRY refactors that consolidate repeated code in a couple of places, and removed an unused function and an unused prop.

This is hopefully a net positive on readability, even if the diff looks funky.  It also consolidates a couple of places where setState was being called consecutively... there are a lot more of those, unfortunately.

A lot of what happens in the top level code here is that the react state changes also need to make calls into a pure JS api on this "view3d" object, and keep the ui consistent with what's internal to the view3d.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -451,14 +451,7 @@ export default class App extends React.Component<AppProps, AppState> {
     const { userSelections, view3d } = this.state;
     const { viewerConfig } = this.props;
     const channelSetting = newChannelSettings || userSelections[CHANNEL_SETTINGS];
-    let alphaLevel =
-      userSelections.imageType === SEGMENTED_CELL && userSelections.mode === ViewMode.threeD
-        ? ALPHA_MASK_SLIDER_3D_DEFAULT
-        : ALPHA_MASK_SLIDER_2D_DEFAULT;
-    // if maskAlpha is defined in viewerConfig then it will override the above
-    if (viewerConfig.maskAlpha !== undefined) {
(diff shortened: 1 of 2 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -483,19 +468,13 @@ export default class App extends React.Component<AppProps, AppState> {
...
-    );
-    let imageDensity = densitySliderToImageValue(userSelections[DENSITY_SLIDER_LEVEL], userSelections[PATH_TRACE]);
-    let imageValues = gammaSliderToImageValues(userSelections[LEVELS_SLIDER]);
-    // set alpha slider first time image is loaded to something that makes sense
-    this.setUserSelectionsInState({ [ALPHA_MASK_SLIDER_LEVEL]: alphaLevel });
-
-    // Here is where we officially hand the image to the volume-viewer
-
+  // set up the Volume into the Viewer using the current initial settings
+  private placeImageInViewer(aimg:Volume, newChannelSettings?):void {
```
frasercl, line 475: Add spaces after types, and looks like the same definitions could be applied to `initializeNewImage` as well if interested. (Type for `newChannelSettings` will certainly have to wait for #57 though, per our earlier conversation.)

frasercl (approved):
Always a fan of not repeating yourself. LGTM

src/aics-image-viewer/components/App/index.tsx
toloudis: i thought this would have been autoformatted better. I'll fix the whitespace and add Volume as a type everywhere I can.

PR #62, opened 2022-09-07
Feature/style organization

frasercl:
Fixes loading problems that caused the test 3d viewer to appear unstyled relative to Cell Feature Explorer, and moves styles from CFE that only impacted the viewer into the viewer itself.

- Configures styles/ant-vars.less to load correctly. There is an identical file in CFE which defines the app's color scheme. Getting this file to load applied the same theming.
  - NOTE: while this does not enable dynamic theming (e.g. light/dark modes), it at least implies a path if we're interested in implementing it. Ideally, ant would use native CSS vars for this, but it doesn't by default, and getting it to is [a bit tricky](https://ant.design/docs/react/customize-theme-variable).
- Configures local font files (for Merriweather Sans and Overpass) to load correctly.
  - NOTE: this does not answer the question of whether those files *should* be there. Most of these fonts are overridden by Ant's font choice, and *all* are overridden in CFE. If we want to keep these font choices, or apply new static fonts, they can be updated in ant-vars or by styling. If we want to default to Ant's choices, which are reasonable but rely on system fonts and are thus inconsistent system-to-system, then all typography-related files can be removed from this repo.
- Adds styles applied to the viewer by CFE to the viewer itself, e.g. to set the background color of the ROI slider drawer and to make the open/close control panel toggle visible.
   - NOTE: the feature/style-organization branch of CFE contains changes that remove these styles from CFE. That branch should be merged if these changes are published in a new version of this package.
- Renames axis-style-override.css to noui-slider-override.css, to be more descriptive of that file's purpose.

Screenshot (note the font on channel titles, e.g. "Membrane," "Labeled structure;" and on the title "Region of interest"):
image


toloudis (approved):
looks good. thanks for fixing this!

Issue #49, opened 2022-08-11
ROI: 2d modes should be single-slice


toloudis:
## Use Case

Subtask of #34
In 2D projection modes, the range slider is not as useful as seeing single slices at a time, and sliders for clipping axes other than the one being looked down are not useful at all (and are currently frozen in this mode anyways).  The UI will be simplified if we can have only one slider with a single slice.

## Solution

2d modes only show single sliders with a single handle

PR #64, opened 2022-09-07
Feature/ROI sliders: single-slice in 2d mode

frasercl:
Finishes resolving #49 by making axis sliders have only one handle in 2D modes.

Given that slider state is stored internally as `[number, number]`, this is accomplished like this:
1. in the `Nouislider` component, if we're in 2d mode, pass only the first number from state as a handle position, not both
2. in event handlers, if we get only a single-element array of values back, treat this element as both the start and end of the slice and store that value in both state fields

Screenshot:
image

### DIFF
--- a/src/aics-image-viewer/components/AxisClipSliders/index.tsx
+++ b/src/aics-image-viewer/components/AxisClipSliders/index.tsx
@@ -177,6 +177,8 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
         const thicknessReduce = isActiveAxis ? step - stepEpsilon : 0.0;
 
         const start = values[0] / max - 0.5;
+        // values.length is the number of handles on this slider:
+        // either one handle (2d mode), or a range with 2 handles (3d mode)
         const end = values[values.length - 1] / max - 0.5 - thicknessReduce / max;
 
         // get a value from -0.5..0.5
@@ -187,6 +189,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
 
   // When user finishes moving the active slider, update slice label
   makeSliderSetFn(axis: string) {
+    // Values may be of length 1 or 2 (see above, in makeSliderUpdateFn); ensure we pass 2 values regardless
     return (values: number[]) => this.setSliderState(axis, [values[0], values[values.length - 1]]);
   }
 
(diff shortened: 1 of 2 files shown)


meganrm (approved):

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -176,7 +177,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
         const thicknessReduce = isActiveAxis ? step - stepEpsilon : 0.0;
 
         const start = values[0] / max - 0.5;
-        const end = values[1] / max - 0.5 - thicknessReduce / max;
+        const end = values[values.length - 1] / max - 0.5 - thicknessReduce / max;
```
toloudis, line 180: i might add a comment here to say that this is meant to handle both cases when values is of length 1 or 2

toloudis: ```// values will be of length 1 when slider has a single handle, and of length 2 when slider has 2 handles. ```

toloudis: there will probably be a future modification to this code to handle true slice thickness (i.e. a single slice should have nonzero max-min) as we discussed in slack thread, and maybe treat thicknessReduce as some kind of minimum clamping if the max-min thickness is too small?  half baked idea but we want to make this as accurate as possible for single-slice viewing

toloudis (approved):

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -186,7 +187,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
 
   // When user finishes moving the active slider, update slice label
   makeSliderSetFn(axis: string) {
-    return (values: number[]) => this.setSliderState(axis, values);
+    return (values: number[]) => this.setSliderState(axis, [values[0], values[values.length - 1]]);
```
toloudis, line 190: a comment about handling single slider vs range slider might be in order here also

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: Right - I'm thinking I'd like to make one more PR to roi-slider-updates after all issues on the tracker are resolved to clarify this chunk of code, i.e. how we translate slices to volume-viewer clipping coordinates, and to address other feedback from this last round of meetings and slack conversation.

PR #66, opened 2022-09-08
Feature/ROI sliders: responsive design

### LINKED ISSUE #50: ROI: responsive design
## Use Case

Subtask of #34
When the window is sufficiently narrow, there is not enough room for the ROI sliders and they would be impractical to use.  

## Solution

responsive design to spec.  We will hide the sliders at some breakpoint.

frasercl:
Resolves #50 by replacing the sliders with an error message at very narrow widths, and allowing play buttons to shift to below the slice numbers to make more room for the slider in 2d mode.

NOTE: the `@media` query at the bottom of AxisClipSliders/styles.css was accidentally included as part of the last pull request, but was added as part of this task. See [this commit](https://github.com/allen-cell-animated/website-3d-cell-viewer/pull/64/commits/cb70a78852a75d872aa43f6bf82b453d418214bc) for those changes.

Screenshots:
image
image

### DIFF
--- a/src/aics-image-viewer/components/AxisClipSliders/styles.css
+++ b/src/aics-image-viewer/components/AxisClipSliders/styles.css
@@ -9,15 +9,16 @@
   @extend %axis-override;
   max-width: 665px;
   position: absolute;
-  margin: auto;
+  margin: 0 auto;
   left: 0;
   right: 0;
   padding: 0 24px;
 
   .slider-row {
     display: flex;
-    justify-content: center;
-    align-items: center;
+    justify-content: flex-end;
+    align-items: flex-start;
+    flex-wrap: wrap;
     color: #bfbfbf;
 
     &.slider-x {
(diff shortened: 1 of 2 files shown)


src/aics-image-viewer/components/AxisClipSliders/styles.css
```diff
@@ -72,8 +80,22 @@
...
+    /* Make room for play buttons */
+    max-width: 755px;
+
+    /* Slider slice display contains 2 numbers in 2d mode, not 3 - shrink accordingly */
+    .slider-slices {
+      flex: 0 0 60px;
+    }
+
+    .axis-slider-container {
+      flex: 0 1 592px;
```
toloudis, line 97: in here I see 592, 755, and 617px specified.  Are these magic numbers coming directly from the design document?

frasercl: In order to get flexbox to scale, center, and wrap things right, I had to specify the widths of several elements (e.g. slice numbers) beyond what was in the design document. (We would have gotten bad results from unspecified widths anyways, e.g. the centered position of a slider being partially dependent on the width of the slice numbers, and therefore on the position of the slider handle.) These large magic numbers are container elements that also need a baseline width, and are derived from the sum of their children's widths. (This probably should at least be documented.)

toloudis: the comments really help a lot

toloudis (approved):

PR #67, opened 2022-09-09
update some dependencies

toloudis:
While trying to do a reftresh on `nbvv`, which integrates this component, I was trying to build and run from main branch on windows and some dependencies were out of sync.  
Probably my config, but I believe it's very safe to update these dependencies:

* jest version update
* newer nouislider versions now ship in a dist/ directory, not distribute/

I will also test out this branch on a linux or mac os before merging anything here.


toloudis:
converted to draft, as react-nouislider depends on an old version of nouislider which uses `distribute` instead of `dist`.  Newer versions use dist but react-uislider has not kept up.

toloudis:
If we are to keep going with nouislider, I recommend we create a fork of mmarkelov/react-nouislider and incorporate the PRs https://github.com/mmarkelov/react-nouislider/pull/53 and https://github.com/mmarkelov/react-nouislider/pull/52 ... then we can maintain our own wrapper and keep nouislider up to date with latest enhancements.

toloudis:
Closing due to being both stale and unproductive

PR #68, opened 2022-09-09
Feature/roi slider polish

frasercl:
Some final polish to ROI sliders, including:
- Update to how sliders map to clipping. Now, for e.g. an axis with 50 slices, slider positions `n` and `m` range from 0-49, and are mapped to clipping values `[n/50 - 0.5, (m+1)/50 - 0.5]`. This ensures at least one slice is visible for all valid slider positions.
- The third slices indicator number (in parentheses) in 3D mode now represents the number of slices currently visible on that axis, rather than the total number of slices. Also includes some very minor style changes to give the indicator some extra room.
- Clipping now resets when FOV changes (single cell vs. full field). Note that the way this is currently implemented creates some potentially unexpected behavior, as the slider range and slider handle positions visibly update separately and in that order.

### DIFF
--- a/src/aics-image-viewer/components/AxisClipSliders/index.tsx
+++ b/src/aics-image-viewer/components/AxisClipSliders/index.tsx
@@ -49,9 +49,16 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
     this.createSlider = this.createSlider.bind(this);
   }
 
-  // Reset sliders and pause if mode has changed
+  // Reset sliders and pause if mode or fov has changed
+  // TODO: this reset-state-on-props-update pattern is somewhat bad practice,
+  //   and indicates some state should potentially be lifted out of this component.
   componentDidUpdate(prevProps: AxisClipSlidersProps) {
-    if (prevProps.mode !== this.props.mode) {
+    const numSlicesChanged = AXES.reduce(
+      (hasChanged, axis) => hasChanged || prevProps.numSlices[axis] !== this.props.numSlices[axis],
+      false
+    );
+
+    if (numSlicesChanged || prevProps.mode !== this.props.mode) {
       window.clearInterval(this.state.intervalId);
       this.setState({ sliders: this.getSliderDefaults(), playing: false });
 
@@ -133,7 +140,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
     const range = { min: 0, max: numSlices - 1 };
 
     return (
-      <div key={axis} className={`slider-row slider-${axis}`}>
+      <div key={axis + numSlices} className={`slider-row slider-${axis}`}>
         <span className="axis-slider-container">
           <span className="axis-slider">
             <Nouislider
--- a/src/aics-image-viewer/components/AxisClipSliders/index.tsx
+++ b/src/aics-image-viewer/components/AxisClipSliders/index.tsx
@@ -67,7 +67,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
     }
   }
 
-  getSliderDefaults = () => mapValues(this.props.numSlices, (max: number) => [0, max]);
+  getSliderDefaults = () => mapValues(this.props.numSlices, (max: number) => [0, max - 1]);
 
   getActiveAxis() {
     switch (this.props.mode) {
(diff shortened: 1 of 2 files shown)


src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -149,7 +157,9 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
           </span>
           <span className="slider-name">{axis.toUpperCase()}</span>
           <span className="slider-slices">
-            {twoD ? `${sliderVals[0]} (${range.max})` : `${sliderVals[0]}, ${sliderVals[1]} (${range.max})`}
+            {twoD
+              ? `${sliderVals[0]} (${numSlices})`
+              : `${sliderVals[0]}, ${sliderVals[1]} (${sliderVals[1] - sliderVals[0] + 1})`}
```
toloudis, line 162: I'm slightly worried that this is now inconsistent for users, although hopefully understandable/interpretable in both cases.  
I would feel a tiny bit more comfortable if the 2d case used something like `index / max` instead of `index (max)` but we can run it by @lynwilhelm .. it's not a big deal for me.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -172,18 +182,14 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
   makeSliderUpdateFn(axis: string) {
     return (values: number[]) => {
       if (this.props.setAxisClip) {
-        const step = 1;
-        const stepEpsilon = 0.04;
+        // get a value from -0.5..0.5
         const max = this.props.numSlices[axis];
-        const isActiveAxis = this.getActiveAxis() === axis;
-        const thicknessReduce = isActiveAxis ? step - stepEpsilon : 0.0;
```
toloudis, line 179: I believe stepEpsilon was there because in 2d modes we had some difficulty showing very thin slabs (~1 slice) and needed to guarantee at least some minimum thickness. Have you seen any issues with single slice viewing in any of the 3 projections?  The old issue was an artifact of the ray marching algorithm in the shader having some tiny offsets to try to anti-alias the render.  We can discuss but if nothing looks wrong then that is great.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: I ended up feeling tentatively comfortable that users would get it while using the feature, since they'll see that third number change in 3D mode, and understand that "number of slices selected" isn't meaningful in 2d mode. But that's maybe more discovery than we want to expect users to do themselves when we don't provide any cue of what the numbers mean, and in fact indicate that they might be the same thing. Maybe worth adding words as Lyndsay suggested to indicate the meaning of the numbers? Something like "13 (of 25)" or "12, 15 (3 slices)"? Definitely worth running by Lyndsay and the team regardless.

Relatedly, once merged, do we/should we want to leave the new changes deployed on nightly for a day or two to let the rest of the team experiment and offer feedback before marking a new version?

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: Nothing was looking wrong to me when I tested it. If that's the case though it might be worth adding a minimum thickness value anyways, since slice thickness can get basically arbitrarily small for more and more detailed models.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -128,11 +135,12 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
...
   createSlider(axis: string, twoD: boolean) {
     const { playing, sliders } = this.state;
+    const numSlices = this.props.numSlices[axis];
     const sliderVals = sliders[axis];
-    const range = { min: 0, max: this.props.numSlices[axis] };
+    const range = { min: 0, max: numSlices - 1 };
 
     return (
-      <div key={axis} className={`slider-row slider-${axis}`}>
+      <div key={axis + numSlices} className={`slider-row slider-${axis}`}>
```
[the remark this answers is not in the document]
toloudis: I see- it's still possible for the number of z slices to be the same for full field and single cell (less likely for x and y), so you might want to use some even more unique way to generate the key.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
toloudis: Using nightly for smoke testing for a day or two makes sense to me, yes.

I still like the conciseness of the parentheses but maybe a hover tooltip over just the parenthetical would help explain it.  
On that note of user discoverability, did you do hover tooltips for the toolbar buttons?

frasercl: Yep - icon-only buttons on the toolbar have tooltips.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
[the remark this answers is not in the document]
frasercl: That case shouldn't be a problem, since the slider range doesn't need to update, and `componentDidUpdate` should handle resetting the handle positions.

toloudis (approved):

PR #71, opened 2022-09-16
Feature/tabbed & typescripted control panel

### LINKED ISSUE #70: tabbed sections of user interface
## Use Case

The tab to close the control panel pokes into the 3d view area intrusively.  We can redesign this to add a vertical strip to the left, put the open/close tab there, and this also provides a nice way to future-proof the ui for added functionality by allowing tabbed sections of controls.

## Solution

[Design spec link](https://xd.adobe.com/view/904d6444-a800-48b6-8844-4f6f116d85ab-c51e/)

frasercl:
Resolves #70: adds a new column to the left of the control panel, moves the control panel collapse button there, and adds new tab controls for switching between categorized control panel settings. (See [design spec](https://xd.adobe.com/view/904d6444-a800-48b6-8844-4f6f116d85ab-c51e/).)

Additionally, converts `ControlPanel` and its children to TypeScript. (This task revealed a few other code maintenance tasks that could be done on these components, but I held off on them so as not to block this new feature and to keep the diff manageable.)

Screenshots:
image
image
image


src/aics-image-viewer/components/GlobalVolumeControls.tsx
```diff
@@ -99,6 +122,7 @@ export default class GlobalVolumeControls extends React.Component {
     return this.createSliderRow(config);
   }
 
+  // TODO: `config` object has no type annotation; this component deserves some DRY edits that may reduce it out
```
toloudis, line 125: once we are fully "typescriptified" we need to have this be a `lint` or `typeCheck` violation.

src/aics-image-viewer/components/GlobalVolumeControls.tsx
```diff
@@ -175,14 +193,15 @@ const STYLES = {
...
+    paddingTop: "18px",
   },
   controlRow: {
     height: "3em",
     display: "flex",
   },
   controlName: {
     flex: 2,
-    whiteSpace: "nowrap",
+    "white-space": "nowrap",
```
toloudis, line 204: really?

toloudis (approved):
LGTM...

src/aics-image-viewer/components/GlobalVolumeControls.tsx
frasercl: Agreed. The intent of this comment (which I did a poor job expressing in few words) is "I would have written a type for `config`, but I don't think it has a good reason to exist in the first place, but if it's going to be removed it'll be as part of a larger refactor of this component that's beyond the scope of my current task."

src/aics-image-viewer/components/GlobalVolumeControls.tsx
frasercl: Both are correct, but the first didn't typecheck. I looked into this, and it seems `whiteSpace` has a union type of string literals for all the allowable values of this property, but the type checker has no idea about that when it decides on a type for `STYLES`, so that property just gets the `string` type. There might be other solutions for this (putting this style inline, putting it in an actual stylesheet, maybe making `STYLES` read only?) but this was the quickest.

toloudis: it's a minor thing but consider adding a little comment saying that the typechecker complains about the camelCased version of this

frasercl: Ah, it's also possible to say `whiteSpace: "nowrap" as "nowrap"`. Any opinions as to which is cleaner?

toloudis: very slight preference for the latter

PR #76, opened 2022-09-21
Feature: screenshot button

### LINKED ISSUE #74: screenshot button
## Use Case

Users would like to be able to quickly and easily capture the view from the 3d viewer

## Solution

The core viewer has already implemented this.  A single button should be added to the interface to call the underlying function.  The viewer will automatically generate and download the file to a generic filename like snapshot.png.  User will not be required to do any other interaction.

I suggest that this feature might be useful enough to deserve to be in the toolbar as a small icon button.

## Alternatives

Could prompt for filename but this adds interactions that may slow the user down.

frasercl:
Resolves #74: adds a screenshot button to the location [specified](https://xd.adobe.com/view/904d6444-a800-48b6-8844-4f6f116d85ab-c51e/screen/a01f264e-ec99-42ab-a84a-90f2e17a533b/specs/). When the user clicks it, the viewer will download a screenshot of the current view named "screenshot.png."

Screenshot:
image


toloudis (approved):
yay

PR #78, opened 2022-10-03
Fix/Convert components to TypeScript

frasercl:
Partially addresses #57 by converting all components (but one) to TypeScript.

Additionally:
- Makes significant edits to `GlobalVolumeControls` and `ChannelsWidgetRow` to cut down on boilerplate
- Renames `colorObjectArrayConverting.ts` to `colorRepresentations.ts` and makes that file responsible for both color types (object and array) used across many different components.
- Removes dependencies: `classnames` (still used by other dependencies, but now unused in the app) and `react-numeric-input` (used only by dead code in `ChannelsWidgetRow`).

Work still to do on #57:
- ts-ify `TFEditor`. This component was large enough to leave for its own future PR.
- ts-ify other utils, functions, etc. outside of `src/components/`.
- add proper method signatures to `App`
- address #32 to the extent a solution might rely on TypeScript.


src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -24,7 +45,7 @@ export default class ViewerWrapper extends React.Component {
     }
   }
 
-  componentDidUpdate(prevProps, _prevState) {
+  componentDidUpdate(prevProps: ViewerWrapperProps, _prevState: {}) {
```
toloudis, line 48: maybe an empty `interface ViewerWrapperState {}` would make this `_prevState` type look slightly cleaner?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -68,23 +89,23 @@ export default class ViewerWrapper extends React.Component {
 
 const STYLES = {
   viewer: {
-    display: "flex",
-    position: "relative",
+    display: "flex" as "flex",
+    position: "relative" as "relative",
```
toloudis, line 93: thank you typescript.

src/aics-image-viewer/components/ChannelsWidget/index.tsx
```diff
@@ -31,16 +32,13 @@ interface ChannelSettings {
   opacity: number;
   color: [number, number, number];
   dataReady: boolean;
-  controlPoints: [];
+  controlPoints: {
+    color: string;
+    opacity: number;
+    x: number;
+  }[];
```
toloudis, line 39: a type definition for this might be in order some time.  it would be interesting to see if we could define it all the way down in volume-viewer and just reuse it up here.

src/aics-image-viewer/components/ChannelsWidget/index.tsx
```diff
@@ -160,10 +151,7 @@ export default class ChannelsWidget extends React.Component<ChannelsWidgetProps,
                       channelControlPoints={thisChannelSettings[LUT_CONTROL_POINTS]}
                       colorizeEnabled={thisChannelSettings[COLORIZE_ENABLED]}
                       colorizeAlpha={thisChannelSettings[COLORIZE_ALPHA]}
-                      isovalue={thisChannelSettings.isovalue}
-                      opacity={thisChannelSettings.opacity}
```
toloudis, line 164: why are these removed?

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+  isosurfaceChecked: boolean;
+  colorizeEnabled: boolean;
+  colorizeAlpha: number;
+  color: [number, number, number];
+  channelControlPoints: {
+    color: string;
+    opacity: number;
+    x: number;
+  }[];
+  channelDataForChannel: any; // volume-viewer Channel type
```
toloudis, line 42: // TODO export Channel type from volume-viewer and import here

toloudis: ...or find a way not to need to save a reference for it.

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+  onOpacityChangeUnwrapped = this.createChannelSettingHandler(OPACITY);
+  onOpacityChange = (newValue: number) => this.onOpacityChangeUnwrapped(newValue / ISOSURFACE_OPACITY_SLIDER_MAX);
+
+  createSliderRow = (name: string, maxValue: number, defaultValue: number, onChange: (newValue: any) => void) => (
+    <Row>
+      <Col span={10}>
+        <label style={STYLES.controlName}>{name}</label>
+      </Col>
+      <Col span={12}>
+        <Slider // TODO: this is the only remaining place we're using antd's slider rather than Nouislider. Replace it?
```
toloudis, line 93: interesting possible problem: the react wrapper we use for nouislider is currently way behind nouislider itself :(

toloudis: the number one reason we use nouislider is for the "range slider" with two handles, and the convenient ability to drag the section between the handles to move it.

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: It's also technically possible to remove `_prevState` entirely, depending on how we feel about incomplete argument lists. It's not used in the body of the function.

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: I looked around for a little while for some type system magic that would convince TS to interpret all the string fields of an object as literal types. Maybe someday I'll learn that it's possible.

toloudis: this might be the best choice: https://stackoverflow.com/questions/46710747/type-string-is-not-assignable-to-type-inherit-initial-unset-fixe (use CSSProperties type)

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
toloudis: i kind of like the completeness, to be explicit about what is going on.

src/aics-image-viewer/components/ChannelsWidget/index.tsx
frasercl: I believe TypeScript's analysis revealed that they were unused in that component. The channel controls set these values via methods, and the only way they're displayed is via the slider tooltip, which comes right out of that component's state.

src/aics-image-viewer/components/ChannelsWidget/index.tsx
frasercl: This is already basically duplicated from `App/types.ts`, so it would definitely useful to have a single source of truth for this interface.

toloudis: I actually meant just the control point itself (x, color, opacity) living down in volume-viewer.  But I agree with what you said too.. duplicate type defs are worse than duplicated code

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: This makes sense to me. Though this would make the type of all the `STYLES` objects look like `{ [class: string]: React.CSSProperties }`. Is it reasonable to inline that everywhere it might appear, or is it worth defining that type in a single place as well?

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
frasercl: I primarily bring it up for design consistency. It's odd to have this little buried corner of the UI use a slider that looks completely different from the rest of the sliders in the app.

toloudis: agreed, it's good to have that comment reminder in there.

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
toloudis: I think you would just do 
```
const myStyles: React.CSSProperties = {
  position: 'absolute',
}
```
and done.

toloudis: Sorry, I see that it's nested.  Yeah I believe you'd have to type the nested objects.

toloudis: you could use the postfix `as React.CSSProperties;` as well.  I think they are equally readable in this scenario.

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -66,7 +87,7 @@ export default class ViewerWrapper extends React.Component {
   }
 }
 
-const STYLES = {
+const STYLES: { [key: string]: React.CSSProperties } = {
```
toloudis, line 90: This works great. I probably would have done
```
const STYLES = {
viewer: {...} as React.CSSProperties,
etc
};
```

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: Multiple other stateless class components use an anonymous object type in generics - e.g. `... extends React.Component<Props, {}>`. Should these components also get empty state interfaces for consistency?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: I preferred typing the overall object in part for completeness. It eliminates a (pretty trivial) extra hoop to jump through if e.g. someone needs to add a style to one of these objects later down the line, and from a readability perspective makes clear that the purpose of the object is only to hold collections of properties.

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
toloudis: I wouldn't be opposed to that.  My general threshold would be as soon as it is needing to be used in multiple places, (ie. as soon as it showed up in componentDidUpdate in addition to the class declaration), then it makes sense to give it a name.

toloudis (approved):

PR #82, opened 2022-10-11
Fix/`App` function signatures, volume-viewer types

frasercl:
## Problem

Function signatures in `App` are a major missing piece of this component's TypeScript conversion story. Additionally, the new version of volume-viewer comes with `.d.ts` type declarations, which produced exciting new type errors.

Stronger typing also introduces an opportunity to remove many symbolic constants, as we can instead rely on type resolution for naming consistency. This PR thus makes significant progress on both #57 and #32.

## Solution

- Add types to `App` method signatures
- Bump volume-viewer and resolve type errors created by including its types
- Add new type: `ChannelState` (for channel settings that appear in state, vs. `ViewerChannelSetting` for how channel settings are passed in via props)
- Remove symbolic constants wherever properties are now being held down by types like `ChannelState`
- Remove alpha slider from channel color pickers (unused, was a bit glitchy, made color types unnecessarily difficult)

## Remaining work

On #57:
- ts-ify `TFEditor`
- ts-ify other non-component files
- go back over everything and make sure everything has proper return types (including `void`? maybe)

On #32:
- Redesign `App.handleChangeToImage` - the huge `switch` statement in this method is keeping most of the remaining symbolic constants alive
- Some of the rest should probably be TypeScript enums - convert them to enums (this also kind of counts as a #57 task)


src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -24,14 +24,14 @@ interface ViewerWrapperProps {
     axisClipSliders: boolean;
   };
   onView3DCreated: (view3d: View3d) => void;
-  setAxisClip: (axis: string, minval: number, maxval: number, isOrthoAxis: boolean) => void;
+  setAxisClip: (axis: "x" | "y" | "z", minval: number, maxval: number, isOrthoAxis: boolean) => void;
```
toloudis, line 27: use AxisName here?

src/aics-image-viewer/components/ChannelsWidget/index.tsx
```diff
@@ -62,27 +48,32 @@ export default class ChannelsWidget extends React.Component<ChannelsWidgetProps,
...
-    const isoChecked = arrayOfNames.filter((name: string) => {
-      const channelSetting = find(channelSettings, { name });
-      return channelSetting && channelSetting[ISO_SURFACE_ENABLED];
+      if (!channelSetting) return;
+      if (channelSetting.volumeEnabled) {
+        volChecked.push(channelIndex);
+      }
+      if (channelSetting.isosurfaceEnabled) {
+        isoChecked.push(channelIndex);
+      }
```
toloudis, line 74: why did this function implementation change?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
toloudis: somewhere in `shared/`, like maybe `shared/enums` might be a place for such a definition, but it's not critical for this PR

toloudis: of course, my other thought is to make it a type that comes out of volume-viewer but there's probably good reason to keep the two layers separated (the ui type might need to diverge from the lower level type, and you'd want a uiAxisToVolviewerAxis() function before passing the axis name down to the lower level interface).  I haven't looked hard enough at the code to see if that would be needed...

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -64,24 +61,24 @@ export default class ChannelsWidgetRow extends React.Component<ChannelsWidgetRow
...
-  createChannelSettingHandler = (settingKey: string) => (newValue: any) => {
+  createChannelSettingHandler = (settingKey: ChannelStateKey) => (newValue: any) => {
     const { channelName, index, changeOneChannelSetting } = this.props;
     changeOneChannelSetting(channelName, index, settingKey, newValue);
   };
 
-  onIsovalueChange = this.createChannelSettingHandler(ISO_VALUE);
-  onOpacityChangeUnwrapped = this.createChannelSettingHandler(OPACITY);
+  onIsovalueChange = this.createChannelSettingHandler("isovalue");
+  onOpacityChangeUnwrapped = this.createChannelSettingHandler("opacity");
```
toloudis, line 81: Are these strings `ChannelStateKey`s ?  it would be cool if there were a typescripty way to enforce that.  

If not, then I think passing these as string literals inline is worse than before, because suddenly the code is more error prone.  If we follow this up quickly with resolving that in another PR (by using an enum or reinstating symbolic constants somehow) then can let it be for now.

src/aics-image-viewer/components/GlobalVolumeControls.tsx
```diff
@@ -47,7 +42,7 @@ export default class GlobalVolumeControls extends React.Component<GlobalVolumeCo
     return newImage || newSliderValue || newPathTraceValue;
   }
 
-  createSliderRow = (label: string, start: number[], max: number, propKey: string) => (
+  createSliderRow = (label: string, start: number[], max: number, propKey: UserSelectionKey) => (
```
toloudis, line 45: how is UserSelectionKey different than ChannelStateKey?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: I've honestly been waiting for a critical mass of types without an obvious home to justify creating shared/types.ts, but there have been pretty few uncategorized utility types so far.

src/aics-image-viewer/components/ChannelsWidget/index.tsx
frasercl: `SharedCheckbox` expects two lists of keys: `allOptions` for the options it controls, and `checkedList` for the options which are currently checked. TypeScript revealed that these lists were different types: `allOptions` was a list of channel indexes, while `checkedList` was a list of channel names. Bizarrely, this worked just fine before, as only the length and not the content of `checkedList` ever matters for the component's behavior. Regardless, it was still a type error (and this list was still being built by two loops when one would do).

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
frasercl: As written, it is a type error to pass any string literal that isn't a `keyof ChannelState`. This isn't as great for refactoring, as you can't hit F2 on the property and have it change this string literal, but it'll still catch and flag the inconsistency when you do.

src/aics-image-viewer/components/GlobalVolumeControls.tsx
frasercl: `type UserSelectionKey = keyof UserSelectionState` (vs. `ChannelStateKey = keyof ChannelState`). I'm not totally sure that these types should have aliases, since they only save a handful of characters each and might be less clear than writing out the full expression, but they are both common enough to merit shortcutting.

toloudis: my question really was leading to whether there's a way to merge the two concepts: are the keys of UserSelectionState really that different from those of ChannelState?

toloudis (approved):

src/aics-image-viewer/components/GlobalVolumeControls.tsx
frasercl: The difference has to do not just with the type it keys into but which `App` methods these handlers call down to, e.g. `handleChangeUserSelection` for `UserSelectionKey` vs. `changeOneChannelSetting` for `ChannelStateKey`. The only `App` method that takes both as written is `handleChangeToImage`, which I left loosely typed because I couldn't get typing as strict as I wanted for complicated reasons of generics and type guards. (I'll get back to it for symbolic constants.)

PR #83, opened 2022-10-12
Feature/new icons

frasercl:
Adds Thao's new icons by:
- Adding dev dependency `@svgr/webpack` to load svgs as React components
- Creating a new shared convenience component `ViewerIcon`

Screenshot:
image

NOTE: We don't yet have a custom icon for the screenshot button (top right).


src/aics-image-viewer/components/shared/ViewerIcon.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Icon } from "antd";
+
+import ICONS from "../../assets/icons";
+
+// antd icons have a habit of appearing off center
+// TODO: come back and verify if this is still an issue after antd upgrade
+const STYLE = { fontSize: "19px", transform: "translateY(-0.5px) translateX(0.5px)" };
+
+/** Wrapper component for easy inclusion of our own custom icons. */
+const ViewerIcon: React.FC<{ type: keyof typeof ICONS }> = ({ type }) => <Icon component={ICONS[type]} style={STYLE} />;
```
toloudis, line 11: cool

toloudis: just as devil's advocate, because `ViewerIcon` is so similar to `Icon`, you could just directly import the individual svgs in the TS files that need them and just use `Icon` right?  and you wouldn't need the assets/icons/index.ts file at all.   The icon style could be a shared constant or a css class.

webpack.common.js
```diff
@@ -77,9 +77,20 @@ module.exports = {
         ],
       },
       {
-        test: /\.(woff|woff2|tff|eot|glyph|svg)$/,
+        test: /\.(woff|woff2|tff|eot|glyph)$/,
         type: "asset/resource",
       },
+      {
+        test: /\.svg$/,
```
toloudis, line 84: is this the first time we are using custom svg assets in this codebase?

frasercl: There are some font svgs, but they're all also imported as other formats too. I'm still not super familiar with the font loading story here either since in practice the fonts we include don't make it to the client.

src/aics-image-viewer/components/shared/ViewerIcon.tsx
frasercl: Yeah I'm honestly 50/50 on it - this component used to be slightly bigger before I added svgr, and if I'd made it as easy as svgr does in the first place I might not have thought to create it. The things it's giving us though is: 1. specific styling (using that STYLE constant, rather than trying to catch custom icons in a global style that might be leaky; 2. easy addition (add to `icons/index.ts` and that key automatically becomes a valid value for `type` and doesn't have to be imported individually at the site it's used).

toloudis (approved):

toloudis:
please do a patch version release after this merges :)

PR #84, opened 2022-10-15
Fix/remove symbolic constants

### LINKED ISSUE #32: refactor: get rid of symbolic constants as object property names

frasercl:
Resolves #32.

The remaining symbolic constants were being kept around for one of two reasons:
1. They were effectively an enum variant. These are replaced with new TypeScript `const enum`s: `ViewMode`, `RenderMode`, and `ImageType`.
2. They were a key to the large switch statement in `handleChangeToImage`. Here, I replaced the switch statement with two maps of handlers for properties of `UserSelectionState` and properties of of `ChannelState`, and private methods `handleChangeUserSelection` and `handleChangeChannelSetting` for interfacing with them. This allows proper type-checking and eliminates the need for the constants. I also reworked some interfaces around these new functions, including changing the existing `handleChangeUserSelection` to `changeUserSelection`.


src/aics-image-viewer/components/Toolbar/ViewModeRadioButtons.tsx
```diff
@@ -1,22 +1,22 @@
...
-      props.onViewModeChange(mode);
+    if (props.mode !== target.value) {
+      props.onViewModeChange(target.value);
     }
   };
 
   return (
     <Radio.Group onChange={onChangeButton} value={props.mode.toString()}>
-      {Object.values(viewModeMapping).map((mode, index) => (
+      {viewModes.map((mode, index) => (
```
toloudis, line 17: I'd remove the comment that describes this as a hack because you are truly defining the UI display order here with the viewModes array.  (Could rename it to viewModesDisplayOrder or some such...)  or write it like this:
```
{[ViewMode.threeD, ViewMode.xy, ViewMode.xz, ViewMode.yz].map((mode, index) => (
```
but maybe that's slightly too verbose?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -788,100 +799,69 @@ export default class App extends React.Component<AppProps, AppState> {
...
+    showAxes: (showing, view3d) => view3d.setShowAxis(showing),
+    showBoundingBox: (showing, view3d, image) => view3d.setShowBoundingBox(image, showing),
+    boundingBoxColor: (color, view3d, image) => view3d.setBoundingBoxColor(image, colorArrayToFloats(color)),
+    backgroundColor: (color, view3d) => view3d.setBackgroundColor(colorArrayToFloats(color)),
+
+    alphaMaskSliderLevel: (value, view3d, image) => {
+      view3d.updateMaskAlpha(image, alphaSliderToImageValue(value));
+      view3d.updateActiveChannels(image);
+    },
+    brightnessSliderLevel: (value, view3d) => {
```
toloudis, line 822: should typescript warn when there's no third parameter?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -969,26 +949,24 @@ export default class App extends React.Component<AppProps, AppState> {
...
-      maxProject: newAlgorithm === MAX_PROJECT,
-      autorotate: newAlgorithm === PATH_TRACE ? false : userSelections.autorotate,
+      pathTrace: newAlgorithm === RenderMode.pathTrace,
+      maxProject: newAlgorithm === RenderMode.maxProject,
+      autorotate: newAlgorithm === RenderMode.pathTrace ? false : userSelections.autorotate,
     });
-    this.handleChangeToImage(PATH_TRACE, newAlgorithm === PATH_TRACE);
-    this.handleChangeToImage(MAX_PROJECT, newAlgorithm === MAX_PROJECT);
+    this.handleChangeUserSelection("pathTrace", newAlgorithm === RenderMode.pathTrace);
+    this.handleChangeUserSelection("maxProject", newAlgorithm === RenderMode.maxProject);
```
toloudis, line 964: I'm slightly dubious that we gained all that much when I see explicit quoted strings being passed in, but then I remember that it's type checked!  A casual reader of this code would not know that at a glance.

frasercl: You're right to be at least a little bit dubious, since proper type-checking for these generic functions is basically dependent on callers passing actual string literals rather than variables. It's possible e.g. for someone to invoke
```
handleChangeUserSelection<ChannelStateKey>(key, value);
```
and have the type of `value` be assessed as the union of all types that appear in `ChannelState`, regardless of which `key` is actually being changed.

Anyways, should the type-checking behavior of this and similar functions be documented somewhere?

src/aics-image-viewer/components/App/index.tsx
frasercl: Callbacks passed to `map` or `reduce` nominally take three or four arguments, but the last one is rarely defined. Given that example alone I think it's reasonable for TypeScript to accept functions that define fewer arguments than they'll actually be called with. In this specific case though, I'm with you in being somewhat unsure of this choice stylistically; I'd absolutely be open to explicitly filling in those unused parameters, e.g. `brightnessSliderLevel: (value, view3d, _image) => { ... }`

toloudis: I think just adding the underscore param would make it more obvious that we expect all the function signatures to be the same

src/aics-image-viewer/components/App/index.tsx
toloudis: I still wonder if it would be better to just let each change function have its own well typed arguments. It's going to move closer and closer to just having explicit functions.
```
handleChangeUserSelectionMaxProject(isMaxProject); // typed as boolean because that's what maxproject wants
```
or
```
handleChangeUserSelection["maxProject"](isMaxProject); 
```
where handleChangeUserSelection is the object that holds on to all the functions, and is allowed to have many functions all with different signatures

frasercl: The main convenience I think we're getting from having a single way into these maps of handlers is having not-undefined guards for `view3d` and `image` in a single place rather than duplicated across a bunch of handlers. But there are good potential solutions for that too. e.g. in `TFEditor` a number of important properties aren't set until `componentDidMount`, and eventually rather than guard these types everywhere they're needed I started defining them in the class interface with a `!:` for "this property will spend part of its lifecycle undefined, but will never be accessed before being given its initial value". If that's the case for `view3d` and `image` too then we can make that change in `AppState` and not have to worry about putting not-undefined checks all over the place. In that case I think it might make more sense to either pass the big handler map around or else just define individual handlers.

toloudis (approved):

PR #86, opened 2022-10-19
Fix/remove duplicate `maskChannelName`

frasercl:
## Problem

`App` currently has two required props named `maskChannelName`, one on the base `AppProps` type and one on `ViewerChannelSettings`. Neither actually affect anything because the segmentation channel name is hardcoded via the constant `CELL_SEGMENTATION_CHANNEL_NAME`.

## Solution

- Remove the prop on `AppProps`, make the one on `maskChannelName` optional.
- Remove `CELL_SEGMENTATION_CHANNEL_NAME` and make the viewer search for the channel named by the prop instead. I have confirmed this still works in CFE, because it already sets this prop to the same value as was previously hardcoded.

### DIFF
--- a/src/aics-image-viewer/shared/constants.ts
+++ b/src/aics-image-viewer/shared/constants.ts
@@ -1,8 +1,7 @@
 import { ColorArray } from "./utils/colorRepresentations";
 
 // Add all exported constants here to prevent circular dependencies
-export const 
-  // View modes
+export const // View modes
   YZ_MODE = "YZ",
   XZ_MODE = "XZ",
   XY_MODE = "XY",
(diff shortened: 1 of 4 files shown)


toloudis (approved):
Because the default is being removed, we must remember to make CFE pass the correct value in

src/aics-image-viewer/shared/constants.ts
```diff
@@ -43,8 +42,7 @@ export const
   DENSITY_SLIDER_LEVEL_DEFAULT = [50],
   LEVELS_SLIDER_DEFAULT: ColorArray = [35.0, 140.0, 255.0],
   OTHER_CHANNEL_KEY = "Other",
-  SINGLE_GROUP_CHANNEL_KEY = "Channels",
```
toloudis, line 46: was SINGLE_GROUP_CHANNEL_KEY completely unused?

frasercl: It's still there, git just logged it as a deletion + addition because it has a semicolon after it now rather than a comma.

PR #89, opened 2022-10-24
Fix/convert `TFEditor` to TypeScript

frasercl:
Converts `TFEditor` to TypeScript, and makes some stray edits to other components to tighten up types.

Also, fixed a couple errors in `TFEditor`:
- Colors were being handed around via inconsistent types, which sometimes caused unexpected behavior. I standardized the component's color types, and added a new conversion function, `colorArrayToString`, to convert colors to CSS strings for D3's benefit.
- The color picker interface would not update while open - colors would update in the volume, but the handles and numbers in `SketchPicker` would not change to reflect this. I created a small wrapper component to store and update color state. (NOTE: this is a bit of a hack as currently written, though it's possible the same functionality could be split out of the current `ColorPicker` component for a more robust solution. The underlying `SketchPicker` component also stores its own state, so it's unclear to me why this fix is necessary.)


toloudis:
What was the unexpected behavior you named in the first bullet, due to inconsistent color types?

toloudis:
Also, do you have a "less hacky" solution in mind for keeping the SketchPicker updated?  When I look at CFE, the color picker does seem to update correctly while open.

frasercl:
> What was the unexpected behavior you named in the first bullet, due to inconsistent color types?

Could have been clearer - I didn't see any user-facing strange behavior, but properties were being assigned as different types (e.g. initialized as arrays, updated as strings) in ways I didn't expect to work.

> Also, do you have a "less hacky" solution in mind for keeping the SketchPicker updated? When I look at CFE, the color picker does seem to update correctly while open.

The `ColorPicker` component used for most color inputs wraps `SketchPicker` and adds two additional features: a swatch for opening/closing the color picker, and a color state for the picker to react to. The control point color picker isn't opened with a swatch, so it uses vanilla `SketchPicker`, but because it then has no state to react to, it (at least on my machine) doesn't update as the color changes. I added a simple wrapper inline to `TFEditor` to fix this, but if it's a solution we could expect to need elsewhere, I'd recommend splitting `ColorPicker` into two components: a stateful `SketchPicker` and a wrapper around the stateful picker that adds the swatch.

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -0,0 +1,818 @@
...
+  /** Set the pixel data we are manipulating */
+  setData() {
+    if (!this.props.channelData) {
+      throw new Error("Transfer Function Editor setData called with no channel data.");
+    }
+    this.updateScales();
+    this.updateAxis();
+  }
+
+  /////// Polymer lifecycle callbacks /////////////
```
toloudis, line 717: this comment is obsolete and should be removed. the original implementation of this component was as a polymer component but it has long since been ported into react

toloudis (approved):
Looks good to me - thanks for the clarification on the updating issue.

package.json
```diff
@@ -58,9 +58,10 @@
     "@babel/preset-typescript": "^7.14.5",
     "@babel/runtime": "^7.14.6",
     "@dr.pogodin/babel-preset-svgr": "^1.4.2",
+    "@types/d3": "^4.13.12",
     "@types/jest": "^27.4.0",
     "@types/lodash": "^4.14.185",
-    "@types/react": "^16.x",
```
toloudis, line 63: is @types/react just not needed anymore?

PR #95, opened 2022-11-15
Fix/toolbar styles

### LINKED ISSUE #93: Toolbar behavior issues
## Description

There are a few issues with the toolbar

## Expected Behavior

1. Binary (on/off) buttons (turntable, axis, bounding box): When turning "off", the "on" styling persists until you click somewhere else again. The styling should go away when the user mouses away so it is obvious that the action was successful.
2. Buttons that trigger a single action (screenshot, reset camera): hover styling persists until you click somewhere else again. The styling should go away when the user mouses away so it is obvious that the action successful.
3. Might be the same issue as 1 and 3, but the same issue is happening with the select box styling after selection.
4. The hover and "on" color for icons does not match the pattern. Like text color does, they should turn #FFF (from #bfbfbf), which makes them much brighter and stand out...dare I say "pop"?:)
5. Icons are not centered. There is a line height being applied by default that should be removed/overridden. 
.ant-btn > .anticon {line-height: 1;}
6. Tooltip backgrounds cannot be seen on black background. Maybe try to create an outline or change the background color to match ant designs tooltip background?
7. Select box dropdown styling doesn't match styling specification. As is, there is an unnecessary purple highlight and the background/text color doesn't have great contrast. See attached image or design doc for style guidance.
8. Tooltip messaging is inconsistent. If it exists for some buttons, it should exist for all. Suggestions:
- Reset camera - perfect as is
- Turntable - Turn on turntable/Turn off turntable
- Axex and bounding box - perfect as is
- Download - Download
- Screenshot - Screenshot

## Environment

Screen Shot 2022-11-03 at 11 16 57 AM

Please reach out to me should there be any things we need to work out together!

frasercl:
Resolve #93.

### Focus
Controls in the toolbar have identical focus, selected, and hover states, which causes the hover/selected states to apparently persist after the user interacts with a control due to focus remaining on the element. This is addressed in multiple ways:
- Per @lynwilhelm's input, hover and focus now have a highlight color to emphasize that these states are related to user action, not app state.
- Update control panel tabs to match this new spec.
- Wherever possible, use [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible), which activates on focused elements only when the user appears to be navigating the UI with the keyboard.
  - Because ant's radio buttons move focus to an invisible child element rather than the visible element, and there is no `:focus-visible` equivalent of [`:focus-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within), styling on `:focus-visible` requires the not-well-supported [`:has`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) selector. Toolbar styles thus test for support with the `@supports` rule, and fall back to normal focus behavior if `:has` is unusable.

### Also:
- Icon colors were not changeable with hover/focus states. Change the color attribute of SVG icons to allow them to pick up the current CSS `color`. (Switch to the rich diff to view this change.)
- Remove transform on icons. This doesn't necessarily resolve the positioning issue (the transform was there to fix it in the first place), but the transform doesn't make things much better anyways, and they at least show up a bit crisper without the transform.
- Add tooltips to download and screenshot button, and make tooltip on turntable button respond to turntable on/off state.
- Style render mode dropdown to specification.

### Unresolved issues from #93:
- Icon positioning remains a bit mysterious to me
- Bordered tooltips

If these problems are worth following up on, they should be documented in new issues on the tracker.


src/aics-image-viewer/components/shared/ViewerIcon.tsx
```diff
@@ -3,9 +3,7 @@ import { Icon } from "antd";
 
 import ICONS from "../../assets/icons";
 
-// antd icons have a habit of appearing off center
-// TODO: come back and verify if this is still an issue after antd upgrade
-const STYLE = { fontSize: "19px", transform: "translateY(-0.5px) translateX(0.5px)" };
+const STYLE = { fontSize: "19px" };
```
toloudis, line 6: love this one

toloudis (approved):
LGTM, this looks really nice but I admittedly am a bad reviewer for advanced CSS :)

lynwilhelm:
@frasercl @toloudis There no way for me to review this independently before it's merged, correct? I am pretty sure everything is great on this one, but not sure what the process for review should be if you guys want to include me.

toloudis:
let's merge it and I will re-generate a build asap so you can see it live

PR #106, opened 2022-12-15
Fix/responsive toolbar

### LINKED ISSUE #105: toolbar buttons overlap when screen is not wide enough
## Description

Some Buttons overlap when screen width is too small.  The buttons are harder to click.

## Expected Behavior

Toolbar buttons should always be available and clickable.

frasercl:
Resolve #105: when centered toolbar controls would otherwise overlap with left- or right-aligned controls, collapse all controls into a single left-aligned block and enter "scrolling mode."

In scrolling mode:
- Left/right scroll buttons will appear if controls are hidden off either edge of the screen. These can be clicked to access those controls, and will disappear when scrolled all the way to one side. A border appears below the toolbar to anchor these buttons, per specification (see linked issue).
- Users may also scroll up/down with the mouse wheel to access controls clipped off the side of the screen.
- The toolbar will scroll to show focused controls for users navigating with keyboard (in all the browsers I tested).

With the addition of state to `Toolbar`, along with the necessity to reference that state in a global event handler (window resize), I refactored `Toolbar` to a class component for clarity.

Screenshots:
image
image

### DIFF
--- a/src/aics-image-viewer/components/Toolbar/index.tsx
+++ b/src/aics-image-viewer/components/Toolbar/index.tsx
@@ -80,13 +80,14 @@ export default function Toolbar(props: ToolbarProps): React.ReactElement {
     checkSize();
   }, []); // Dependency-free effect will only run on mount
 
+  const scrollX = (amount: number): number => (barRef.current!.scrollLeft += amount);
   // Translate vertical scrolling into horizontal scrolling
-  const scrollX: WheelEventHandler<HTMLDivElement> = (e) => {
+  const scrollHandler: WheelEventHandler<HTMLDivElement> = (e) => {
     e.preventDefault();
     if (e.deltaY === 0) {
       return;
     }
-    barRef.current!.scrollLeft += e.deltaY;
+    scrollX(e.deltaY);
   };
 
   const twoDMode = props.mode !== ViewMode.threeD;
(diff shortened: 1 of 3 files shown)


src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -41,116 +42,214 @@ interface ToolbarProps {
...
+    const centerRect = centerRef.current!.getBoundingClientRect();
+    const rightRect = rightRef.current!.getBoundingClientRect();
 
-  const renderGroup1 =
-    renderConfig.viewModeRadioButtons || renderConfig.resetCameraButton || renderConfig.autoRotateButton;
-  const renderGroup4 = renderConfig.showAxesButton || renderConfig.showBoundingBoxButton;
+    if (this.state.scrollMode) {
+      // Leave scroll mode if there is enough space for centered controls not to overlap left/right-aligned ones
+      const barWidth = barRef.current!.getBoundingClientRect().width;
+      const requiredWidth = Math.max(leftRect.width, rightRect.width) * 2 + centerRect.width + 30;
```
toloudis, line 82: I suggest making the 30 a named symbolic constant.  It looks like a magic number.

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -41,116 +42,214 @@ interface ToolbarProps {
...
+        this.setState({ scrollMode: false });
+      }
+    } else {
+      // Enter scroll mode if centered controls are overlapping either left/right-aligned ones
+      if (leftRect.right > centerRect.left || centerRect.right > rightRect.left) {
+        this.setState({ scrollMode: true });
+      }
+    }
+    this.checkScrollBtnVisible();
+  }, 50);
```
toloudis, line 93: This 50 could also be a named constant.  Although its meaning is a bit more self evident, it's not clear why it should be 50.

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -41,116 +42,214 @@ interface ToolbarProps {
...
+    const turntableToggleTitle = autorotate ? "Turn off turntable" : "Turn on turntable";
+
+    return (
+      <div className={`viewer-toolbar-container${scrollMode ? " viewer-toolbar-scroll" : ""}`}>
+        <div
+          className="viewer-toolbar-scroll-left"
+          style={{ display: scrollBtnLeft ? "flex" : "none" }}
+          onClick={() => this.scrollX(-100)}
+        >
+          <ViewerIcon type="closePanel" style={{ fontSize: "12px", transform: "rotate(180deg)" }} />
```
toloudis, line 146: should these styles be set in a more systematic way?  Is the 12px something that gets reused in a lot of places, for example?

src/aics-image-viewer/components/Toolbar/index.tsx
toloudis: Just to be sure I understand, is this rate-limiting the checkSize calls to no more than 1 every 50ms ?

toloudis (commented):
minor cleanup stuff but assuming it is all working, this is an awesome solution that is also validated by other professional tools doing something similar.

src/aics-image-viewer/components/Toolbar/index.tsx
frasercl: I don't believe any other icon is quite this small, and given that `ViewerIcon` sets its own `style` property with a default font size (I'm not extremely happy with this pattern but can't think of anything that would make me happier), I think it makes sense to explicitly departure from the default size of 19px in props. Though in that case this prop could still be condensed down from a general style setter to a specific `size` prop and other styles set via plain CSS.

src/aics-image-viewer/components/Toolbar/index.tsx
frasercl: Yeah, `_.debounce` delays calling a function until it hasn't been called in the last 50 (or whatever) ms. This is actually the second time I've used `debounce` with a limit of 50 - it's a totally arbitrary choice, but could be worth its own entry in constants.ts?

toloudis: yes or right here in this file, jsut give it a name like CHECKSIZE_DEBOUNCE_RATE or whatever

toloudis (approved):

PR #108, opened 2023-01-26
Fix/update volume loading

frasercl:
Update volume-viewer to 3.0.0 and use its new volume loading API, which removes the single `VolumeLoader` class in favor of specialized per-format loader classes which implement a common interface.


toloudis (approved):
good to know it was that simple.  Guess we got the exports right!

Issue #102, opened 2022-12-13
icon for metadata info panel


toloudis:
We need a new icon for the new metadata info panel.

See #38 and #104

frasercl:
Worth noting as well that we currently depend on some other icons from ant's library in production. Out of the full list [here](https://3x.ant.design/components/icon/), we use in rough order of prominence: `camera`, `setting`, `loading`, `step-backward`, `step-forward`, `down`. Some of these are prominent enough to deserve a custom design (`camera` is used in the screenshot button and lives in the toolbar alongside custom icons), others may be fine as is (`down` is the arrow in dropdowns and fits just fine with the rest of the icons in the app).

Additionally, our current antd version (v3) is all-or-nothing on icon loading: if you need one icon, your app builds with all of them. This is less an argument for making more of our own icons than it is for addressing #42.

toloudis:
see #104 for reference

thao-do:
Will attach final icons here after discussion to pick one. 

![Image](https://user-images.githubusercontent.com/25212364/223527618-b2b2aa8c-7c48-496d-a4cf-de78a1c7d4a2.png)

frasercl:
![image](https://user-images.githubusercontent.com/53030214/230964737-2e4e170b-8e7e-41c4-8831-26fa63874c1a.png)

PR #109, opened 2023-01-26
Fix/move axis and scale indicators with clipping drawer

### LINKED ISSUE #73: reconcile the 3d axis position with the clipping sliders
## Use Case

we wanted the 3d clipping sliders to be at the bottom of the 3d view so they have a chance to have a wider space.
However, the 3d axis display is presented at the lower left, and when the Clipping drawer is open, it overlays on top and hides the axes.

## Solution

Not certain of the best solution.  
* I can make a strong argument for not overlaying and instead resizing the view.  (Same goes for the toolbar).  
* However it might be a nice option to be able to specify which corner the axes go in (as a prop maybe?).  
* Also possibly we can just shift the axes up when the drawer is open.

frasercl:
## Problem

#73: the clipping panel covers the axis indicator when open.

The introduction of scale bars, which appear in the bottom-right corner of the viewport by default, creates a similar but worse problem: not only are scale bars covered by the open clipping panel, they are also covered by the button to open the clipping panel when the panel is closed.

## Solution

Use `View3d`'s new methods `setAxisPosition` and `setScaleBarPosition` to move the indicators out of the way. This requires adding a new prop `onVisibilityChange` to the `BottomPanel` component so that this state can be passed up. This resolves #73, though per comments on that issue it may be worth opening another ticket to investigate other related changes that can be made to the UI.

Additionally, the bottom panel button uses an icon from antd that's nearly identical to one of our own custom icons, so I replaced it with the custom icon, and upgraded the ergonomics of the `ViewerIcon` utility component along the way. (See comment on #102 for why we should prefer our own icons.)

## Screenshots
image
image
image

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -237,7 +237,8 @@ export default class App extends React.Component<AppProps, AppState> {
     this.updateStateOnLoadImage = this.updateStateOnLoadImage.bind(this);
     this.initializeNewImage = this.initializeNewImage.bind(this);
     this.onView3DCreated = this.onView3DCreated.bind(this);
-    this.onClippingPanelOpen = this.onClippingPanelOpen.bind(this);
+    this.onClippingPanelVisibleChange = this.onClippingPanelVisibleChange.bind(this);
+    this.onClippingPanelVisibleChangeEnd = this.onClippingPanelVisibleChangeEnd.bind(this);
     this.createChannelGrouping = this.createChannelGrouping.bind(this);
     this.beginRequestImage = this.beginRequestImage.bind(this);
     this.loadNextImage = this.loadNextImage.bind(this);
@@ -1043,9 +1044,8 @@ export default class App extends React.Component<AppProps, AppState> {
     this.state.view3d?.resetCamera();
   }
 
-  onClippingPanelOpen(open: boolean): void {
+  onClippingPanelVisibleChange(open: boolean): void {
     const CLIPPING_PANEL_HEIGHT = 130;
-    const CLIPPING_PANEL_OPEN_TIME = 300;
 
     const { view3d, userSelections } = this.state;
     if (view3d) {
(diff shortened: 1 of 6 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1035,8 +1040,36 @@ export default class App extends React.Component<AppProps, AppState> {
...
 
   onResetCamera(): void {
-    if (this.state.view3d) {
-      this.state.view3d.resetCamera();
+    this.state.view3d?.resetCamera();
+  }
+
+  onClippingPanelOpen(open: boolean): void {
+    const CLIPPING_PANEL_HEIGHT = 130;
+    const CLIPPING_PANEL_OPEN_TIME = 300;
```
toloudis, line 1048: is this related to some value associated with opening a antd Drawer?  I wonder if we could get the correct actual value from somewhere else.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1035,8 +1040,36 @@ export default class App extends React.Component<AppProps, AppState> {
   changeBoundingBoxShowing = (showing: boolean): void => this.changeUserSelection("showBoundingBox", showing);
 
   onResetCamera(): void {
-    if (this.state.view3d) {
-      this.state.view3d.resetCamera();
+    this.state.view3d?.resetCamera();
+  }
+
+  onClippingPanelOpen(open: boolean): void {
+    const CLIPPING_PANEL_HEIGHT = 130;
```
toloudis, line 1047: would be nice to get this height from somewhere else where it is specified

src/aics-image-viewer/components/App/index.tsx
toloudis: Drawer has a `afterVisibleChange` prop which might help here?

toloudis (approved):
Do we need to get a UX review on this?

PR #117, opened 2023-02-17
Feature/metadata viewer

### LINKED ISSUE #38: pass image metadata in to viewer
## Use Case

It can be useful to see some dataset-specific relevant information about an image while viewing it.  For example, in Cell Feature Explorer, we could display the values of all the measured features for each image.

## Solution

Along with a url to an image to load in the viewer, allow an arbitrary list of name-value pairs to be sent in and displayed.
They should be displayed in their own left-hand-panel section in tabular form.
This should be implemented after the ui overhaul is completed, and we have a code structure in place to add sections like this.

### LINKED ISSUE #115: Rendering Adjusments controls don't show up for images without names
## Description

If an unnamed image is loaded then the Rendering Adjustments are never seen.

## Expected Behavior

Rendering Adjustments should always be available 

## Reproduction

https://allen-cell-animated.github.io/website-3d-cell-viewer/?url=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr

Code checking for imageName:
image (4)

frasercl:
Resolves #38: creates a new "Metadata" tab in the control panel, which can display image metadata as key-value pairs and render data nested in subcategories.

Adds the following props to configure and populate the metadata panel:
- `renderConfig.metadataViewer`: boolean, whether to show the panel at all
- `metadata`: a JSON object whose values are either string, number, boolean, or nested objects of these types
- `metadataConfig`: an array of strings specifying which categories of metadata to derive from the underlying image, and in what order. If this prop is specified, the app will add an "Image" category at the top of the metadata table containing the categories specified. Available categories are currently `dimensions`, `originalDimensions`, `physicalDimensions`, `pixelPhysicalSize`, `channels`, `timeSeriesFrames`, `userData`. (Some of these may need more consideration, or removal altogether. For that matter, I'm fine changing the structure of this API entirely.)

Also, removes a null-check for `imageName` in the render function of `GlobalVolumeControls`. The image doesn't need to have a name in order to have rendering properties to adjust. This resolves #115.

### Screenshot
(from CFE)
image

### DIFF
--- a/public/index.tsx
+++ b/public/index.tsx
@@ -264,6 +264,7 @@ function runApp() {
       cellDownloadHref={args.cellDownloadHref}
       viewerConfig={viewerConfig}
       viewerChannelSettings={args.initialChannelSettings}
+      metadataConfig={["dimensions", "pixelPhysicalSize", "physicalDimensions", "channels", "userData"]}
     />,
     document.getElementById("cell-viewer")
   );
--- a/public/index.tsx
+++ b/public/index.tsx
@@ -264,7 +264,6 @@ function runApp() {
       cellDownloadHref={args.cellDownloadHref}
       viewerConfig={viewerConfig}
       viewerChannelSettings={args.initialChannelSettings}
-      metadataConfig={["dimensions", "pixelPhysicalSize", "physicalDimensions", "channels", "userData"]}
     />,
     document.getElementById("cell-viewer")
   );
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -1188,27 +1188,25 @@ export default class App extends React.Component<AppProps, AppState> {
     },
     channels: (image) => ({ Channels: image.num_channels }),
     timeSeriesFrames: (_image) => ({ "Time series frames": 1 }), // TODO
-    userData: (image) => {
-      const { userData } = image.imageInfo;
-      return userData ? { "User data": userData as MetadataRecord } : {};
-    },
+    userData: (image) => image.imageInfo.userData as MetadataRecord,
   };
 
   getMetadata(): MetadataRecord {
-    let customMetadata = {};
     const { metadata, metadataConfig } = this.props;
     const { image } = this.state;
 
-    if (image && metadataConfig) {
+    let customMetadata = {};
+    if (image && metadataConfig && metadataConfig.length > 0) {
       metadataConfig.forEach((category) => {
         if (category in this.metadataSelectors) {
           const newMetadata = this.metadataSelectors[category](image);
           customMetadata = { ...customMetadata, ...newMetadata };
         }
       });
+      return { Image: customMetadata, ...metadata };
+    } else {
+      return metadata || {};
     }
-
-    return { Image: customMetadata, ...metadata };
   }
 
   render(): React.ReactNode {
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -1,7 +1,7 @@
 // 3rd Party Imports
 import { Layout } from "antd";
 import React from "react";
-import { includes, isEqual, find, map, debounce } from "lodash";
+import { includes, isEqual, find, map, debounce, mapValues } from "lodash";
 import {
   ControlPoint,
   IVolumeLoader,
@@ -25,7 +25,7 @@ import {
   ChannelStateChangeHandlers,
   ChannelGrouping,
 } from "../../shared/utils/viewerChannelSettings";
-import { AxisName, IsosurfaceFormat, MetadataFormatRecord, MetadataRecord } from "../../shared/types";
+import { AxisName, IsosurfaceFormat, MetadataRecord } from "../../shared/types";
 import { ImageType, RenderMode, ViewMode } from "../../shared/enums";
 import {
   PRESET_COLORS_0,
@@ -1161,15 +1161,11 @@ export default class App extends React.Component<AppProps, AppState> {
     return { x: 0, y: 0, z: 0 };
   }
 
-  getExtraMetadata(): { metadata: MetadataRecord; metadataFormat: MetadataFormatRecord } {
+  getExtraMetadata(): MetadataRecord {
     if (!this.props.renderConfig.dimensionsInMetadataViewer) {
-      return { metadata: {}, metadataFormat: {} };
+      return {};
     }
-    const dimensionFormat = { unit: "px" };
-    return {
-      metadata: { Dimensions: this.getNumberOfSlices() },
-      metadataFormat: { x: dimensionFormat, y: dimensionFormat, z: dimensionFormat },
-    };
+    return { Dimensions: mapValues(this.getNumberOfSlices(), (num) => num + "px") };
   }
 
   render(): React.ReactNode {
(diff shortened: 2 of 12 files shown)


src/aics-image-viewer/components/GlobalVolumeControls.tsx
```diff
@@ -64,7 +64,6 @@ export default class GlobalVolumeControls extends React.Component<GlobalVolumeCo
   );
 
   render(): React.ReactNode {
-    if (!this.props.imageName) return null;
```
toloudis, line 67: resolves #115 ?

public/index.tsx
```diff
@@ -264,6 +264,7 @@ function runApp() {
       cellDownloadHref={args.cellDownloadHref}
       viewerConfig={viewerConfig}
       viewerChannelSettings={args.initialChannelSettings}
+      metadataConfig={["dimensions", "pixelPhysicalSize", "physicalDimensions", "channels", "userData"]}
```
toloudis, line 267: if none specified, what is the default behavior?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1159,6 +1167,48 @@ export default class App extends React.Component<AppProps, AppState> {
...
+    const { metadata, metadataConfig } = this.props;
+    const { image } = this.state;
+
+    let customMetadata = {};
+    if (image && metadataConfig && metadataConfig.length > 0) {
+      metadataConfig.forEach((category) => {
+        if (category in this.metadataSelectors) {
+          const newMetadata = this.metadataSelectors[category](image);
+          customMetadata = { ...customMetadata, ...newMetadata };
+        }
```
toloudis, line 1204: Just a thought: my overall preference is that this layer be a bit more "dumb" and to push these details all down into the Volume and VolumeLoaders.  Then you'd just have a Volume.getDisplayMetadata which returns all of it.  The categories could still be standardized down at that layer.  And I suppose they could still be filtered up at this level, although I'm not sure what the use case is for showing some categories and not others.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1159,6 +1167,48 @@ export default class App extends React.Component<AppProps, AppState> {
...
+    },
+    physicalDimensions: (image) => {
+      const { physicalScale, physicalUnitSymbol, normalizedPhysicalSize } = image;
+      const [x, y, z] = normalizedPhysicalSize.toArray().map((dim: number) => dim * physicalScale + physicalUnitSymbol);
+      return { "Physical size": { x, y, z } };
+    },
+    pixelPhysicalSize: (image) => {
+      const { pixel_size, physicalUnitSymbol } = image;
+      const [x, y, z] = pixel_size.map((dim) => dim + physicalUnitSymbol);
+      return { "Physical size per pixel": { x, y, z } };
```
toloudis, line 1187: is this displayed as 3 rows labeled "x", "y", and "z" respectively?

toloudis: also it should be noted that this is "physical size per original pixel" unless you do some more math.. I think.. worth double checking

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1159,6 +1167,48 @@ export default class App extends React.Component<AppProps, AppState> {
     return { x: 0, y: 0, z: 0 };
   }
 
+  private metadataSelectors: MetadataSelectors = {
+    dimensions: (image) => {
+      const { x, y, z } = image;
+      return { Dimensions: { x: x + "px", y: y + "px", z: z + "px" } };
+    },
+    originalDimensions: (image) => {
```
toloudis, line 1175: dimensions vs originalDimensions is a really interesting problem.  I think when we load multiresolution zarr data, originalDimensions are going to be wrong; they are not the "level zero" raw dimensions.

tsconfig.base.json
```diff
@@ -10,7 +10,6 @@
     "moduleResolution": "Node",
     "noImplicitAny": true,
     "noImplicitReturns": true,
-    "preserveConstEnums": true,
```
toloudis, line 13: what was this doing?

public/index.tsx
frasercl: If this prop is missing then the "Image" category won't appear in the panel.

src/aics-image-viewer/components/App/index.tsx
frasercl: Something like an additional field on `Volume`, set by `VolumeLoader` with whatever is available from the given format, accessible by a getter?

I guess my thought behind the category filtering was for dissemination applications like CFE, which might want to curate their metadata a bit rather than potentially overwhelming their users. But I could go either way on whether the option is useful even in that case. I was thinking of allowing clients to provide their own metadata formatting function to allow some ability to rename long property names for display (e.g. `structureSegmentationMethod`), which would also allow custom filtering. But I would be fine going with a "you get what you get" approach.

tsconfig.base.json
toloudis: Another situation where the datastructure you use for I/O needs to have different types than the run-time data structure, for flexibility.  Once an external user persists an enum value, it's permanent.   We would call the persistent ones "PermId"s  (as in permanent) and have mappings in the I/O code to map to and from the runtime data structures.  The PermId enums would be in source files that were only associated with the file format and with big comments that said, never change values, remove items or reorder this!)

toloudis:
Notes from our standup:
* VolumeLoaders should populate a metadata object that can be stored in the Volume alongside the ImageInfo
* This code can treat that object as opaque and just display it without really caring what is inside
* This code can still apply filtering (and possible renaming) on top level property names within that object
* Although it would be better for the object to provide its own display names for the properties!
* by default the panel should show all of the metadata if the prop is omitted

toloudis:
see https://github.com/allen-cell-animated/volume-viewer/pull/98

frasercl:
New commits reflect the new version of volume-viewer taking responsibility for generating metadata. This allows for insertion of fields that may only be relevant to certain file formats (e.g. selected multiscale for OME-Zarr) at load time, before file formats get largely abstracted away. Since the UI is no longer responsible for generating its own metadata and can't as easily apply categories when doing so (and since it sounds like categories have only limited usefulness anyways), I replaced `metadataConfig` with `metadataFormatter`, an optional function prop that allows clients to filter metadata, rename fields, or otherwise transform the image-derived metadata object however they like.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -1159,6 +1160,22 @@ export default class App extends React.Component<AppProps, AppState> {
     return { x: 0, y: 0, z: 0 };
   }
 
+  getMetadata(): MetadataRecord {
+    const { metadata, metadataFormatter } = this.props;
+    const { image } = this.state;
+
+    let imageMetadata = image?.imageMetadata as MetadataRecord;
```
toloudis, line 1167: Should we put the MetadataRecord type down in volume-viewer?

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: Potentially, though it matters to what extent `MetadataRecord` is a contract about what values are allowable in metadata objects vs. just a convenience to easily keep values well-typed while `MetadataViewer` is recursively rendering them. I can't think of many values that would realistically appear in a metadata object but wouldn't be covered by `MetadataRecord`, even if TypeScript can come up with edge cases. And even if it were in volume-viewer, it would still be reading external data and making guesses about its type. I don't have much of a problem making that guess here instead.

Hm - I was leaning yes when I started, then talked myself into no as I wrote.

PR #118, opened 2023-02-23
Fix/prune app state

frasercl:
## Problem

We still have a now-unused mechanism for displaying images as a "slideshow," which maintains a concept of a previous/next image in a sequence. Additionally, a number of other top-level state values are unused or can be derived from props, which is the preferred pattern in React.

## Solution

Remove unused slideshow feature, unused state, and associated methods.

Removed props:
- `nextImgPath`, `prevImgPath`, `preLoad`; from slideshow feature

Removed state:
- `nextImg`, `prevImg`; from slideshow feature
- `cellId`, `hasCellId`, `cellPath`, `fovPath`, `path`; derivable from props
- `channelDataReady`, `files`, `openFilesOnly`, `queryErrorMessage`; unused (`files` was typed as `null`, e.g.)

Removed `App` class properties:
- `stateKey`, from slideshow feature
- `openImageInterval`, unused

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -122,20 +122,14 @@ const defaultProps: AppProps = {
     slice: undefined, // or integer slice to show in view mode XY, YZ, or XZ.  mut. ex with region
   },
   baseUrl: "",
-  nextImgPath: "",
-  prevImgPath: "",
   cellId: "",
   cellDownloadHref: "",
   fovDownloadHref: "",
-  preLoad: false,
   pixelSize: undefined,
   canvasMargin: "0 0 0 0",
 };
 
 export default class App extends React.Component<AppProps, AppState> {
-  private openImageInterval: number | null;
-  private stateKey: "image" | "prevImg" | "nextImg";
-
   static defaultProps = defaultProps;
   constructor(props: AppProps) {
     super(props);
@@ -167,8 +161,6 @@ export default class App extends React.Component<AppProps, AppState> {
 
     this.state = {
       image: null,
-      nextImg: null,
-      prevImg: null,
       view3d: null,
       files: null,
       cellId: props.cellId,
@@ -212,9 +204,6 @@ export default class App extends React.Component<AppProps, AppState> {
       path: "",
     };
 
-    this.openImageInterval = null;
-    this.stateKey = "image";
-
     this.openImage = this.openImage.bind(this);
     this.loadFromRaw = this.loadFromRaw.bind(this);
     this.onChannelDataLoaded = this.onChannelDataLoaded.bind(this);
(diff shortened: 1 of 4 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -340,42 +310,20 @@ export default class App extends React.Component<AppProps, AppState> {
     return makeChannelIndexGrouping(channels, viewerChannelSettings);
   }
 
-  stopPollingForImage(): void {
```
toloudis, line 343: for some background/context:  In a very early implementation, this viewer used to be able to request an image, and have that image returned by a service that could on-demand convert the file to a downsampled version. (The downsampled version would then be cached on the server after that first conversion) But that processing could take time and so the app would continuously poll in case of errors,crashes etc.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -641,42 +582,6 @@ export default class App extends React.Component<AppProps, AppState> {
     }
   }
 
-  loadPrevImage(): void {
```
toloudis, line 644: hopefully this functionality can be re-implemented just by changing props from outside the component.
I believe an example of this is here:
https://imsc.allencell.org/?page=3d-viewer
(The "Stages" up/down arrows invoking next/prev image)

toloudis: I can't remember why we made it happen inside the viewer like this.  Maybe @meganrm remembers...!

toloudis: maybe it was so that the viewer internals could cache the loaded prev and next images, for faster switching?

frasercl: That looks to be correct - the `preLoad` prop seems to have controlled whether to grab surrounding images on startup, and these functions started prefetching images on switch. I guess implementing the equivalent outside the app would require exposing a `volume` prop or similar, and let the client take care of loading itself?

frasercl: Is the IMSC using website-3d-cell-viewer or base volume-viewer?

toloudis: my mistake, IMSC is directly embedding volume-viewer.  that was smart :)
 
 One initial thought I had was that website-3d-cell-viewer (or volume-viewer) could just keep a cache of Volume objects keyed by url (or LoadSpec = url + other params) and a prop could be how many Volumes to cache (`cacheSize=`).  It would evict least recently used ones when full and a new one is requested.   This would work both for time series animation playback in addition to this "next/prev" slideshow type of thing.   The key thing is that you'd have to have that single instance of the volume cache stay alive across many props changes as new Volumes are requested via website-3d-cell-viewer props.

toloudis (approved):

Issue #110, opened 2023-01-30
internals: design json format


toloudis:
## Use Case

Infrastructure for shareable links. As a developer, I want to see how all of the application state would be organized and documented, and review this with other developers.

## Solution

We will design and document a data structure that contains the following data:
* Camera
* Lights
* Volume LoadSpec
* All Volume display settings: advanced rendering settings and per-channel settings
* Viewer UI state (not literal react state)
* versioning for future-proofing against changes

The specification should be a typescript type that can be converted directly to and from json.  The data structure should capture as many of the currently known settings as possible, to help assess technical risk and prepare us for difficult areas of future implementation.  

We should also investigate a best known reasonable documentation format for documenting json schemas and/or typescript types, if there is such a thing.   Reasonable here is defined as: well supported, easy to use, easy to keep documentation up-to-date, and doesn't uglify the code too much

frasercl:
Basically solved by designing a URL query parameter format. (See also #111)

PR #119, opened 2023-03-01
Feature/reorganize viewer settings

frasercl:
## Problem

While writing a specification for how app state should be serialized and stored for #110, I found that `App` props and state were holding similar but not identical representations of the viewer configuration. Refactoring the app so that both representations are identical provides a clearer common interface for saving this state, solves a handful of state-management problems (of which differing representations were a symptom), and provides a more solid footing for refactoring `App` to a functional, hooks-based component.

## Solution

This PR contains a half-finished version of the specification for #110 in savedState.ts, marked with `TODO`s where I have yet to clear up some uncertainty. The bulk of these changes are to the representations of global viewer state in `App` props (the `viewerConfig` field, now removed) and state (the `UserSelectionState` type).

### Changes to `UserSelectionState`

- Renamed fields for `GlobalVolumeControls` sliders to match props, retyped single-element number arrays (`[number]`) to `number`s, and fixed these sliders writing stringified numbers to state
  - `alphaMaskSliderLevel` -> `maskAlpha`
  - `brightnessSliderLevel` -> `brightness`
  - `densitySliderLevel` -> `density`
  - `levelsSlider` -> `levels`
- Consolidated `maxProject` and `pathTrace` boolean flags as `renderMode` using `RenderMode` enum (booleans can be derived where needed)
- Lifted `controlPanelClosed` and `channelSettings` into top level of state
- Added `imageType` to props using the `ImageType` enum, with values `"cell"` or `"fov"`
- Made several changes to how clipping is handled:
  - `AppProps` previously had two optional fields: `region: [number, number, number, number, number, number]` or `slice: number`, which were theoretically mutually-exclusive ways to set clipping state but in reality weren't hooked up to anything. `UserSelectionState` now has field `region: { x: [number, number], y: [number, number], z: [number, number] }` which will set clipping state properly.
  - Clipping state was fully under the control of the `AxisClipSliders` component and expressed as integer slice indexes. I lifted this state into `App` using the format above, where the numbers now represent clipping values in the range [0, 1].
- Renamed `UserSelectionState` to `GlobalViewerSettings`, to hopefully better express its purpose

### Other changes to props interface

- Renamed fields: `viewerConfig` -> `viewerSettings`, `renderConfig` -> `showControls`. Several internal types are also renamed to match.
- All fields on `viewerSettings` and `showControls` are now optional, and will have defaults applied internally


src/aics-image-viewer/components/App/types.ts
```diff
@@ -1,8 +1,48 @@
...
+import { PerAxis } from "../../shared/types";
 import { ColorArray } from "../../shared/utils/colorRepresentations";
 import { ChannelGrouping, ChannelState, ViewerChannelSettings } from "../../shared/utils/viewerChannelSettings";
 
+type ControlNames =
+  | "alphaMask"
+  | "autoRotateButton"
+  | "axisClipSliders"
+  | "brightnessSlider"
+  | "colorPicker"
```
toloudis, line 12: what's "colorPicker"?  I can't think of what this is for

src/aics-image-viewer/components/App/types.ts
```diff
@@ -1,8 +1,48 @@
 import { View3d, Volume, ImageInfo } from "@aics/volume-viewer";
-import { ImageType, ViewMode } from "../../shared/enums";
+import { ImageType, RenderMode, ViewMode } from "../../shared/enums";
+import { PerAxis } from "../../shared/types";
 import { ColorArray } from "../../shared/utils/colorRepresentations";
 import { ChannelGrouping, ChannelState, ViewerChannelSettings } from "../../shared/utils/viewerChannelSettings";
 
+type ControlNames =
+  | "alphaMask"
```
toloudis, line 8: probably should be `alphaMaskSlider` for consistency? Or consider removing the suffix from all the others, to be agnostic to the control type?

src/aics-image-viewer/components/App/types.ts
frasercl: It's for the "customize" section in "advanced settings," which contains only color pickers. Could probably be split into `backgroundColorPicker` and `boundingBoxColorPicker`

src/aics-image-viewer/components/App/types.ts
```diff
@@ -1,8 +1,48 @@
...
+export interface GlobalViewerSettings {
+  viewMode: ViewMode;
+  renderMode: RenderMode;
+  imageType: ImageType;
+  showAxes: boolean;
+  showBoundingBox: boolean;
+  boundingBoxColor: ColorArray;
+  backgroundColor: ColorArray;
+  autorotate: boolean;
+  maskAlpha: number; // props.viewerConfig.maskAlpha || ALPHA_MASK_SLIDER_3D_DEFAULT,
```
toloudis, line 36: can ditch these comments at the end of the lines?

src/aics-image-viewer/shared/utils/sliderValuesToImageValues.ts
```diff
@@ -32,8 +32,8 @@ export const gammaSliderToImageValues = (sliderValues: [number, number, number])
 // Density and brightness are overloaded for the two supported rendering modes in the volume viewer.
 // These formulae are somewhat ad-hoc based on what will subjectively look reasonable in the viewer, and should be considered tweakable.
 // The volume-viewer expects to see numbers from 0..1 for density, brightness, and alpha.
-export const densitySliderToImageValue = (sliderValues: number[], _isPathTrace: boolean): number => +sliderValues[0] / 100.0;
+export const densitySliderToImageValue = (sliderValues: number, _isPT: boolean): number => +sliderValues / 100.0;
```
toloudis, line 35: i guess these were formerly arrays of number because of the nouislider returning its values as array because it can have multiple slider handles

src/aics-image-viewer/components/App/types.ts
frasercl: hmm... and it doesn't appear to be hooked up to anything anyways. Interesting.

src/aics-image-viewer/components/App/types.ts
frasercl: A lot of the symbolic constants they reference may be a bit obsolete too, and could instead be refactored a bit into a single default props object in constants.ts

toloudis:
Changes look fine but I have a couple of concerns generally, about extensibility and resilience to various unexpected future changes.

1. If we think of this component as a product that can be embedded by a variety of clients, the props are part of the public API.   Shouldn't the state data types (considered private and internal) be changeable without incurring changes to the public api (the props)?
2. For similar reasons, I'm still skeptical of the ability to serialize parts of the state object literally as-is, without a translation into a serialization-only data structure.   An example of this is an enum: You can never remove an entry, or reorder that enum once you have used it to save user settings out to files.  But you may well have a very good reason for the internal enum to have a value removed or to be re-ordered.

I still think these code changes are a nice cleanup, even if they lead to a future codebase where things are decoupled again.

toloudis (approved):

Issue #113, opened 2023-02-08
Milestone 5: Shareable Links and Channel Settings Redesign


toloudis:
## Shareable Links and Channel Settings Redesign

As a user, I would like to be able to send a link to my colleague so they can see the same data in the same view that I have set up.  
I would like to have a json file format to save viewer settings so that I can later edit (both by hand and with scripts) and copy such files and reload them into the viewer. 

- [x] #107 
- [x] #171

Channel settings redesign, reset button, optimization, and misc.

- [x] #191
- [x] #259 
- [ ] #120 -> deferred to next milestone
- [x] #80 
- [x] #131
- [ ] #174 -> initial attempt done but needs UX feedback, non-blocking for this milestone
- [x] #263 
- [x] #264 
- [x] #265 
- [x] #281

Optional:
- [x] #285

frasercl:
New issues to add:
- Refactor `App` to functional component, to get the state discipline benefits of `useEffect` and `useReducer` hooks
- Refactor `App` props to use load specification (similar/identical? to volume-viewer's `LoadSpec` type)

toloudis:
"Share" Button in UI
can we do this so that users don't have to save separate files? do we have any chance of compressing all of it into the url for real

ShrimpCryptid:
Notes from discussion:
- Include channel settings updates from Milestone 4 optional section?
- Reset to defaults button? #174

Issue #107, opened 2023-01-24
Shareable links


toloudis:
### Background

Users want to be able to share visualizations.  The viewer has lots of settings where users can customize the appearance of their data.   

### User needs statement(s)
As a user, I want to communicate a visualization that I have generated, to my colleagues.
As a user, I want to generate a link that I can send to someone else, and they will see what I see.

### Size
sm

### Requirements
User has a way to share their visualization with another user, so that the second user can open it seamlessly with the same settings.

### Deliverables
High fidelity design for generating and sharing the visualization

### Stakeholder(s)
@frasercl @toloudis

### Timeline
TBD; this is a high impact feature for our users but delivery is not time sensitive

### Notes
- We expect that a single url will not capture all the settings.
- A file is a natural way to store these settings

toloudis:
see #97

toloudis:
Adding another requirement from users:  if we generate a json file, they may make many copies of it and both manually and programmatically edit the files, and host them so they can be accessed via url.

toloudis:
"Share" Button in UI
can we do this so that users don't have to save separate files? do we have any chance of compressing all of it into the url for real

toloudis:
Share button could have one option for generating a URL and another option for saving a json.  UI then needs a way to load a json in addition to "share/saving" one.

lynwilhelm:
@toloudis Is this the developer implementation ticket? If so, I will remove it being sent to my ux tracking project. I like that UX tickets exist separate from an implementation ticket though.

toloudis:
Yes this one is for implementation now, because #171 exists

Issue #123, opened 2023-03-09
XY mode shows a blend of 2 slices instead of a single slice


toloudis:
## Description

When "interpolation" is on, the XY mode shows what appears to be a blend between slices instead of honing in on a single actual slice.  This may be related to the half-voxel offset we see when toggling interpolation in 3d mode.

## Expected Behavior

Expect interpolation to have no effect in XY mode...

## Reproduction

Toggle interpolation when in XY mode.  A way to verify with old CFE texture atlas data is to look at the PNG file from the Network tab in devtools. There you can clearly see the single slices.  Using segmentation channels may make this more obvious as the data will be less noisy. 

## Environment

Some day we will have to have a more simplified shader for XY mode.  (YZ and XZ may be more complicated as image data is generally organized by Z slices)

PR #126, opened 2023-03-22
Feature/functionize app

frasercl:
In order to properly set or change the state of the app, two things need to happen: UI state values need to be changed using React's API, and viewer state needs to be changed using volume-viewer's API. Since these procedures are currently relatively decoupled, every pathway to changing app state (including interacting with UI; changing props; or loading a new image, which may require settings to be re-propagated to volume-viewer) must individually consider which volume-viewer methods to call to effect the change they need to make. This can be tough to read and reason about, creates duplicated code, and may lead to unpredictable interactions between multiple updates. Since Milestone 3 (#113) stands to add a new state-changing pathway (#107), now is a reasonable time to consider alternate approaches.

Modern React, based on function components with hooks, encourages imperative updates that always occur in response to props or state updates to be made within the [`useEffect`](https://react.dev/reference/react/useEffect) hook. This hook accepts a callback and an array of dependent values, and re-runs the callback whenever one or more of those values changes between renders. In effect, this allows imperative updates to "react" to props/state changes the same way UI updates do. Using this pattern frees different state-changing mechanisms from having to worry about which volume-viewer methods to call, and makes the addition of new state-changing mechanisms easier and cleaner.

This PR rewrites (or attempts to rewrite) `App` as a function component which calls volume-viewer methods within `useEffect`.

### DIFF
--- a/src/aics-image-viewer/components/App/ChannelUpdater.tsx
+++ b/src/aics-image-viewer/components/App/ChannelUpdater.tsx
@@ -70,6 +70,9 @@ const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, vi
 
   useImageEffect(
     (image) => {
+      if (controlPoints.length < 2) {
+        return;
+      }
       const gradient = controlPointsToLut(controlPoints);
       image.setLut(index, gradient);
       view3d.updateLuts(image);
--- a/src/aics-image-viewer/components/App/ChannelUpdater.tsx
+++ b/src/aics-image-viewer/components/App/ChannelUpdater.tsx
@@ -8,27 +8,27 @@ import { UseImageEffectType } from "./types";
 
 interface ChannelUpdaterProps {
   index: number;
-  state: ChannelState;
+  channelState: ChannelState;
   view3d: View3d;
   image: Volume | null;
-  imageLoaded: boolean;
+  channelLoaded: boolean;
 }
 
 /**
  * A component that doesn't render anything, but reacts to the provided `ChannelState`
  * and keeps it in sync with the viewer.
  */
-const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, state, view3d, image, imageLoaded }) => {
+const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, view3d, image, channelLoaded }) => {
   const { volumeEnabled, isosurfaceEnabled, isovalue, colorizeEnabled, colorizeAlpha, opacity, color, controlPoints } =
-    state;
+    channelState;
 
-  // Effects to update channel settings should check if image is present and loaded first
+  // Effects to update channel settings should check if image is present and channel is loaded first
   const useImageEffect: UseImageEffectType = (effect, deps) => {
     useEffect(() => {
-      if (image && imageLoaded) {
+      if (image && channelLoaded) {
         return effect(image);
       }
-    }, [...deps, image, imageLoaded]);
+    }, [...deps, image, channelLoaded]);
   };
 
   useImageEffect(
--- a/src/aics-image-viewer/components/App/ChannelUpdater.tsx
+++ b/src/aics-image-viewer/components/App/ChannelUpdater.tsx
@@ -32,58 +32,64 @@ const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, vi
   };
 
   useImageEffect(
-    (image) => {
-      view3d.setVolumeChannelEnabled(image, index, volumeEnabled);
-      view3d.updateLuts(image);
+    (currentImage) => {
+      view3d.setVolumeChannelEnabled(currentImage, index, volumeEnabled);
+      view3d.updateLuts(currentImage);
     },
     [volumeEnabled]
   );
 
-  useImageEffect((image) => view3d.setVolumeChannelOptions(image, index, { isosurfaceEnabled }), [isosurfaceEnabled]);
+  useImageEffect(
+    (currentImage) => view3d.setVolumeChannelOptions(currentImage, index, { isosurfaceEnabled }),
+    [isosurfaceEnabled]
+  );
 
-  useImageEffect((image) => view3d.setVolumeChannelOptions(image, index, { isovalue }), [isovalue]);
-
-  useImageEffect((image) => view3d.setVolumeChannelOptions(image, index, { isosurfaceOpacity: opacity }), [opacity]);
+  useImageEffect((currentImage) => view3d.setVolumeChannelOptions(currentImage, index, { isovalue }), [isovalue]);
 
   useImageEffect(
-    (image) => {
-      view3d.setVolumeChannelOptions(image, index, { color });
-      view3d.updateLuts(image);
+    (currentImage) => view3d.setVolumeChannelOptions(currentImage, index, { isosurfaceOpacity: opacity }),
+    [opacity]
+  );
+
+  useImageEffect(
+    (currentImage) => {
+      view3d.setVolumeChannelOptions(currentImage, index, { color });
+      view3d.updateLuts(currentImage);
     },
     [color]
   );
 
   useImageEffect(
-    (image) => {
+    (currentImage) => {
       if (colorizeEnabled) {
         // TODO get the labelColors from the tf editor component
-        const lut = image.getHistogram(index).lutGenerator_labelColors();
-        image.setColorPalette(index, lut.lut);
-        image.setColorPaletteAlpha(index, colorizeAlpha);
+        const lut = currentImage.getHistogram(index).lutGenerator_labelColors();
+        currentImage.setColorPalette(index, lut.lut);
+        currentImage.setColorPaletteAlpha(index, colorizeAlpha);
       } else {
-        image.setColorPaletteAlpha(index, 0);
+        currentImage.setColorPaletteAlpha(index, 0);
       }
-      view3d.updateLuts(image);
+      view3d.updateLuts(currentImage);
     },
     [colorizeEnabled]
   );
 
   useImageEffect(
-    (image) => {
+    (currentImage) => {
       if (controlPoints.length < 2) {
         return;
       }
       const gradient = controlPointsToLut(controlPoints);
-      image.setLut(index, gradient);
-      view3d.updateLuts(image);
+      currentImage.setLut(index, gradient);
+      view3d.updateLuts(currentImage);
     },
     [controlPoints]
   );
 
   useImageEffect(
-    (image) => {
-      image.setColorPaletteAlpha(index, colorizeEnabled ? colorizeAlpha : 0);
-      view3d.updateLuts(image);
+    (currentImage) => {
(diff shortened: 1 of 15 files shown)


toloudis:
Should the target branch be main or is there some bug fixing still needed?

src/aics-image-viewer/components/App/ChannelUpdater.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const useImageEffect: UseImageEffectType = (effect, deps) => {
+    useEffect(() => {
+      if (image && channelLoaded) {
+        return effect(image);
+      }
+    }, [...deps, image, channelLoaded]);
+  };
+
+  useImageEffect(
+    (image) => {
```
toloudis, line 35: should we rename this to distinguish it from the component prop `image`?

src/aics-image-viewer/components/App/ChannelUpdater.tsx
```diff
@@ -0,0 +1,94 @@
...
+ * and keeps it in sync with the viewer.
+ */
+const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, view3d, image, channelLoaded }) => {
+  const { volumeEnabled, isosurfaceEnabled, isovalue, colorizeEnabled, colorizeAlpha, opacity, color, controlPoints } =
+    channelState;
+
+  // Effects to update channel settings should check if image is present and channel is loaded first
+  const useImageEffect: UseImageEffectType = (effect, deps) => {
+    useEffect(() => {
+      if (image && channelLoaded) {
```
toloudis, line 28: this `image` is the one that got passed in as a prop, right?  if so then I would either rename this one or the ones that are used in all the function args below.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -93,7 +94,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
     const numSlices = this.props.numSlices[axis];
     const clipVals = this.props.region[axis];
     const sliderVals = [Math.round(clipVals[0] * numSlices), Math.round(clipVals[1] * numSlices)];
-    const range = { min: 0, max: numSlices - 1 };
+    const range = { min: 0, max: numSlices - (twoD ? 1 : 0) };
```
toloudis, line 97: is this a behavior change?

toloudis: i think i see; it is compensated for, below

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -143,435 +134,227 @@ const defaultProps: AppProps = {
...
-      this.beginRequestImage();
-    }
-  }
+  // State management /////////////////////////////////////////////////////////
 
-  componentDidUpdate(prevProps: AppProps, prevState: AppState): void {
-    const { cellId, rawDims, rawData } = this.props;
-    const { channelSettings, view3d, image } = this.state;
+  // TODO is there a better API for values that never change?
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
toloudis, line 155: what is the unused var? is it _setView3d?  could you replace it with `_ /*setView3d*/` ?

frasercl: Yeah, it's `_setView3d`.

...come to think of it, there's truly no reason this line can't be `const [view3d] = useState(() => new View3d());`, my brain was just too locked into the pattern of deconstructing both values out of the returned array.

frasercl:
> Should the target branch be main or is there some bug fixing still needed?

I'm not aware of any major remaining bugs

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -143,435 +134,227 @@ const defaultProps: AppProps = {
...
-    const channelSetting = newChannelSettings || channelSettings;
-    view3d.removeAllVolumes();
-    view3d.addVolume(aimg, {
-      channels: aimg.channel_names.map((name) => {
-        const ch = this.getOneChannelSetting(name, channelSetting);
-        if (!ch) {
-          return {};
+      const activeAxis = activeAxisMap[viewMode];
+
+      // TODO the following behavior/logic is very specific to a particular application's needs
```
toloudis, line 217: it looks like this comment goes with the code up above here (line 199-200)

toloudis: oh i see some similar logic below too.  I agree with the comment - it seems pretty weird.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -143,435 +134,227 @@ const defaultProps: AppProps = {
...
-      color: thisChannelsSettings.color,
-      isosurfaceEnabled: isoenabled,
-      isovalue: thisChannelsSettings.isovalue,
-      isosurfaceOpacity: thisChannelsSettings.opacity,
-    });
-
+    keepLuts = false
+  ): void => {
     // if we want to keep the current control points
+    // TODO this function is never called with `keepLuts = true`. Should it ever be? On FOV switch e.g.?
```
toloudis, line 320: there are definitely scenarios where the viewer should be told to load fresh volume data but keep the per-channel settings the same (for example, playing through a time lapse).  However, I don't know how `keepLuts` would be populated in that scenario yet.  

There is probably some version of this codebase's history where keepLuts was set to true - my guess is it was associated with the "nextImg/prevImg" workflow.

src/aics-image-viewer/components/App/index.tsx
frasercl: I believe you wrote the comment actually

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -601,343 +384,148 @@ export default class App extends React.Component<AppProps, AppState> {
...
+  };
 
-  loadFromRaw(): void {
-    const { rawDims, rawData } = this.props;
-    if (!rawData || !rawDims) {
-      console.error("ERROR loadFromRaw called without rawData or rawDims being set");
-      return;
-    }
+  const setChannelStateForNewImage = (channelNames: string[]): ChannelState[] | undefined => {
+    setChannelGroupedByType(makeChannelIndexGrouping(channelNames, props.viewerChannelSettings));
```
toloudis, line 392: some day this will be easier to debug if 
```
const grouping = makeChannelIndexGrouping(...);
setChannelGroupedByType(grouping);
```

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -601,343 +384,148 @@ export default class App extends React.Component<AppProps, AppState> {
...
-    const imagePixelSize = pixelSize ? pixelSize.slice() : [1, 1, 1];
-    return (value: number) => {
-      const pixelSize = imagePixelSize.slice();
-      pixelSize[i] = value;
-      this.state.image?.setVoxelSize(pixelSize);
-    };
-  }
+    setImage(aimg);
+    setChannelGroupedByType(makeChannelIndexGrouping(rawDims.channel_names, props.viewerChannelSettings));
+    setChannelSettings(channelSetting);
```
toloudis, line 502: is this sequence of {placeImageInViewer, setImage, ..., setChannelSettings etc} repeated a couple times?

toloudis: for example, is setImage ever called without a call to placeImageInViewer?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -952,202 +540,277 @@ export default class App extends React.Component<AppProps, AppState> {
...
-  }
+    window.addEventListener("resize", onResizeDebounced);
+    return () => window.removeEventListener("resize", onResizeDebounced);
+  }, []);
 
-  toggleControlPanel(value: boolean): void {
-    if (this.props.onControlPanelToggle) {
-      this.props.onControlPanelToggle(value);
+  // Hook to trigger image load: on mount, when `cellId` changes, when `imageType` changes
+  useEffect(() => {
```
toloudis, line 591: is there a reason why this is separated from the other "on mount" function?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -952,202 +540,277 @@ export default class App extends React.Component<AppProps, AppState> {
...
+    [viewerSettings.levels]
+  );
+
+  useImageEffect(
+    (image) => view3d.setInterpolationEnabled(image, viewerSettings.interpolationEnabled),
+    [viewerSettings.interpolationEnabled]
+  );
+
+  useImageEffect(
+    (image) => view3d.setVolumeTranslation(image, props.transform?.translation || [0, 0, 0]),
```
toloudis, line 699: again, it makes me nervous when this `image` has the same name as the state `image`

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -952,202 +540,277 @@ export default class App extends React.Component<AppProps, AppState> {
...
+        </Content>
       </Layout>
-    );
-  }
-}
+    </Layout>
+  );
+};
+
+export default App;
```
toloudis, line 816: 300 fewer lines of code. 👍 
I still think about having a completely separate view3d interface class that does the translation between a bunch of our react state and the volume-viewer's api.  To try to keep the react coupling looser and put that responsibility into its own code module.  I have no idea if that would really make things better. This is looking good.

src/aics-image-viewer/components/App/index.tsx
toloudis: good point, that sounds best.

toloudis:
> > Should the target branch be main or is there some bug fixing still needed?
> 
> I'm not aware of any major remaining bugs

Great, then let's merge this right into main

src/aics-image-viewer/components/App/index.tsx
frasercl: It doesn't look to be, though this pattern only appears in two places for both ways of getting an image into the viewer (url and raw data), and those call sites don't have any state setters besides `setImage` in common. Nonetheless, I could see moving `setImage` into `placeImageInViewer` because you'll never not want to do both.

frasercl:
Worth noting that this refactor does introduce a small visual behavior change: since lut application is now done in an effect, and therefore after the first render, the volume will visibly "flash" a time or two on load as channels are initially rendered without a transfer function applied. If this is a problem, I can work on fixes.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: This did end up becoming a subtle behavior change, which I think boils down to the borders of the right sliders' range being increased by one. Before, both sliders went from 0 to n-1 and could touch; now the sliders go from 0 to n and must be at least one step apart. This way you can still grab and move the range when only a single slice is visible. (If I recall correctly, being able to move a single slice like that emphasized the need for a solution to #123 - the combination of two slices was extra visible in 3d where the slice could be viewed edge-on.)

toloudis (approved):

PR #132, opened 2023-04-10
add metadata and screenshot icons

### LINKED ISSUE #102: icon for metadata info panel

frasercl:
Resolve #102: add @thao-do's new icons for metadata panel button and screenshot button.

### Screenshot
image


thao-do:
@frasercl they look a little funky, let me double check the width and weight line

frasercl:
These now look like this on my screen:
image
I can't quite tell why the right edge of the download icon is getting cut off like that, will investigate

thao-do:
@frasercl here are 2 slight variations: one with minus 1 pixel on each side of the base (-m1), and the other one with minus 2 pixels on each side of the base (-m2). Please try with m1 first, then m2. 
[icon-download.zip](https://github.com/allen-cell-animated/website-3d-cell-viewer/files/11265332/icon-download.zip)

frasercl:
After some research and investigation, according to [this StackOverflow answer](https://stackoverflow.com/questions/54832394/a-border-of-1px-in-chrome-and-0-8px-in-firefox) the visual issues I was seeing appear to be related to [a Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=477157) which occurs on scaled displays. I was able to confirm that the icons look fine under other conditions. Given that these issues are showing up on a single browser under specific conditions, I think this is probably fine to merge as-is, though the narrower download icon may be a reasonable mitigation strategy for an icon that is particularly sensitive to having ~0.5px cut off a side.

toloudis:
> After some research and investigation, 

Thanks for looking into it deeply.  The bug you linked is really old (going back like 14 years!) and sadly is 
```
Status: REOPENED → RESOLVED
Closed: 3 years ago → 2 years ago
Resolution: --- → WONTFIX
```

toloudis (approved):

frasercl:
That said, the first narrower icon (m1) doesn't have this problem under the unusual circumstances in which it appears. I'll go ahead and swap that version in, unless @thao-do would prefer that the original revision remained given this new information.

thao-do (approved):
Please use m1 as you have stated. Thanks!

PR #139, opened 2023-05-03
Feature/time series props

frasercl:
Adds a field for time series step to props and state: `GlobalViewerSettings.time`. No UI controls this bit of state yet, but changing its value triggers the appropriate image to be loaded. Image loads triggered by only a time point change do not trigger a reset to 3D mode or recalculated transfer functions and mask alpha, as a completely new image would.

This addresses allen-cell-animated/volume-viewer#93, and may be enough to close it.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -405,7 +405,7 @@ const App: React.FC<AppProps> = (props) => {
     return newChannelSettings;
   };
 
-  const placeImageInViewer = (aimg: Volume, newChannelSettings?: ChannelState[]): void => {
+  const placeImageInViewer = (aimg: Volume, newChannelSettings?: ChannelState[], doResetMaskAlpha = true): void => {
     setImage(aimg);
 
     const channelSetting = newChannelSettings || channelSettings;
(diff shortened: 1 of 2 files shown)


frasercl:
TODO: bounds check against `image.imageInfo.times`?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -135,9 +136,10 @@ const defaultProps: AppProps = {
 };
 
 /** A `useState` that also creates a getter function for breaking through closures */
-function useStateWithGetter<T>(initialState: T): [T, (value: T) => void, () => T] {
+function useStateWithGetter<T>(initialState: T | (() => T)): [T, (value: T) => void, () => T] {
   const [state, setState] = useState(initialState);
-  const stateRef = useRef(initialState);
+  const isSetterFunction = (val: T | (() => T)): val is () => T => typeof val === "function";
+  const stateRef = useRef(isSetterFunction(initialState) ? initialState() : initialState);
```
toloudis, line 142: this useStateWithGetter function is a bit cryptic to me and lines like 141-142 hard to parse.  I don't know if there is a more readable formulation of line 141.  
Maybe a couple of more verbose comments saying what the intent is, and what is the difference between the `T` version and the `(()=>T)` version?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -427,30 +429,23 @@ const App: React.FC<AppProps> = (props) => {
...
-
   const openImage = async (): Promise<void> => {
     const { fovPath, cellPath, baseUrl } = props;
     const path = viewerSettings.imageType === ImageType.fullField ? fovPath : cellPath;
     const fullUrl = `${baseUrl}${path}`;
 
-    if (path === currentlyLoadedImagePath) {
+    // If this is the same image at a different time, keep luts. If same image at same time, don't bother reloading.
+    const samePath = path === currentImageLoadSpec?.subpath;
+    if (samePath && viewerSettings.time === currentImageLoadSpec?.time) {
```
toloudis, line 448: why is there a viewerSettings.time? Should it maybe be viewerSettings.loadSpec.time?  Is the idea of viewerSettings.time to store the value of the time slider?

frasercl: Right, `viewerSettings.time` is the property which would eventually be controlled by the time slider. My thought is that `viewerSettings` is basically the subset of `View3d` and `Volume` properties which are controlled by the UI and so need to be duplicated in React state.

src/aics-image-viewer/components/App/index.tsx
frasercl: Yeah I got weird with this one. On a second look there's something far simpler I could have done to support what I was trying to support. (Which is the convention that, if the argument to `useState` is a function, that function is assumed to hold an expensive calculation of the initial value of that state, rather than itself being the initial value of state.)

src/aics-image-viewer/components/App/index.tsx
toloudis: ok the only concern here is that we also have a current load spec which also specifies a "current time".

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: I replaced `currentlyLoadedImagePath` state (just a string) with the load spec used to fetch the current image, figuring that was the best way to support more closely examining the specific source of an image and detecting changes other than URL that still require reloading. There may be a better name than `currentImageLoadSpec` that makes this relatively limited purpose clearer?

toloudis: IMO the naming is just fine - the situation with what do we load and from where is just more complicated now. And we are currently set up to only load a single t . The load spec is supposed to encapsulate "everything you need to load the volume data": multiresolution level, time etc.

Issue #127, opened 2023-03-22
Time Series


lynwilhelm:
### Background
We want to support scientists that want to view time series data that's already slow with their existing tools. There is a premise that doing this with zarr data will be more performant. 

### User needs statement(s)
As a _[person inspecting microscopy images]_, I need/want to _[be able to load a time series at a particular time point/sample and bounce around to different times]_, so that I can _[quickly get a view of a huge file and find interesting phenomenon]_.

### Size
_Expected level of user research, discovery and design effort (xs, sm, m, lg, xl)_
S

### Requirements
- An input for typing in a time point
- A way for them to step incrementally forward and back
- Slider with single advance snap
- Need to see total time points and know which one they are currently on

- Ability to buffer to different time points will be later discussion

### Deliverables
- _High-fi designs + dev specifications_

### Stakeholder(s)
_Who is directly involved? (developers, main points of contact, final approver)_
Dan, Cameron

### Timeline
2 weeks

### Notes
_Additional notes here_

lynwilhelm:
Design complete. Design link: https://www.figma.com/file/TGor606RxgS2L8uVcHAEEb/Time-series?node-id=102%3A2315&t=s7LZF8MHRIaueJyT-1
Will initiate dev ticket creation in bi monthly meeting.

PR #152, opened 2023-05-24
Feature/time slider

frasercl:
Adds a time slider (which doesn't have an associated issue?)

- The slider only appears for volumes with multiple time points
- Apart from spinboxes and extra styling (to come later), the slider appears as specified in #127
- While the slider is being dragged, the numeric readout updates but the new time point is not loaded
- When the slider handle is released, the new time point is loaded

### Screenshot:
image

### DIFF
--- a/src/aics-image-viewer/components/AxisClipSliders/index.tsx
+++ b/src/aics-image-viewer/components/AxisClipSliders/index.tsx
@@ -78,8 +78,13 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
     this.createAxisSlider = this.createAxisSlider.bind(this);
   }
 
-  // Pause if mode or fov has changed
   componentDidUpdate(prevProps: AxisClipSlidersProps): void {
+    // Make extra sure the number next to the time slider is accurate
+    if (prevProps.time !== this.props.time) {
+      this.setState({ timeReadout: this.props.time });
+    }
+
+    // Pause if mode or fov has changed
     const numSlicesChanged = AXES.reduce(
       (hasChanged, axis) => hasChanged || prevProps.numSlices[axis] !== this.props.numSlices[axis],
       false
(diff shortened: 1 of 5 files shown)


src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -67,7 +111,7 @@ export default class AxisClipSliders extends React.Component<AxisClipSlidersProp
     const max = this.props.numSlices[activeAxis];
     const currentLeftSliderValue = Math.round(this.props.region[activeAxis][0] * max);
     const leftValue = (currentLeftSliderValue + delta + max) % max;
-    this.updateClipping(activeAxis, leftValue, leftValue);
+    this.updateClipping(activeAxis, leftValue, leftValue + 1);
```
toloudis, line 114: Confused about why this changed.  Is this resolving a bug in single slice mode?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -17,9 +17,11 @@ interface ViewerWrapperProps {
   loadingImage: boolean;
   viewMode: ViewMode;
   appHeight: string;
-  image: Volume | null;
+  hasImage: boolean;
```
toloudis, line 20: this is nice, a much less heavyweight dependency

toloudis (approved):
changes LGTM!  one minor question about the updateClipping calls.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: Yes, this seems to be something I missed when changing this component during the functional refactor. Used to be that the min and max were offset by 1 automatically. Now they're not, so moving the slice with the step/play buttons would flatten it down to nothing. (Ought to take a look now or between milestones at whether slider numbers still make as much sense as they ought to.)

PR #153, opened 2023-06-01
Feature/slider spinboxes

### LINKED ISSUE #134: Add spin boxes as inputs for clipping axis values
Add input boxes as spin boxes (with up/down arrows to increment/decrement) as in antd.
These will be used to control the axis slider values and should also cause slider updates to keep everything in sync.
For XYZT axis sliders only... Consider making a single control (combined slider+spinbox) to be reusable in future!

See #127 for visual spec.

frasercl:
Resolve #134: add spinboxes to allow manual entry of values in ROI/Time sliders. Remove now-redundant step buttons next to the play button in 2d mode.

I initially tried to add ant's [`InputNumber`](https://ant.design/components/input-number) component (as mentioned in the issue) and override its styles until it matched specification. But I eventually decided that this would likely take me as long as writing a new component, which would produce a cleaner and more legible result anyways. So, by observing `InputNumber` in the inspector and referencing a bit of its source (relevant file [here](https://github.com/react-component/input-number/blob/master/src/InputNumber.tsx)), I added the `NumericInput` shared component. `NumericInput` supports a reasonable subset of the original component's props, should be equivalently accessible, and behaves as specified by default without requiring a tangled mess of contradictory CSS.

### Screenshots

image
image

### DIFF
--- a/src/aics-image-viewer/components/AxisClipSliders/styles.css
+++ b/src/aics-image-viewer/components/AxisClipSliders/styles.css
@@ -90,7 +90,7 @@
     font-weight: bold;
   }
 
-  /* TODO this will need to be expanded (in 2d too!) */
+  /* TODO this will need to be expanded (here and in 2d!) */
   .slider-values {
     flex: 0 0 105px;
     margin-left: 18px;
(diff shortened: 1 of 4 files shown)


src/aics-image-viewer/components/AxisClipSliders/styles.css
```diff
@@ -7,8 +7,8 @@
   --color-z: #0099ff;
 
   @extend %axis-override;
-  /* .slider-group-title (50) + .axis-slider (480) + .slider-name (10) + .slider-slices (105) + margins & padding (80) */
-  max-width: 725px;
+  /* .slider-group-title (50) + .axis-slider (480) + .slider-name (10) + .slider-values (160) + margins & padding (80) */
+  max-width: 780px;
```
toloudis, line 11: These could be true expressions if all the other numbers were stored in vars, right?  Not that we should right now but something to consider for ease of maintenance.

toloudis (approved):
Looks great.

PR #156, opened 2023-06-14
Feature/viewer 3.2.0

frasercl:
Update volume-viewer to 3.2.0 and use its new APIs and features.

Since the last version, the API for volume loaders has changed, `View3d` gained a method for changing time in a time series, and a new time step indicator has been added.

#### Changes
- Update image loading to use the new `IVolumeLoader` API; stores created loaders in a ref for use when loading time series images
- Use the new `setTime` method to move through time (using the stored loader) without creating a new `Volume`
- Use the new time step indicator; add logic to move scale bars out of its way when present
- Remove the current image's `LoadSpec` as an element of state, since `Volume` objects now store their own `LoadSpec`s
- Vertically centers the clipping sliders (unrelated to the update, but it was a bit of the design I missed)
- Fix a bug where, if multiple loads of the same image were processing at the same time, only the first load to complete would trigger a channel update and subsequent ones would leave channels in a state that didn't match settings. This state was technically possible to create previously by switching from "Single cell" to "Full field" and back very fast, but became a clear problem when it became possible to quickly scrub through time. I fixed this by replacing the boolean `channelLoaded` prop on `ChannelUpdater` with a `version` number which increments on channel load, so that many concurrent loads can all create a new value for this prop when they arrive.

### Screenshot
image

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -45,7 +45,6 @@ import {
   INTERPOLATION_ENABLED_DEFAULT,
   AXIS_MARGIN_DEFAULT,
   SCALE_BAR_MARGIN_DEFAULT,
-  SCALE_BAR_TIME_SERIES_OFFSET,
 } from "../../shared/constants";
 
 import ChannelUpdater from "./ChannelUpdater";
@@ -164,12 +163,7 @@ const App: React.FC<AppProps> = (props) => {
     }
     return { x: 0, y: 0, z: 0 };
   };
-  const getNumberOfTimesteps = (): number => {
-    if (image) {
-      return image.imageInfo.times;
-    }
-    return 0;
-  };
+  const numberOfTimesteps = image?.imageInfo.times || 1;
 
   // State for image loading/reloading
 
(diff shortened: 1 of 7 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -534,37 +540,48 @@ const App: React.FC<AppProps> = (props) => {
 
   const resetCamera = useCallback((): void => view3d.resetCamera(), []);
 
+  const setIndicatorPositions = (panelOpen: boolean, hasTime: boolean): void => {
```
toloudis, line 543: since this is just a plain function, would it be slightly more efficient potentially to lift it up out of this component and pass in view3d?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -302,10 +304,14 @@ const App: React.FC<AppProps> = (props) => {
...
+    setChannelVersions(new Array(numberOfChannels).fill(0));
+  };
+
   const setOneChannelLoaded = (index: number): void => {
-    const newLoadedChannels = getLoadedChannels().slice();
-    newLoadedChannels[index] = true;
-    setLoadedChannels(newLoadedChannels);
+    const newVersions = getChannelVersions().slice();
+    newVersions[index]++;
+    setChannelVersions(newVersions);
```
toloudis, line 314: I expected to find some code that says "if the version that just arrived is behind the current version, then don't set this channel data".  Is that implicit somewhere or am I misunderstanding how version works here?

frasercl: `version` is really behaving like an extension of the boolean field it replaced here: it's still getting set to "false" (0) when loading begins, then "true" (1) when data arrives, but any more recent requests that began between those two events will still have the headroom to change the value (to... extra-true?) and trigger an update. It's not thinking at all about which request is more recent. We probably ought to track the order of requests, but I was nervous about giving this array of numbers that responsibility, since its length may change with number of channels

frasercl: Actually the better thing to do may be
- keep a request counter in a ref
- on requests, increment the counter and have the channel loaded callback close around the value when it's created
- when the callback runs, check the value against the ref
- this keeps channel loading/loaded state certain, so we can return to booleans here

src/aics-image-viewer/components/App/index.tsx
toloudis: I figure every time the react component's "function" is invoked, you are paying the cost of creating this new inner function every time

src/aics-image-viewer/components/App/index.tsx
toloudis: I think I get it.  I'm not sure we NEED to account for async requests arriving out of order, but it is definitely possible for that to happen.

toloudis (approved):
LGTM.
I think the way we handle the arrival of requests for pixel data is up for big discussion/refactor when we talk about caching and background requests.

PR #164, opened 2023-06-26
Deploy releases to external repository via gh-pages

### LINKED ISSUE #116: publish stable build in addition to nightly
## Use Case

External users would like a public url to access the standalone viewer.  However, the current public url is a nightly build which may be unstable from day to day.

## Solution

In addition to publishing a nightly, which is useful, we should also publish the latest versioned release to a public url, which we would consider "stable".

toloudis:
Resolves #116

1. Add some code to actually display current version numbers (for both the app front end and the underlying viewer) to console.
2. Add a github action to deploy versioned releases to website-3d-cell-viewer-release if they publish successfully.

Some changes to eslint just to simplify things.  "no-undef" was removed as a recommendation from the internet, because "Typescript basically does this for you"

So this leaves our current nightly build unchanged but on versioned tag releases we deploy to allen-cell-animated.github.io/website-3d-cell-viewer-release/


public/index.tsx
```diff
@@ -257,7 +262,7 @@ function runApp() {
     <ImageViewerApp
       cellId={args.cellid.toString()}
       baseUrl={args.baseurl}
-      appHeight="90vh"
+      appHeight="100vh"
       canvasMargin="0 120px 0 0"
```
frasercl, line 266: Any reason not to turn this into "0" and fill the screen horizontally to?

webpack.common.js
```diff
@@ -12,68 +12,75 @@ const themeVariables = lessToJs(
...
+      new webpack.ProvidePlugin({
+        THREE: "three",
+        jQuery: "jquery",
+        $: "jquery",
+      }),
+      new webpack.DefinePlugin({
+        WEBSITE3DCELLVIEWER_VERSION: JSON.stringify(require("./package.json").version),
+        VOLUMEVIEWER_VERSION: JSON.stringify(require("./node_modules/@aics/volume-viewer/package.json").version),
+        WEBSITE3DCELLVIEWER_BUILD_ENVIRONMENT: JSON.stringify(env.env),
+      }),
```
frasercl, line 44: Could this get merged with the define plugin above, on 31-33?

frasercl (approved):
Couple minor notes, but looks good overall

ShrimpCryptid (approved):
Tested locally and the debug messages show up. Looks good!

webpack.common.js
toloudis: oops yes of course.

toloudis: nothing in the repo seems to use APP_VERSION so I removed it

public/index.tsx
toloudis: I'll try this and if nothing obviously bad happens will push it.

PR #166, opened 2023-06-29
don't change mask alpha with view mode/FOV

### LINKED ISSUE #88: masking behaves unintuitively with full field
## Use Case

When user switches from single cell to full field, the cell mask alpha is still applied but the user does not expect the results they see.  
For example, why isn't only the single cell still masked?  
It is not clear to user that a named channel is being used as the mask in both cases.

## Solution

I think the viewer could have a "select mask channel name" in the UI and maybe an on/off switch in addition to the mask alpha value slider

### LINKED ISSUE #99: switching 3D to XY and back does not preserve mask alpha
## Description

Mask alpha value is being changed on me when I switch modes.

## Expected Behavior

Mask alpha should not change or at least change in a predictable and sensible way.

## Reproduction

Load viewer; mask alpha defaults to 50%.  
Set mask alpha to 100%.
Switch to XY mode.  Mask alpha moves to 0.
Switch back to 3D mode.  Mask alpha moves back to 50%!

frasercl:
Resolve #99, possibly resolve #88: per UX feedback, remove the behavior of resetting mask alpha when view mode or FOV changes.


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -215,15 +214,6 @@ const App: React.FC<AppProps> = (props) => {
   // To do this it requires access to the current state value, not the one it closed over)
   const [channelSettings, setChannelSettings, getChannelSettings] = useStateWithGetter<ChannelState[]>([]);
 
-  const getInitialAlphaLevel = (): number => {
-    if (props.viewerSettings?.maskAlpha !== undefined) {
-      return props.viewerSettings.maskAlpha;
```
toloudis, line 220: are we losing the ability to apply this setting now?

frasercl: `maskAlpha` will still respond to an initial value passed in via props, if that's the question. The point of this function was to derive a new value whenever something happened that reset mask alpha. But without it, the prop still gets applied in the `viewerSettings` initializer.

toloudis (approved):

PR #167, opened 2023-07-06
fix zarr loading from CFE

toloudis:
Most CFE datasets use a `baseUrl` and then combine it with a `cellPath` which is the individual image name. 
The use of `subpath` was doubling up the filename and causing bad urls when fetching zarr data.  `subpath` is really supposed to be used for multiresolution level and not much else. (A future change will have to address this naming)

This changeset just stops putting any information in subpath, and adjusts the standalone app inputs accordingly.  

This should fully enable CFE datasets to support zarr as the file format for 3d viewer!

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -452,7 +452,8 @@ const App: React.FC<AppProps> = (props) => {
 
     // If this is the same image at a different time, keep luts. If same image at same time, don't bother reloading.
     const currentLoadSpec = image?.loadSpec;
-    const samePath = path === currentLoadSpec?.subpath;
+    const samePath = fullUrl === currentLoadSpec?.url;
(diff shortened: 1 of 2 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -462,7 +462,7 @@ const App: React.FC<AppProps> = (props) => {
```
frasercl, line 462: Can't comment there, but should the expression on L455 above become `fullUrl === currentLoadSpec?.url`?

frasercl (approved):

src/aics-image-viewer/components/App/index.tsx
toloudis: Thank you! I will fix that.

ShrimpCryptid (approved):
Looks good! Had one quick question about a variable's use but that's it.

PR #170, opened 2023-09-29
Feature/viewer 3.3.0

frasercl:
Bump volume-viewer to version 3.3.0, update to the current API, and use new cache and z-slice features.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -175,12 +175,12 @@ const App: React.FC<AppProps> = (props) => {
 
   // TODO is there a better API for values that never change?
   const [view3d] = useState(() => new View3d());
-  const loaderRef = useRef<IVolumeLoader | null>(null);
   const [image, setImage] = useState<Volume | null>(null);
+  const imageUrlRef = useRef<string>("");
 
   const getNumberOfSlices = (): PerAxis<number> => {
     if (image) {
-      const { x, y, z } = image;
+      const { x, y, z } = image.imageInfo.volumeSize;
       return { x, y, z };
     }
     return { x: 0, y: 0, z: 0 };
@@ -340,7 +340,7 @@ const App: React.FC<AppProps> = (props) => {
     }
     view3d.updateLuts(aimg);
 
-    if (aimg.channelNames()[channelIndex] === props.viewerChannelSettings?.maskChannelName) {
+    if (aimg.channelNames[channelIndex] === props.viewerChannelSettings?.maskChannelName) {
       view3d.setVolumeChannelAsMask(aimg, channelIndex);
     }
 
@@ -424,7 +424,7 @@ const App: React.FC<AppProps> = (props) => {
     view3d.removeAllVolumes();
     view3d.addVolume(aimg, {
       // Immediately passing down channel parameters isn't strictly necessary, but keeps things looking saner on load
-      channels: aimg.channel_names.map((name) => {
+      channels: aimg.channelNames.map((name) => {
         const ch = getOneChannelSetting(name, channelSetting);
         if (!ch) {
           return {};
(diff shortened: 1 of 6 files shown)


frasercl:
I'm having trouble fixing some bugs where newly loaded data doesn't appear under certain conditions. In particular:

1. Switch to XY mode.
2. Move to a different timestep. **The slice does not appear.**
3. Move the Z slider. Slices start showing up correctly again.
4. Switch to 3D. **The volume does not appear.**
5. Switch to XY. The slice appears, at the resolution of the loaded 3d volume.
6. Switch back to 3D. Now the volume appears.

Any help or ideas?

toloudis:
> I'm having trouble fixing some bugs where newly loaded data doesn't appear under certain conditions. To replicate:
> 
> 1. Switch to XY mode.
> 2. Move the time slider. **The slice does not appear.**
> 3. Move the Z slider. Slices start showing up correctly again.
> 4. Switch to 3D. **The volume does not appear.**
> 5. Switch to XY. The slice appears, at the resolution of the loaded 3d volume.
> 6. Switch back to 3D. Now the volume appears.
> 
> Any help or ideas?

I just tried this:
When I first switch to XY mode the aspect ratio of the volume is off.  Then updating the time slider does eventually update the slice correctly, once the data has arrived, although it flashes through a wrongly-tiled version of the view.

toloudis:
Also I am noticing that the axis display in YZ mode is wrong.  (blue should be Z and is pointing the wrong way)

src/aics-image-viewer/shared/utils/hooks.ts
```diff
@@ -0,0 +1,25 @@
...
+export function useStateWithGetter<T>(initialState: T | (() => T)): [T, (value: T) => void, () => T] {
+  const [state, setState] = useState(initialState);
+  const stateRef = useRef(state);
+  const wrappedSetState = useCallback((value: T) => {
+    stateRef.current = value;
+    setState(value);
+  }, []);
+  const getState = useCallback(() => stateRef.current, []);
+  return [state, wrappedSetState, getState];
+}
```
ShrimpCryptid, line 13: I know you just moved this as part of this change, but this hook is really interesting to me. It's like, giving you all the instantly-updating behavior of a ref but still ensuring state updates happen normally. Neat!

ShrimpCryptid:
@frasercl the issues I'm seeing are consistent with what Dan is describing, where I don't have issues with slices/volumes not appearing, but I do see an incorrect aspect ratio.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -339,8 +331,9 @@ const App: React.FC<AppProps> = (props) => {
       changeChannelSetting(channelIndex, "controlPoints", newControlPoints);
     }
     view3d.updateLuts(aimg);
+    view3d.onVolumeData(aimg, [channelIndex]);
```
frasercl, line 334: The solution to all the problems I described earlier was to add this line. The test app in volume-viewer makes a similar call, and it seemed to be the only thing missing that could account for that app not having problems where this one did. I'm thrilled to have this fixed, but this seems like a slightly bizarre API choice? Feels like we're doing `view3d`'s work for it...

toloudis: The example viewer in volume-viewer seems to call this in its own onChannelDataArrived, so it's consistent with that, at least.

toloudis: It's a call to update the VolumeDrawable, and it does seem like we could connect that callback internally inside volume-viewer if we wanted to, and if the order of operations was correct.

toloudis:
I think I'm seeing a reproducible bug with channels being enabled/disabled:
Load time series data by url.
Switch to Z mode.
Turn off the green channel.
Increment T (wait for data to arrive)
Increment Z
The green channel came back even though it is still unchecked.

ShrimpCryptid:
Bug:
1. Load time series data by url: ?url=https://animatedcell-test-data.s3.us-west-2.amazonaws.com/20200323_F01_001/P13-C4.zarr
2. Switch to XY mode
3. Z shows middle Z-slice (20?) but ROI slider is still set to 0

ShrimpCryptid:
I'm also seeing disabled channels get reenabled, like Dan's bug report. It seems to happen if I go to a slice I already have cached, and then toggling the green channel on and off again disables it correctly.

ShrimpCryptid:
Still seeing the bug with channels getting re-enabled...

toloudis:
> Still seeing the bug with channels getting re-enabled...

I'm not seeing this on #175

frasercl:
> Do we want to merge #175 into this branch first, or merge this branch first into main? I've been holding off on fully reviewing it since I assumed the first. @frasercl

First seems right, particularly since that PR looks ready to go

PR #175, opened 2023-11-21
Camera mode in url params

toloudis:
Added a url parameter to set camera mode at initialization of the standalone viewer. 

In doing so, there is one behavior change: camera views will now remember their clipping slider settings, even when loading new images.

We also discovered something out of sync between the axis clipping sliders and the initial view state, so we changed the way the slider state is held by the app.  

Pair programmed with @frasercl

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -699,24 +699,25 @@ const App: React.FC<AppProps> = (props) => {
   const usePerAxisClippingUpdater = (axis: AxisName, [minval, maxval]: [number, number], slice: number): void => {
     useImageEffect(
       (currentImage) => {
-        // if (viewerSettings.viewMode === ViewMode.threeD) {
-        //   view3d.setAxisClip(currentImage, axis, -0.5, 0.5, false);
-        // } else {
-        //   const isOrthoAxis = activeAxisMap[viewerSettings.viewMode] === axis;
-        //   view3d.setAxisClip(currentImage, axis, minval - 0.5, maxval - 0.5, isOrthoAxis);
-        // }
-        //view3d.setCameraMode(viewerSettings.viewMode);
+        let isOrthoAxis = false;
+        let axismin = 0.0;
+        let axismax = 1.0;
         if (viewerSettings.viewMode === ViewMode.threeD) {
-          view3d.setAxisClip(currentImage, axis, minval - 0.5, maxval - 0.5, false);
+          axismin = minval;
+          axismax = maxval;
+          isOrthoAxis = false;
         } else {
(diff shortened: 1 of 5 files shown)


ShrimpCryptid:
I notice when setting the view mode in the URL that the 3D view flashes briefly (`http://localhost:9020/?view=Y`). Is that an issue, or out of scope for this change?

ShrimpCryptid (approved):
LGTM!
Also, `npm` recommends that an audit be run due to some dependency vulnerabilities.

toloudis:
Going to see if I can fix that initial draw in 3d mode when a different view was specified but we might defer it to a different PR.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -576,6 +544,11 @@ const App: React.FC<AppProps> = (props) => {
     return () => window.removeEventListener("resize", onResizeDebounced);
   }, []);
 
+  // one-time init after view3d exists and before we start loading images
+  useEffect(() => {
+    view3d.setCameraMode(viewerSettings.viewMode);
+  });
+
```
frasercl, line 551: This can also be accomplished by converting the regular `setCameraMode` hook to a `useImageLoadEffect`. Either approach is good, though here I believe you also need to give `useEffect` an empty dependency array to make it run once on load rather than on every render.

toloudis: oh yeah, I'll add the empty array.  I'd like to see this be the one place where other global init on view3d might happen as we expand the options.  Anyway, for now it's simple enough.

I thought useImageLoadEffect depended on there being an image. The problem this is fixing is that the view was staying in 3d until after the image data arrives.

frasercl: `useImageLoadEffect` (which is distinct from `useImageEffect`) will also run immediately on image load, without waiting for a React render cycle, to prevent flashes like this. The way you have it might still be a better approach for this case though, since this property isn't reset by image loads. Load effects are mainly properties like brightness, render mode, interpolation etc. which are owned by `VolumeDrawable` and need to be resynced when the current volume is replaced.

frasercl (approved):
Should be good to go!

ShrimpCryptid:
Re-requesting review from myself so I can poke at the view initialization code; I'm still seeing a jump as it switches between the volume and slice rendering modes, I think.

ShrimpCryptid:
> Re-requesting review from myself so I can poke at the view initialization code; I'm still seeing a jump as it switches between the volume and slice rendering modes, I think.

Looks like it's fixed, nevermind!

ShrimpCryptid (approved):

PR #179, opened 2024-01-09
Allow unloaded channels to be enabled

frasercl:
Quick fix: currently, channel settings are not updated on channels that are never loaded. When allen-cell-animated/volume-viewer#173 is published, disabled channels may never be loaded when the user moves to a new time point. Currently, this would cause these channels to be stuck in the disabled state. This change allows the enabled/disabled setting to be changed on unloaded channels.

### DIFF
--- a/src/aics-image-viewer/components/App/ChannelUpdater.tsx
+++ b/src/aics-image-viewer/components/App/ChannelUpdater.tsx
@@ -31,18 +31,19 @@ const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, vi
     }, [...deps, image, version]);
   };
 
-  useImageEffect(
-    (currentImage) => {
-      view3d.setVolumeChannelEnabled(currentImage, index, volumeEnabled);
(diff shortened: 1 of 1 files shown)


src/aics-image-viewer/components/App/ChannelUpdater.tsx
```diff
@@ -31,18 +31,19 @@ const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, vi
     }, [...deps, image, version]);
   };
 
-  useImageEffect(
```
ShrimpCryptid, line 34: Quick question because I'm not as familiar with this repo, why the change from `useImageEffect` to a `useEffect` dependent on `image`?

ShrimpCryptid: Okay, answered my own question. `useImageEffect` checks for changes to the image but only executes the body if the image is loaded.

ShrimpCryptid (approved):

toloudis (approved):

PR #180, opened 2024-01-31
Use workerized loading

frasercl:
The latest version of volume-viewer introduces the `VolumeLoaderContext`, which can offload volume loaders to a WebWorker where they won't block rendering and updates. This PR updates volume-viewer and replaces the current single-threaded loader creation with creation mediated by `VolumeLoaderContext`.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -2,15 +2,7 @@
 import React, { useCallback, useEffect, useRef, useState } from "react";
 import { Layout } from "antd";
 import { debounce } from "lodash";
-import {
-  createVolumeLoader,
-  LoadSpec,
-  RENDERMODE_PATHTRACE,
-  RENDERMODE_RAYMARCH,
-  View3d,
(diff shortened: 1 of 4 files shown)


ShrimpCryptid (approved):

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -161,9 +160,8 @@ const App: React.FC<AppProps> = (props) => {
 
   // State management /////////////////////////////////////////////////////////
 
-  // TODO is there a better API for values that never change?
   const view3d = useConstructor(() => new View3d());
-  const volumeCache = useConstructor(() => new VolumeCache(250_000_000 * 4));
+  const loadContext = useConstructor(() => new VolumeLoaderContext(250_000_000 * 4, 8, 3));
```
toloudis, line 164: turn these numbers into named symbolic constants. Also would be _ideal_ if this app just used the viewer defaults (ie. make these the defaults).

toloudis (approved):

PR #181, opened 2024-02-06
Time play button

### LINKED ISSUE #163: Play controls for time slider in XY Mode
## Use Case

Allow users to start playback in XY Mode

## Solution

- Imitate the Z-slider playback controls

frasercl:
Resolves #163: adds a play button to the time slider and improves playback usability.

Also, since I was in there and making extensive changes anyway, I refactored `AxisClipSliders` to a modern function component.

Most playback improvements are facilitated by a new class called `TimeControls`, which handles playback without having to depend on React's lifecycle. `TimeControls` and `AxisClipSliders` together enforce the following behavior:
- Only one axis may be playing at a time
- Playback may not move faster than we can load data
- Playback pauses on view mode change or when the user manually changes some clipping value
- ..._except_ if the user has grabbed the slider for the currently playing axis, in which case playback will temporarily hold until they release it

Other improvements to playback and slider behavior include:
- The loading spinner overlay does not appear during playback
- Slice sliders are aware of whether the entire extent of their dimension is loaded. If it is not, the slice updates only when the slider is released. (Only the time slider did this before.)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -170,14 +171,9 @@ const App: React.FC<AppProps> = (props) => {
...
-  const getNumberOfSlices = (): PerAxis<number> => {
-    if (image) {
-      const { x, y, z } = image.imageInfo.volumeSize;
-      return { x, y, z };
-    }
-    return { x: 0, y: 0, z: 0 };
-  };
-  const numberOfTimesteps = image?.imageInfo.times || 1;
+  const numSlices: PerAxis<number> = image?.imageInfo.volumeSize ?? { x: 0, y: 0, z: 0 };
+  const numSlicesLoaded: PerAxis<number> = image?.imageInfo.subregionSize ?? { x: 0, y: 0, z: 0 };
```
toloudis, line 175: in terms of Z slices, is this always either 1 or the total Z ?

frasercl: Currently yeah, this should in practice x and y should always be equal to `volumeSize`, and z should match `volumeSize` in 3d and be 1 in 2d

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -71,183 +88,184 @@ const SliderRow: React.FC<SliderRowProps> = ({ label, vals, valsReadout = vals,
   );
 };
 
-interface AxisClipSlidersProps {
+type PlaySliderRowProps = {
+  label: string;
+  val: number;
+  max: number;
+  playing: boolean;
+  entireAxisLoaded?: boolean;
```
toloudis, line 96: This var name made me pause.  I'm assuming this is only true when in 3D mode?  If so, it probably deserves a comment because I think it is rare that it could happen in any other way.  Maybe it could be named "canPlayWithoutAsync" or something similar...

frasercl: `updateWhileSliding`? Probably better to at least describe it in terms of what it changes about the slider's behavior. It's also true of the slice slider when looking down X or Y.

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -0,0 +1,80 @@
+import { AxisName } from "../types";
+
+type PlayAxisName = AxisName | "t";
+
+const PLAY_STEP_INTERVAL = 125;
+
+export default class PlayControls {
```
toloudis, line 7: 👍 I like any refactoring that separates app logic from UI.  It is interesting to consider whether this class could even live down in the volume-viewer.

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -0,0 +1,80 @@
+import { AxisName } from "../types";
+
+type PlayAxisName = AxisName | "t";
+
+const PLAY_STEP_INTERVAL = 125;
```
ShrimpCryptid, line 5: Nit: I think it would be worth adding units (milliseconds) to this constant name

```suggestion
const PLAY_STEP_INTERVAL_MS = 125;
```

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export default class PlayControls {
+  playingAxis: PlayAxisName | null = null;
+  playWaitingForLoad = false;
+  playHolding = false;
```
ShrimpCryptid, line 10: I like that there's an explicit holding state. In colorizer I'm tracking this outside my timing controller and pausing/playing,  which is a little messier.

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -0,0 +1,80 @@
...
+      return;
+    }
+    // If the volume is not loaded, wait for it to load before continuing
+    if (!this.getVolumeIsLoaded?.()) {
+      this.playWaitingForLoad = true;
+      return;
+    }
+
+    this.stepAxis(this.playingAxis);
+    this.playTimeoutId = window.setTimeout(this.playStep.bind(this), PLAY_STEP_INTERVAL);
```
ShrimpCryptid, line 33: curiosity question, what happens if you don't bind `playStep` here?

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -0,0 +1,80 @@
...
+      this.playWaitingForLoad = false;
+      this.playStep();
+    }
+  }
+
+  /**
+   * Pause playback on the currently playing axis.
+   * `willResume` marks this as a temporary suspension, e.g. while the user is scrubbing along the playing axis.
+   */
+  pause(willResume = false): void {
```
ShrimpCryptid, line 48: I like the language of `pause` and `stop`... maybe easier to read than adding the boolean flag here

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    this.setPlayingAxis(axis);
+    this.playStep();
+  }
+
+  /** If `axis` is currently playing, begin a temporary hold on playback while other input is pending. */
+  startHold(axis: PlayAxisName): void {
+    this.playHolding = true;
+    this.pause(axis === this.playingAxis);
+  }
```
ShrimpCryptid, line 71: So, just to check my understanding, playback will stop if the user interacts with a different axis?

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -71,183 +88,184 @@ const SliderRow: React.FC<SliderRowProps> = ({ label, vals, valsReadout = vals,
...
+        label={props.label}
+        vals={[props.val]}
+        valsReadout={props.entireAxisLoaded || !sliderHeld ? undefined : [valReadout]}
+        max={props.max}
+        onSlide={props.entireAxisLoaded ? wrappedOnChange : wrappedSetValReadout}
+        onChange={props.entireAxisLoaded ? undefined : wrappedOnChange}
+        onStart={wrappedOnStart}
+        onEnd={wrappedOnEnd}
+      />
+      <Tooltip placement="top" title="Play through sequence">
```
ShrimpCryptid, line 134:
```suggestion
      <Tooltip placement="top" title="Play through sequence" trigger=["hover", "focus"]>
```
Accessibility for tooltip, this allows tab navigation to also trigger it. (There are a bunch of instances in colorizer where I need to go back and fix this, eep!)

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -71,183 +88,184 @@ const SliderRow: React.FC<SliderRowProps> = ({ label, vals, valsReadout = vals,
...
 };
 
-interface AxisClipSlidersProps {
+type PlaySliderRowProps = {
+  label: string;
+  val: number;
+  max: number;
+  playing: boolean;
+  entireAxisLoaded?: boolean;
+  onPlayPause: (play: boolean) => void;
```
ShrimpCryptid, line 97: `onTogglePlayback` ?

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -71,183 +88,184 @@ const SliderRow: React.FC<SliderRowProps> = ({ label, vals, valsReadout = vals,
...
+        onSlide={props.entireAxisLoaded ? wrappedOnChange : wrappedSetValReadout}
+        onChange={props.entireAxisLoaded ? undefined : wrappedOnChange}
+        onStart={wrappedOnStart}
+        onEnd={wrappedOnEnd}
+      />
+      <Tooltip placement="top" title="Play through sequence">
+        <Button
+          className="slider-play-button"
+          onClick={() => props.onPlayPause(!props.playing)}
+          icon={props.playing ? "pause" : "caret-right"}
```
ShrimpCryptid, line 138: Ideally add an `aria-label` attribute for accesibility
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -71,183 +88,184 @@ const SliderRow: React.FC<SliderRowProps> = ({ label, vals, valsReadout = vals,
...
+        <PlaySliderRow
           label={axis.toUpperCase()}
-          vals={sliderVals}
-          max={numSlices - (twoD && numSlices > 1 ? 1 : 0)}
-          onSlide={this.makeSliderCallback(axis)}
+          val={Math.round(props.slices[axis] * numSlices)}
+          max={numSlices}
+          onChange={(val) => updateSlice(axis, val)}
+          onStart={() => props.playControls.startHold(axis)}
+          onEnd={() => props.playControls.endHold()}
```
ShrimpCryptid, line 212: I feel like `onStart` and `onEnd` on their own aren't super descriptive... maybe `onStartInteraction` would be better?

src/aics-image-viewer/shared/utils/playControls.ts
frasercl: Without `bind`, JavaScript helpfully binds the window object to `this` within `playStep`... so since the first thing `playStep` does is null-check some properties, and since those properties don't exist on `window`, playback just stops after one frame.

src/aics-image-viewer/shared/utils/playControls.ts
frasercl: Right - docs could potentially be a bit clearer on this

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: These are mainly named for parity with the noUiSlider events they eventually attach to: https://refreshless.com/nouislider/events-callbacks/

src/aics-image-viewer/shared/utils/playControls.ts
ShrimpCryptid: Ah, thank you, JavaScript!

(thanks for the explanation!)

toloudis: An equivalent alternative would be to make a simple arrow function, right?  setTimeout(()=>this.playStep()...

src/aics-image-viewer/shared/utils/playControls.ts
toloudis: I kinda wonder if renaming `willResume` would make this clearer, but the comment seems clear enough too.  Could even call it `isTemporary` (or the negation)

src/aics-image-viewer/components/AxisClipSliders/index.tsx
toloudis: a one line comment to say "these names map with nouislider event names" is a good idea.

toloudis (approved):
LGTM, I like the last renaming too.

src/aics-image-viewer/shared/utils/playControls.ts
```diff
@@ -20,6 +20,7 @@ export default class PlayControls {
   }
 
   private playStep(): void {
+    console.log(this);
```
ShrimpCryptid, line 23: Remove?

src/aics-image-viewer/components/AxisClipSliders/index.tsx
frasercl: The current (quite old) version of Ant used in this project doesn't recognize an array of triggers as a valid type for `trigger`, even though this change does produce the expected behavior? I've had to `@ts-ignore` it for now.

We have an ancient issue out to update to a newer Ant, but never made much progress because upgrading Ant breaks a lot of stuff in ways that are tedious to fix.

ShrimpCryptid (approved):
LGTM!

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -95,6 +98,12 @@ type PlaySliderRowProps = {
   playing: boolean;
   updateWhileSliding?: boolean;
   onTogglePlayback: (play: boolean) => void;
+  // These event handlers attach to the events of the same names provided by noUiSlider.
+  // Their behavior is documented at https://refreshless.com/nouislider/events-callbacks/
+  /**
+   * `onChange`'s behavior dependends on `updateWhileSliding`: if true, it's called on slide and on release;
```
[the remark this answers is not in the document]
frasercl: dependendendends

PR #184, opened 2024-02-13
Show timestep indicator on start

frasercl:
Because we never explicitly enabled the timestep indicator on startup for volumes with a time dimension, it would not show up until the user triggered an event which changed its enabled/disabled state. This one-line fix makes sure it's enabled to begin with.


toloudis (approved):
LGTM! Just one question: I assume it doesn't show if the time series has only one time, correct?

frasercl (commented):
> Just one question: I assume it doesn't show if the time series has only one time, correct?

Yup - that `hasTime` argument is always set from `imageInfo.times > 1`

ShrimpCryptid (approved):

PR #185, opened 2024-02-15
Don't update transfer function on reloads

frasercl:
### Problem

Currently, the transfer function is automatically updated every time new data is loaded into the volume. Scientists would prefer if it remained constant between time steps and z slices.

### Solution

Track when data is being loaded into the `Volume` for the first time, and only regenerate the transfer function on that load.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -179,6 +179,8 @@ const App: React.FC<AppProps> = (props) => {
 
   // State for image loading/reloading
 
+  // `true` when this is the initial load of an image
+  const initialLoadRef = useRef(true);
   // `true` when image data has been requested, but no data has been received yet
   const [sendingQueryRequest, setSendingQueryRequest] = useState(false);
   // `true` when all channels of the current image are loaded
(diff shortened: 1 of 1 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -293,20 +295,14 @@ const App: React.FC<AppProps> = (props) => {
...
-    } else {
-      // need to choose initial LUT
+  const onChannelDataLoaded = (aimg: Volume, thisChannelsSettings: ChannelState, channelIndex: number): void => {
+    // If this is the first load of this image, auto-generate initial LUTs
+    if (initialLoadRef.current || !thisChannelsSettings.controlPoints) {
       const newControlPoints = initializeLut(aimg, channelIndex, props.viewerChannelSettings);
       changeChannelSetting(channelIndex, "controlPoints", newControlPoints);
+    } else {
+      const lut = controlPointsToLut(thisChannelsSettings.controlPoints);
+      aimg.setLut(channelIndex, lut);
```
toloudis, line 305: i guess the old `keepLuts` was unused?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -179,6 +179,8 @@ const App: React.FC<AppProps> = (props) => {
 
   // State for image loading/reloading
 
+  // `true` when this is the initial load of an image
+  const initialLoadRef = useRef(true);
```
toloudis, line 183: is this a standard way to keep a one-shot variable?  Why a ref?  is it sort of like state that doesn't force a re-render?

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: Refs are different from state in two ways: they don't force a re-render, and they are always up-to-date within closures. `onChannelDataLoaded` is called from within a closure that is created alongside `image` and never changed thereafter, so a ref is the right tool for the job.

src/aics-image-viewer/components/App/index.tsx
frasercl: The value which was passed in as `keepLuts` was set outside the closure in which it was captured and used, so it always had a stale value (which was always `false`).

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -322,6 +318,7 @@ const App: React.FC<AppProps> = (props) => {
     if (aimg.isLoaded()) {
       view3d.updateActiveChannels(aimg);
       setImageLoaded(true);
+      initialLoadRef.current = false;
```
meganrm, line 321: it's surprising to me there isn't a setRef? instead of manually setting the `.current`

meganrm (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: The idea is that refs aren't values from which the rendered component is derived, so you don't need to let React know that you're updating them and that it should potentially re-render. I think their primary/original purpose is to attach to rendered DOM elements so that you can do imperative stuff to them outside the React cycle, though they're also legal to use in cases like this one, where the value influences some side effect of a future render without changing its result.

PR #186, opened 2024-02-16
Prioritize prefetching playing axis

frasercl:
volume-viewer added a concept of "prefetch priority" (#181) which enables forcing the zarr loader to prioritize a given axis or set of axes when prefetching. This feature was designed for playback in this viewer: when playing through time in z-slice mode, we should not be issuing prefetch requests for adjacent z slices, e.g. Update volume-viewer and connect axis playing state to prefetch priority.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -9,6 +9,9 @@ import {
   RENDERMODE_RAYMARCH,
   View3d,
   Volume,
+  IVolumeLoader,
+  WorkerLoader,
+  PrefetchDirection,
 } from "@aics/volume-viewer";
 
 import {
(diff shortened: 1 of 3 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -204,7 +214,10 @@ const App: React.FC<AppProps> = (props) => {
   // Playback state goes here, but the play/pause buttons that mainly control this class are down in `AxisClipSliders`
   const playControls = useConstructor(() => new PlayControls());
   const [playingAxis, setPlayingAxis] = useState<AxisName | "t" | null>(null);
-  playControls.onPlayingAxisChanged = setPlayingAxis;
+  playControls.onPlayingAxisChanged = (axis) => {
+    loader.current?.setPrefetchPriority(axis ? [axisToLoaderPriority[axis]] : []);
+    setPlayingAxis(axis);
+  };
```
toloudis, line 220: cool! do you see some performance gain from having prefetched more relevant data?

toloudis (approved):

ShrimpCryptid (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: Generally what I've seen is that playback runs at its (relatively low) framerate limit until it catches up with ongoing prefetching, then slows down dramatically. My hope for this change is that it will delay or prevent that slowdown point, though I haven't done a lot of work to test how well it accomplishes that.

PR #195, opened 2024-02-24
synchronize channel loading during playback only

toloudis:
Depends on api changes from https://github.com/allen-cell-animated/volume-viewer/pull/185
This calls a function to tell the volume loader to wait for all channels to arrive before updating the displayed volume, only during playback modes.  Fixes #190 after both pull requests go through.

After the dependent PR is approved, there will be a volume-viewer version bump added to this PR.   That will fix the CI breakage.

Estimated time to review: ~5 minutes

PR #199, opened 2024-03-27
Feature/lut refactor

toloudis:
Update with latest volume-viewer including automatic LuT remapping when playing through time or z. 
This reflects some public api changes around the handling of the LuTs.


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -220,6 +220,7 @@ const App: React.FC<AppProps> = (props) => {
   const [playingAxis, setPlayingAxis] = useState<AxisName | "t" | null>(null);
   playControls.onPlayingAxisChanged = (axis) => {
     loader.current?.setPrefetchPriority(axis ? [axisToLoaderPriority[axis]] : []);
+    loader.current?.syncMultichannelLoading(axis ? true : false);
```
toloudis, line 223: this line ensures that multichannel playback waits for all channels to arrive before redrawing.

ShrimpCryptid (approved):

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -314,8 +315,13 @@ const App: React.FC<AppProps> = (props) => {
       const newControlPoints = initializeLut(aimg, channelIndex, props.viewerChannelSettings);
       changeChannelSetting(channelIndex, "controlPoints", newControlPoints);
     } else {
-      const lut = controlPointsToLut(thisChannelsSettings.controlPoints);
-      aimg.setLut(channelIndex, lut);
+      // try not to update lut from here if we are in play mode
+      if (playingAxis !== null) {
+        // do nothing here?
+        // tell gui that we have updated control pts?
+        //changeChannelSetting(channelIndex, "controlPoints", aimg.getChannel(channelIndex).lut.controlPoints);
```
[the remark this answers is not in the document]
toloudis: Kinda. It's a reminder that this chunk of code has to be looked at again soon, as I think we have other conditions where we need to consider whether to update things.

src/aics-image-viewer/components/App/index.tsx
toloudis: also see #195

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -51,7 +51,14 @@ export function controlPointsToLut(controlPoints: ControlPoint[]): Uint8Array {
   ctx.fillStyle = grd;
   ctx.fillRect(0, 0, 256, 1);
   const imgData = ctx.getImageData(0, 0, 256, 1);
-  return new Uint8Array(imgData.data.buffer);
+
+  const lut = new Lut();
+  lut.lut = new Uint8Array(imgData.data.buffer);
+  lut.controlPoints = controlPoints;
+
+  // TODO Replace this whole function with new Lut().createFromControlPoints(controlPoints) ?
```
frasercl, line 59: Probably easy enough to run a test and compare the results of this function with `createFromControlPoints`. I think in general I'd like to see this happen in the name of less code duplication and because this function's reliance on drawing on a global `canvas` element makes me a bit nervous. (Though maybe it's more efficient than the `for` loops in `createFromControlPoints`? Could run a quick test on that too.) I'm just fine letting this PR through without this change though.

frasercl (commented):
Looks good! Excited to have this in!

frasercl (approved):

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
toloudis: Yeah my feeling is this probably should be done but on its own and later.

PR #207, opened 2024-04-03
Allow multiple data sources for zarr volumes

frasercl:
Review time: pretty quick (~20min)

Closes allen-cell-animated/volume-viewer#78: allow users to take advantage of volume-viewer's new ability to load zarrs from multiple data sources.

### Viewer component: replace the `cellPath`, `fovPath`, and `baseUrl` props with `imageUrl` and `parentImageUrl`.

Like the props they replace, `imageUrl` and `parentImageUrl` specify the paths from which to load image data. They correspond to the old `cellPath` and `fovPath` props, respectively, in that they specify what CFE calls "single cell" and "full field" images. However, they are different from the old set of data-loading props in the following ways:

- With `baseUrl` gone, **both URLs must be fully qualified**. No more automatically picking two files out of a single parent directory.
- **Both props may be `string[]`.** When passed down to volume-viewer to get a loader, a `string[]` will create a (zarr) loader which loads channels from multiple volumes. (NOTE: JSON loaders may also accept arrays of URLs, but will interpret each image as a different time point, not an additional source of channel data.)
- **`parentImageUrl` may be `undefined`.**

CFE will have to be updated to use these new props.

### Public app: remove `image` url parameter and add `url2`.

`image`, when present, behaved like the `cellPath` to `url`'s `baseUrl`. This was basically never used, since it was just as easy to specify the full url in `url`. So now it's gone.

In its place is `url2`, which adds a second zarr source (or JSON time step, I guess) to the loaded image. After some discussion about how we might support arbitrarily many source urls, we decided that just two sources would be good enough to support 95% of realistic use cases for this feature. (If anyone has any bright ideas about how to safely delimit a list of arbitrary URLs within another URL, let me know.)

## To test:

- `npm start`
- Open http://localhost:9020/?url=https://animatedcell-test-data.s3.us-west-2.amazonaws.com/variance/1.zarr/&url2=https://animatedcell-test-data.s3.us-west-2.amazonaws.com/variance/10.zarr/.
- You should now be able to mix and match channels from both zarr volumes.

### DIFF
--- a/public/index.tsx
+++ b/public/index.tsx
@@ -83,8 +83,8 @@ const args = {
   cellid: 2025,
   imageUrl: (BASE_URL + "AICS-22/AICS-22_8319_2025_atlas.json") as string | string[],
   parentImageUrl: BASE_URL + "AICS-22/AICS-22_8319_atlas.json",
-  fovDownloadHref: "https://files.allencell.org/api/2.0/file/download?collection=cellviewer-1-4/?id=F8319",
-  cellDownloadHref: "https://files.allencell.org/api/2.0/file/download?collection=cellviewer-1-4/?id=C2025",
+  parentImageDownloadHref: "https://files.allencell.org/api/2.0/file/download?collection=cellviewer-1-4/?id=F8319",
+  imageDownloadHref: "https://files.allencell.org/api/2.0/file/download?collection=cellviewer-1-4/?id=C2025",
   initialChannelSettings: VIEWER_3D_SETTINGS,
 };
 const viewerSettings: Partial<GlobalViewerSettings> = {
@@ -123,7 +123,7 @@ async function loadDataset(dataset: string, id: string) {
 
   const datasetData = await db.selectDataset(datasetMeta.manifest!);
   const baseUrl = datasetData.volumeViewerDataRoot + "/";
-  args.cellDownloadHref = datasetData.downloadRoot + "/" + id;
+  args.imageDownloadHref = datasetData.downloadRoot + "/" + id;
   // args.fovDownloadHref = datasetData.downloadRoot + "/" + id;
 
   const fileInfo = await db.getFileInfoByCellId(id);
@@ -193,9 +193,9 @@ if (params) {
     args.cellid = 1;
     args.imageUrl = imageUrl;
     // this is invalid for zarr?
-    args.cellDownloadHref = decodedUrl;
+    args.imageDownloadHref = decodedUrl;
     args.parentImageUrl = "";
-    args.fovDownloadHref = "";
+    args.parentImageDownloadHref = "";
     // if json, then use the CFE settings for now.
     // (See VIEWER_3D_SETTINGS)
     // otherwise turn the first 3 channels on and group them
@@ -222,8 +222,8 @@ if (params) {
     const baseUrl = "http://dev-aics-dtp-001.corp.alleninstitute.org/dan-data/";
     args.imageUrl = baseUrl + params.file;
     args.parentImageUrl = baseUrl + params.file;
-    args.fovDownloadHref = "";
-    args.cellDownloadHref = "";
+    args.parentImageDownloadHref = "";
+    args.imageDownloadHref = "";
     runApp();
   } else if (params.dataset && params.id) {
     // ?dataset=aics_hipsc_v2020.1&id=232265
@@ -243,8 +243,8 @@ function runApp() {
       parentImageUrl={args.parentImageUrl}
       appHeight="100vh"
       canvasMargin="0 0 0 0"
-      parentImageDownloadHref={args.fovDownloadHref}
-      imageDownloadHref={args.cellDownloadHref}
+      parentImageDownloadHref={args.parentImageDownloadHref}
+      imageDownloadHref={args.imageDownloadHref}
       viewerSettings={viewerSettings}
       viewerChannelSettings={args.initialChannelSettings}
     />,
--- a/public/index.tsx
+++ b/public/index.tsx
@@ -59,7 +59,7 @@ export const VIEWER_3D_SETTINGS: ViewerChannelSettings = {
   maskChannelName: "SEG_Memb",
 };
 
-type ParamKeys = "mask" | "ch" | "luts" | "colors" | "image" | "url" | "file" | "dataset" | "id" | "view";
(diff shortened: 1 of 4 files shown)


public/index.tsx
```diff
@@ -149,61 +186,20 @@ if (params) {
...
+    const imageUrl = decodedUrl2 ? [decodedUrl, decodedUrl2] : decodedUrl;
 
     args.cellid = 1;
-    args.baseurl = decodedurl;
-    args.cellPath = decodedimage;
+    args.imageUrl = imageUrl;
     // this is invalid for zarr?
-    args.cellDownloadHref = decodedurl + decodedimage;
-    args.fovPath = "";
+    args.cellDownloadHref = decodedUrl;
```
toloudis, line 196: rename to imageDownloadHref and parentImageDownloadHref?

toloudis, line 196: rename to imageDownloadHref and parentImageDownloadHref?

toloudis (commented):
On the topic of delimited lists of urls, your list is already delimited by`&` so a comma delimited list in one single url query parameter is not so crazy.

public/index.tsx
```diff
@@ -73,12 +73,16 @@ function parseQueryString(): Params {
...
+};
+
+const BASE_URL = "https://s3-us-west-2.amazonaws.com/bisque.allencell.org/v1.4.0/Cell-Viewer_Thumbnails/";
 const args = {
-  //baseurl: "http://dev-aics-dtp-001.corp.alleninstitute.org/cellviewer-1-4-0/Cell-Viewer_Thumbnails/",
-  baseurl: "https://s3-us-west-2.amazonaws.com/bisque.allencell.org/v1.4.0/Cell-Viewer_Thumbnails/",
   cellid: 2025,
-  cellPath: "AICS-22/AICS-22_8319_2025_atlas.json",
-  fovPath: "AICS-22/AICS-22_8319_atlas.json",
+  imageUrl: (BASE_URL + "AICS-22/AICS-22_8319_2025_atlas.json") as string | string[],
```
ShrimpCryptid, line 84: Why do you need this cast here?

ShrimpCryptid (commented):
Left a couple questions for you!

toloudis (approved):

public/index.tsx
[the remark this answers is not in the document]
frasercl: It's close enough to `AppProps` to give it a type!

public/index.tsx
```diff
@@ -97,6 +101,39 @@ const viewerSettings: Partial<GlobalViewerSettings> = {
...
 };
 
+async function loadDataset(dataset: string, id: string) {
+  const db = new FirebaseRequest();
+
+  const datasets = await db.getAvailableDatasets();
+
+  let datasetMeta: DatasetMetaData | undefined = undefined;
+  for (const d of datasets) {
+    const innerDatasets = d.datasets!;
```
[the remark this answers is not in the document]
frasercl: It appears as though the type here (`DatasetMetaData`) is recursive in this `datasets` field (`datasets` has type `{ [key: string]: DatasetMetaData }`). it may be that we expect exactly one level of nesting via this property, and the author of the type didn't want to create a separate type to do that? I'm not sure exactly though, but this assertion was there before.

frasercl:
Most recent change modifies the API described above: `url2` is gone; now multiple source URLs may be passed in as a comma-delimited list in `url`, ideally sanitized with `encodeURIComponent`. This seems to be the actual ideal way to encode multiple URLs in a single URL. We should make sure to be diligent in the future about using `encodeURIComponent` in URL sharing dialogs or auto-updating URLs.

public/index.tsx
```diff
@@ -73,15 +73,39 @@ function parseQueryString(): Params {
...
-  initialChannelSettings: VIEWER_3D_SETTINGS,
+const decodeUrl = (url: string): string => {
+  const decodedUrl = decodeURIComponent(url);
+  return decodedUrl.endsWith("/") ? decodedUrl.slice(0, -1) : decodedUrl;
+};
+
+/** Tries to parse a string as a list of URLs */
+const tryDecodeURLList = (url: string, delim = ","): string[] | undefined => {
+  if (!url.includes(delim)) {
+    return undefined;
```
rugeli, line 84: Nit: Just a thought - rather than returning `undefined` in cases of invalid input, might we consider returning an empty array instead? This would make the behavior more predictable and debugging more straightforward.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -584,7 +584,7 @@ const App: React.FC<AppProps> = (props) => {
     view3d.setCameraMode(viewerSettings.viewMode);
   }, []);
 
-  // Hook to trigger image load: on mount, when image source props/state change (`cellId`, `imageType`, `time`)
+  // Hook to trigger image load: on mount, when image source props/state change (`cellId`, `imageType`, data)
```
rugeli, line 587: add `` around "data"?

rugeli (approved):
Overall looks good! Since I'm not deeply familiar with the codebase, please feel free to disregard my comment if it doesn't align with the established patterns.

public/index.tsx
frasercl: The idea of this function is to either parse a list of two or more URLs out of a `string`, or to determine that the `string` is not a well-formatted URL list (likely because it's just a single URL, though maybe because it's incorrectly formatted). Returning `undefined` in that second case lets me concisely fall back to parsing a single URL below, on line 209:

`const imageUrls = tryDecodeURLList(params.url) ?? decodeUrl(params.url);`

Basically, "get a URL list out of this string, or if you can't do that, format out a single URL." I've updated the doc comment to be a bit more specific about the function's purpose.

src/aics-image-viewer/components/App/index.tsx
frasercl: This was meant to capture both `rawData` and `rawDims` - changes to either indicate the presence of new *data* needs to be loaded into the viewer (though they'd likely both change at the same time). But you're right that it's confusing - updated.

public/index.tsx
toloudis: For the sake of discussion, the function could also return `[url]` if there's no comma in it! (Hint: I have already approved this pr)

Issue #188, opened 2024-02-20
landing page


toloudis:
as designed.

Link to design:
https://www.figma.com/file/EJCCnElvRvObznBhN519sl/Components---CFE-%2B-3DWV?type=design&node-id=215%3A546&mode=design&t=L8nMtmh7C0xfVPfJ-1

toloudis:
see if we can include opengraph tags

lynwilhelm:
@ShrimpCryptid Link to designs added above

ShrimpCryptid:
TODO thought dump of stuff that still needs to be done:
- [ ] Copy in the code from the colorizer
- [ ] Add styling variables to match with designs, and clean up/update the CSS variables already in the project
- [x] Add background video for the volume viewer.
  - [x] Host the video somewhere else, or in the project? (it's 15 MB...)
  - [x] Make it responsive to `prefers-reduced-motion`
- [ ] Update fonts to Helvetica + Open Sans to match the designs
- [ ] Loading volumes from landing page load buttons... discuss w/ Cameron how files should be loaded, will likely be affected by current PRs (https://github.com/allen-cell-animated/website-3d-cell-viewer/pull/207)

PR #208, opened 2024-04-04
feature: Landing page initial pass

ShrimpCryptid:
*Estimated review size: medium, 20-30 minutes.\**

(\*Note: This is almost exactly the same as the original banner PR in nucmorph colorizer, which is why I've requested you both specifically! This will hopefully be faster to review since it's familiar code.)

Initial (UNSTYLED) layouts for #188, the landing page.

- Adds styled-components and icon library dependencies (Dan + Cameron + I had a long discussion about whether or not to use it, and decided that it was OK since it wouldn't need to be included with the viewer itself).
- Copies existing landing page code from the nucmorph-colorizer.
  - CSS variables are defined differently in nucmorph-colorizer, so most of the styling is currently broken. This will be fixed in a follow-up PR, but I want to split out what I have for ease of review.
- Adds a `types.ts` for the landing page to define expected format of the landing page content.

Current landing page:
image


src/landing-page/components/LandingPage/utils.tsx
```diff
@@ -0,0 +1,64 @@
+import styled, { css } from "styled-components";
```
ShrimpCryptid, line 1: This file's contents copied from https://github.com/allen-cell-animated/nucmorph-colorizer/blob/main/src/styles/utils.tsx.

src/landing-page/components/LandingPage/index.tsx
```diff
@@ -0,0 +1,301 @@
+import { faUpRightFromSquare } from "@fortawesome/free-solid-svg-icons";
```
ShrimpCryptid, line 1: Compare with https://github.com/allen-cell-animated/nucmorph-colorizer/blob/main/src/routes/LandingPage.tsx if desired. This change is directly copied from it, with a few minor changes:
- Removed Link components on the "Load" buttons since it's currently unclear how load actions will work.
- Updates the feature highlight copy and title description based on designs.
- Commented out header component; currently unimplemented.

src/landing-page/types.ts
ShrimpCryptid, line 1: See https://github.com/allen-cell-animated/nucmorph-colorizer/blob/main/src/types/index.tsx.

src/landing-page/components/LandingPage/index.tsx
```diff
@@ -0,0 +1,301 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { Button, Tooltip } from "antd";
+import React, { ReactElement } from "react";
+
+import { landingPageContent } from "./content";
+import { DatasetEntry, ProjectEntry, ViewerArgs } from "../../types";
+import styled from "styled-components";
+import { FlexColumnAlignCenter, FlexColumn, FlexRowAlignCenter, VisuallyHidden, FlexRow } from "./utils";
+
+const Banner = styled(FlexColumnAlignCenter)`
```
meganrm, line 11: can you do a dev chat on the pros /cons of styled-components?

src/landing-page/components/LandingPage/index.tsx
```diff
@@ -0,0 +1,301 @@
...
+  width: 100%;
+  flex-direction: column;
+  gap: 12px;
+
+  & h3 {
+    font-weight: 600;
+  }
+`;
+
+const DatasetList = styled.ol`
```
meganrm, line 125: why are you using an ordered list here?

meganrm (approved):
seems good to me

src/landing-page/components/LandingPage/index.tsx
ShrimpCryptid: Looking into this, I think an unordered list makes more sense, yeah. 

This is what the lists of datasets look like in the colorizer tool: 
image

src/landing-page/components/LandingPage/index.tsx
```diff
@@ -0,0 +1,301 @@
...
+};
+
+export default function LandingPage(props: LandingPageProps): ReactElement {
+  // Rendering
+
+  // TODO: Should the load buttons be link elements or buttons?
+  // Currently both the link and the button inside can be tab-selected.
+  const renderDataset = (dataset: DatasetEntry, index: number): ReactElement => {
+    return (
+      <DatasetCard>
```
frasercl, line 185: `index` is unused here, but this function is used in a `map` and the component it renders may want a `key` property. Looks like `index` is used for that in the equivalent file in nucmorph-colorizer - should it be used here too?

src/landing-page/types.ts
```diff
@@ -0,0 +1,28 @@
...
+// TBD what URL parameters to include here
+export type ViewerArgs = {
+  baseurl: string;
+  cellid: number;
+  cellPath: string;
+  fovPath: string;
+  fovDownloadHref: string;
+  cellDownloadHref: string;
+  viewerSettings: ViewerChannelSettings;
+};
```
frasercl, line 12: Based on your suggestion, in #207 I retyped a similar data structure (the one you may have based this type on?) as `AppProps` within an `Omit`. This is just a note - I think it would be perfectly fine to merge this as-is, given that (as you note in the comment above) shaking down the structure and handling of URL parameters is a broader issue that can be addressed separately.

src/landing-page/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  loadParams: ViewerArgs;
+};
+
+export type ProjectEntry = {
+  name: string;
+  description: string;
+  publicationLink?: URL;
+  publicationName?: string;
+  loadParams?: ViewerArgs;
+  datasets?: DatasetEntry[];
```
frasercl, line 26: Clarifying question: it's possible for a project to both have loadable data itself and have a list of datasets with loadable data? (I think this is fine if so, not implying that this type needs changing)

frasercl (approved):
Only one of my comments might need anything done about it, and it's simple enough for me to approve preemptively.

src/landing-page/types.ts
ShrimpCryptid: Yes, correct!

src/landing-page/components/LandingPage/index.tsx
ShrimpCryptid: That makes sense, thank you!

PR #214, opened 2024-04-12
feature: Open Sans font files

ShrimpCryptid:
*Estimated review size: small, 10 minutes*

Removes the Merriweather Sans and Overpass font files and replaces them with Open Sans. This change ensures that the viewer looks the same even when embedded!

We are importing Open Sans from Google Fonts, which means that it should work in almost all browser contexts. For full compatibility with older browsers, however, we need to include the WOFF, WOFF2, TTF, EOT, and SVG file formats. (https://css-tricks.com/snippets/css/using-font-face-in-css/#h-deepest-possible-browser-support)

Google Fonts will also not work in contexts without internet access, so having these fonts embedded in the project ensures that the viewer will look consistent.

| Google Font | Font Files |
| --- | --- |
| image | image |

There are some super minor differences between the fonts included (500 vs. 600 weight on the semibold) but otherwise they should look identical.


src/aics-image-viewer/assets/styles/typography.css
ShrimpCryptid, line 1: Added additional fallbacks for other languages and emojis. I don't expect these other fonts to be used, but it's good to include just in case!

ShrimpCryptid: Fallback fonts copied from the parent app, which I think is currently provided by Ant.

image

meganrm (approved):

blairlyons (approved):

PR #217, opened 2024-04-16
update rawData interface

toloudis:
Time to review: ~10-15 min?  

Volume-viewer has a new RawArrayLoader so that raw array data can be passed through the same IVolumeLoader code path as files.

Currently https://github.com/allen-cell-animated/nbvv is the only client that passes raw array data in. 

This PR updates the viewer props interface to use the new code path, and exports the appropriate data structures so that `nbvv` is ready to use it.

There's one key change in the loading code that initiates the volume data request only after the front end has had a chance to set up React state for channel settings.  This is because the RawArrayLoader is completely synchronous and fires the channel data callback immediately.


ShrimpCryptid (approved):
Definitely feels a lot cleaner!

blairlyons (approved):

Issue #81, opened 2022-10-03
modify styling to align with ant's system


frasercl:
## Use Case

The viewer is currently dependent on many less vars being set to specific values to look "correct." Importers of this component have to include a vendored copy of ant-vars.less and configure it to build with ant's styles, which should not be their responsibility. This also points to a more general problem we have with fighting against the default styles of ant components, rather than using ant's mechanisms to set a more sensible default style and overriding that in more limited ways. If resolved, this would also clear the way to allowing multiple themes (e.g. light/dark) and more theme customization options.

## Solution

- Convert our styles to less to integrate better with ant's styling system.
- Wherever possible, enforce our general style preferences as defaults in ant variables.
- Clean up and remove custom styles made redundant by the previous step.
- Potentially define our own reduced set of theme variables to tweak specific colors and other appearance details specific to this app.
  - Document these if so.
- Create one or more theme files that users of this component can import, rather than vendoring our preferred variable settings.
- Modify build process to ensure that this file(s) make it into the published npm package.

## Alternatives

We could also keep our styles in their current format and use ant's [dynamic theme option](https://ant.design/docs/react/customize-theme-variable) to allow the use of native, runtime-modifiable CSS variables. This would first require an upgrade to antd 4 (see #42).

frasercl:
Closed by #219. Could be revisited as part of our never-ending march to keep up with changing styling standards, or to implement specific additional features like themes (light/dark, etc.).

PR #219, opened 2024-04-17
refactor: Upgrade Ant version

ShrimpCryptid:
Upgrades the Ant Design component library to 5.16.2 from 3.26.20 and fixes (most of) the elements that broke as a result.

*Estimated review size: M/L, 30-40 minutes.*

- Updates Ant Design to 5.16.2. 
- Changes `StyleProvider` to use a unified `theme` variable that is passed down through CSS variables and the Ant ConfigProvider.
- Removes less files and plugins.
- Fixes various CSS/design problems that arose as part of the change. (see list below)
- Updates Ant components that had changed APIs.

**Video talk-through (🔊):**

https://github.com/allen-cell-animated/website-3d-cell-viewer/assets/30200665/b06c35cf-0b7b-437d-a7e0-c6eca979dbc0

Questions/Discussion:
- Are there better ways we can be handling CSS variables and styling?
- See note here about modifying our styling to align with Ant: #81 

Thanks for contributing!


ShrimpCryptid:
TODO list:

- [ ] Fix underlines on links so they only appear when focused
- [ ] Fix card spacing on sections headers
- [ ] Fix underline color on title
- [ ] Button hover and selection colors
- [ ] Fix H2 text colors (should be white)
- [ ] Fix font size on main page
- [ ] Fix disabled buttons having hover color

src/aics-image-viewer/components/StyleProvider/styles.css
ShrimpCryptid, line 1: Moved into `index.tsx` via StyledComponents

public/index.tsx
```diff
@@ -26,6 +26,7 @@ const routes = [
   {
     path: "viewer",
     element: <AppWrapper />,
+    errorElement: <ErrorPage />,
```
ShrimpCryptid, line 29: Added the ErrorPage to help with error handling in the App. This will catch any errors that aren't directly handled by the image viewer component.

src/aics-image-viewer/assets/styles/variables.css
```diff
@@ -3,7 +3,7 @@
 /* $app-height: 100%; */
 /* $dark-gray: #313131; */
 /* $purple: #8048f3; */
-$highlight-purple: #9e6aff;
+$highlight-purple: #aa88ed;
```
ShrimpCryptid, line 6: Kind of a cheat here. Highlight purple was being used for button hovering, so I replaced it with the new light purple from the palette. This eventually needs to be replaced.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -762,97 +763,99 @@ const App: React.FC<AppProps> = (props) => {
...
-          channelSettings={channelSettings}
-          showBoundingBox={viewerSettings.showBoundingBox}
-          backgroundColor={viewerSettings.backgroundColor}
-          boundingBoxColor={viewerSettings.boundingBoxColor}
-          maskAlpha={viewerSettings.maskAlpha}
-          brightness={viewerSettings.brightness}
-          density={viewerSettings.density}
-          levels={viewerSettings.levels}
-          interpolationEnabled={viewerSettings.interpolationEnabled}
+    <StyleProvider>
```
ShrimpCryptid, line 766: Recommend turning off whitespace when viewing this file. The only change was to add the `StyleProvider` component around the contents of App, so that it looks the same when embedded or run within the website.

Note that this does mean that `styled-components` is now being used within the viewer component itself.

src/aics-image-viewer/components/ControlPanel/styles.css
```diff
@@ -95,18 +96,6 @@
...
-  }
-
-  .ant-checkbox-indeterminate .ant-checkbox-inner {
-    background-color: transparent;
-
-    :after {
-      background-color: white;
-    }
-  }
-
```
ShrimpCryptid, line 109: Now handled by StyleProvider

src/aics-image-viewer/components/Toolbar/styles.css
```diff
@@ -67,77 +67,56 @@
...
-
-    &[ant-click-animating-without-extra-node="true"]::after {
-      box-shadow: none !important;
-      background-color: #222222;
-    }
-  }
-
-  span.btn-borderless {
-    background-color: transparent !important;
-    border-color: transparent !important;
```
ShrimpCryptid, line 140: Styling now mostly managed by `StyleProvider`. Icon buttons behave the same as other buttons when hovered/clicked.

src/website/components/ErrorPage.tsx
ShrimpCryptid, line 1: Fixed some styling that was broken by new default margins on text/headers.

src/website/components/LandingPage/index.tsx
```diff
@@ -173,10 +178,10 @@ const InReviewFlag = styled(FlexRowAlignCenter)`
   height: 22px;
   flex-wrap: wrap;
 
-  & > p {
+  && > p {
     margin-bottom: 0;
     color: var(--color-statusflag-text);
-    font-size: 10px;
+    font-size: 11px;
```
ShrimpCryptid, line 184: Goes off-design slightly but I think it's a bit more readable this way.

interim17 (approved):
This in general looks good to me on first read through and poking around when running it. I'm approving, although it was a lot of CSS to read through so it's possible I've missed things and I'm available if you want to walk through it together or have me take a second pass.

src/aics-image-viewer/components/StyleProvider/index.tsx
```diff
@@ -1,9 +1,363 @@
...
+
+type AppTheme = typeof theme;
+
+/** Makes theme styling available to child components via CSS variables. */
+const CssProvider = styled.div<{ $theme: AppTheme }>`
+  width: 100%;
+  height: 100%;
+
+  ${({ $theme }) => {
+    return css`
```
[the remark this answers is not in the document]
ShrimpCryptid: We can change the theme on the fly with this 👀👀👀👀

src/aics-image-viewer/components/BottomPanel/styles.css
```diff
@@ -3,30 +3,24 @@
...
   .ant-drawer-content-wrapper {
     display: flex;
     flex-direction: column;
     justify-content: flex-end;
-    height: unset !important;
-    position: absolute;
+    height: 190px !important;
+    transition: all 0.3s, max-width 0s 0s;
+    opacity: 85%;
+    box-shadow: none;
```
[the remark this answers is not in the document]
ShrimpCryptid: Gotcha. I'll see if I can re-center them again.

src/aics-image-viewer/components/Toolbar/DownloadButton.tsx
```diff
@@ -10,22 +11,29 @@ interface DownloadButtonProps {
...
+        {
+          key: "2",
+          label: (
+            <a href={fovDownloadHref}>
+              <ViewerIcon type="download" /> Full field image
+            </a>
+          ),
+        },
+      ],
+    };
```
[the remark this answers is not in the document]
ShrimpCryptid: Oh! I see what you mean, I'll add one.

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -252,7 +252,7 @@ export default class Toolbar extends React.Component<ToolbarProps, ToolbarState>
                 <ViewerIcon type="camera" />
               </Button>
             </Tooltip>
-          </span>
+          </div>
```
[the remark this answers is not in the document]
ShrimpCryptid: Inspected just to be sure, it looks like the toolbar icons are the same color on my end. It's possible that the difference might be from the icons having different thicknesses? Let me know if I'm missing something.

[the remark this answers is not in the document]
ShrimpCryptid: image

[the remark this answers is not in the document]
interim17: Hm, maybe yeah just the thinner line on a dark background looked like a darker shade to my eye, I wish the lines were thicker but that's not up to dev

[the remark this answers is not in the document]
ShrimpCryptid: I can add a ticket to revisit these. I wouldn't mind making that change tbh!

src/aics-image-viewer/components/StyleProvider/index.tsx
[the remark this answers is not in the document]
frasercl: Still not 100% on styled-components, but this is by no means too intrusive for the moment

src/aics-image-viewer/components/ControlPanel/index.tsx
```diff
@@ -45,29 +45,30 @@ export default function ControlPanel(props: ControlPanelProps): React.ReactEleme
 
   // TODO key is a number, but ClickParam assumes keys will always be strings
```
frasercl, line 46:
```suggestion
  // TODO key is a number, but MenuInfo assumes keys will always be strings
```
:(

src/aics-image-viewer/components/ControlPanel/index.tsx
```diff
@@ -100,6 +101,7 @@ export default function ControlPanel(props: ControlPanelProps): React.ReactEleme
           bordered={false}
           className="control-panel"
```
frasercl, line 102: Someday we'll come through and replace this `Card` component with something more sensible for its use... I think there's an issue out for that actually

frasercl (commented):
This is solid work and super exciting! Looking forward to improving this app without having to step around the weirder bits of Ant 3 (less files, icon fonts...)

Some minor visual issues I'm still seeing:
- I'm getting scroll bars on the bottom and right of the clipping panel
- Toggle-able buttons on the toolbar no longer enter their "active" state (I believe they should get a background when e.g. the axes are toggled on)
- There's some weird margin around items in the metadata viewer
- Ant's dropdowns are different now. Might be worth checking in with UX on whether we should do anything about it?
- Should the text in the render mode dropdown button be a bit easier to see when the dropdown is open?
- Might not be worth fixing with a rework coming soon, but isosurface settings sliders are also tough to see now.
- The side panel is a very slightly different color than the header bar (this is the UI being out of compliance previously, not an issue this update created!)

All in all, I think this is an impressively short list given the complexity of this issue.

src/aics-image-viewer/components/Toolbar/DownloadButton.tsx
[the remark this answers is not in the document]
frasercl: Looks like you were ahead of my comment and already don't apply a tooltip if the button opens a dropdown.

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -156,9 +156,9 @@ export default class Toolbar extends React.Component<ToolbarProps, ToolbarState>
           onScroll={this.checkScrollBtnVisible}
         >
           <span className="viewer-toolbar-left" ref={this.leftRef} />
-          <span className="viewer-toolbar-center" ref={this.centerRef}>
+          <div className="viewer-toolbar-center" ref={this.centerRef}>
```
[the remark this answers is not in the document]
ShrimpCryptid: Semantically `span` is typically used for inline styling and containers, like changing styles for single words. I think `div` makes more sense for something like this, where we're organizing groups of buttons.

Thank you for the suggestion, I'll make this change!

ShrimpCryptid:
Writing Cameron's feedback as a TODO list:
- [ ] I'm getting scroll bars on the bottom and right of the clipping panel
- [ ] Toggle-able buttons on the toolbar no longer enter their "active" state (I believe they should get a background when e.g. the axes are toggled on)
- [ ] There's some weird margin around items in the metadata viewer
- [ ] Ant's dropdowns are different now. Might be worth checking in with UX on whether we should do anything about it?
- [ ] Should the text in the render mode dropdown button be a bit easier to see when the dropdown is open?
- [ ] Might not be worth fixing with a rework coming soon, but isosurface settings sliders are also tough to see now.
- [ ] The side panel is a very slightly different color than the header bar (this is the UI being out of compliance previously, not an issue this update created!)

src/aics-image-viewer/components/StyleProvider/index.tsx
```diff
@@ -256,6 +262,12 @@ const CssProvider = styled.div<{ $theme: AppTheme }>`
     color: var(--color-button-icon-disabled-text);
   }
 
+  .btn-active {
+    color: var(--color-button-icon-activated-text);
+    background-color: var(--color-button-icon-activated-bg);
+    border: 1px solid var(--color-button-icon-activated-outline);
+  }
+
```
frasercl, line 270: As far as I recall, this is a specific class name given to buttons in the toolbar. Any advantages or concerns with including it in a global style like this?

(This is not to imply that you ought to change it, in general I've been frustrated by how little this app has built a unified design system into the code, just curious about this specific case)

ShrimpCryptid: In this case it does make more sense to tuck this into the toolbar. If we put it here in global styling, it's possible that if the classname changes or the style breaks in the future no one will think to remove this.

The ideal case for me actually is to have IconButton or ToggleButton be its own component in its own module and govern all of its internal styling. I don't want to make that change right now, but I'm doing something similar for the Header in a later PR because the Primary/Secondary/Tertiary button styles in our figma design doc doesn't match up 1:1 with how Ant defines buttons.

frasercl (approved):

PR #226, opened 2024-04-29
fix: Remove hash routing on GitHub pages

### LINKED ISSUE #213: Use browser routing (history API) instead of hash routing on
## Use Case

Upcoming PR #212 sets all "production mode" builds to use hash router instead of browser router.

## Solution

Configure S3 to correctly route to the base `index.html`, and then modify the logic in `index.tsx` so it uses `createBrowserRouter` instead of `createHashRouter`.

## Alternatives

ShrimpCryptid:
Closes #213, removing hash routing on GitHub pages. This means we don't need to include the `#` character for subpaths anymore, so our S3 and GitHub Pages will use consistent link formats.

Before: `https://allen-cell-animated.github.io/website-3d-cell-viewer/#/viewer`
After: `https://allen-cell-animated.github.io/website-3d-cell-viewer/viewer`

*Estimated size: medium, 20-25 minutes*

## Changes
This uses a custom 404 HTML page catch missing pages and reroute the browser to the viewer's base `index.html` page. The base page checks for the redirect and, if found, redirects internally to the correct page.

See https://github.com/rafgraph/spa-github-pages for the original solution.

- Adds `route_utils.tsx` for handling routing of URLs, and associated unit tests.
- Adds a custom `reroute.tsx` and `404.html` page to reroute missed traffic to the root single-page-app.
- Updates `index.tsx` to intercept the redirects and route the client to the correct page.
- Updates the GitHub actions for deploying to GitHub Pages and S3 to add missing basename information.
- Configures webpack to handle multiple entrypoints.

## Testing:
(this link will break after today so ping me if you want to see it again)
- Open https://allen-cell-animated.github.io/website-3d-cell-viewer/viewer. The viewer should reroute to our custom viewer.


toloudis (approved):
The functions that do the url manipulation look complicated and unreadable enough to merit unit tests but that could be for a future ticket.

public/index.tsx
```diff
@@ -30,14 +31,7 @@ const routes = [
...
 
-let router: Router;
-if (WEBSITE3DCELLVIEWER_BUILD_ENVIRONMENT === "dev") {
-  router = createBrowserRouter(routes);
-} else {
-  // Production mode.
-  // TODO: Use createBrowserRouter when building to S3.
-  router = createHashRouter(routes);
-}
+const router = createBrowserRouter(routes, { basename: basename });
```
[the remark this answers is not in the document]
ShrimpCryptid: I don't think so, it was there before I added it! It looks like it's being used just for a debug message:

```
console.log(`website-3d-cell-viewer ${WEBSITE3DCELLVIEWER_BUILD_ENVIRONMENT} build`);
```

Does that seem safe to remove?

ShrimpCryptid:
> The functions that do the url manipulation look complicated and unreadable enough to merit unit tests but that could be for a future ticket.

yeah.........

ShrimpCryptid:
Taking this back to draft while I do some reworking!

public/gh-reroute/404.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
```
ShrimpCryptid, line 1: image

^ here's what this page looks like! It should only briefly flash across the screen, so I tried to make sure it wouldn't be too obtrusive. I unfortunately don't think I can include a link back to the root of the viewer without some more complex configuration, so I settled on the github repo for now.

webpack.common.js
```diff
@@ -14,24 +14,33 @@ const themeVariables = lessToJs(
...
       new HtmlWebpackPlugin({
+        filename: "index.html",
         template: "./public/index.html",
+        chunks: ["index"],
+      }),
+      new CleanWebpackPlugin(),
+      new HtmlWebpackPlugin({
+        filename: "404.html",
+        template: "./public/gh-reroute/404.html",
+        chunks: ["reroute"],
```
ShrimpCryptid, line 36: Saves `gh-reroute/404.html` as `404.html` with reroute script linked.

public/gh-reroute/404.html
ShrimpCryptid: This may show for a little longer than it used to... I might be able to make it faster by not including the Open Sans font override, but it looks more consistent this way.

ShrimpCryptid:
@toloudis re-requesting your review because I've made some substantial changes. The reroute code now lives in a typescript file with unit tests, and I added more styling to the 404 page (as well as some webpack config changes).

public/gh-reroute/404.html
toloudis: Showing this page only happens in gh-pages builds, right?  Will end users of staging or production see this?

toloudis: Also, was this made worse because of refactoring the code?  Was it less noticeable before?

ShrimpCryptid: This will only be shown on GitHub pages, yes. I think it's worse because it's loading the separate code bundle in addition to just the plain HTML, and there's some boilerplate that makes the overall filesize larger.

public/gh-reroute/404.html
```diff
@@ -0,0 +1,64 @@
...
+          &:focus-within {
+            text-decoration: underline;
+          }
+        }
+      }
+    </style>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..800;1,300..800&display=swap"
```
meganrm, line 42: it doesn't seem like you're using all the weights here, could removing these cut down on the load time?

meganrm (approved):

public/gh-reroute/404.html
ShrimpCryptid: That's a good idea, thank you!

website/utils/route_utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  return new URL(newUrl);
+}
+
+export function isQueryStringPath(url: URL): boolean {
+  return url.search !== "" && url.search.startsWith("?/");
+}
+
+/**
+ * Converts a query string back into a complete URL. Used in combination with `convertUrlToQueryStringPath()`.
+ * to redirect the browser for single-page apps when the server cannot be configured.
```
toloudis, line 46: I think we can be even more specific and say this should only be used for gh-pages routing and NOT for our main deployment. (Same for the other function above, too). Maybe this whole file could be called gh_route_utils

public/gh-reroute/404.html
```diff
@@ -0,0 +1,61 @@
...
+          &:focus-within {
+            text-decoration: underline;
+          }
+        }
+      }
+    </style>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link href="https://fonts.googleapis.com/css2?family=Open+Sans&display=swap" rel="stylesheet" />
+    <!-- Webpack will inject the `gh-reroute/index.tsx` script into this webpage during build. -->
```
toloudis, line 42: if webpack is truly inlining the code then that would probably be the fastest load time.  I'm wondering if it is just creating a js script link that this page is loading dynamically.

toloudis (approved):
I wish there were a simpler way to deal with it.   I'm not even really convinced we needed this.  But happy that it's working.

public/gh-reroute/404.html
ShrimpCryptid: It's currently a linked script. I'm looking into plugins to actually inject the JS into HTML...

ShrimpCryptid: Okay, toyed with this a bit and I think what I have is working pretty well. I configured the script to load immediately (by disabling the `defer` attribute) and it will auto-clear the contents of the page, so you'll only ever see the "Rerouting..." text if JavaScript is disabled. This made the reroute happen much faster and feel snappier!

PR #232, opened 2024-05-02
feature: Help menu implementation

### LINKED ISSUE #220: Help menu
## Use Case

Implement help dropdown based on designs

## Solution

_Please describe your ideal solution_

## Alternatives

ShrimpCryptid:
Closes #220, implementing the help menu in the top header bar.

*Estimated size: small-medium, 20 minutes*

## Changes
- Edits CSS variables to reflect the primary/secondary/tertiary styling defined in the the original UX Figma file.
- Adds a `SecondaryButton` styled component. (purple outline that goes to full highlight color when hovered)
- Adds outline to the modals and dropdowns and adjusts content styling.
  - Adds the `getContainer` or `getPopupContainer` to all the places we are using dropdowns to ensure styling is applied.
- Adds the `HelpDropdown` component, containing a menu of buttons and links.

## Screenshots

image

**Video walkthrough (🔊):**

https://github.com/allen-cell-animated/website-3d-cell-viewer/assets/30200665/40809ec8-a282-493b-8c04-ab0933d130e3

### DIFF
--- a/src/aics-image-viewer/components/ControlPanel/index.tsx
+++ b/src/aics-image-viewer/components/ControlPanel/index.tsx
@@ -41,6 +41,9 @@ const ControlTabNames = {
 export default function ControlPanel(props: ControlPanelProps): React.ReactElement {
   const [tab, setTab] = React.useState(ControlTab.Channels);
 
+  const controlPanelContainerRef = React.useRef<HTMLDivElement>(null);
+  const getDropdownContainer = controlPanelContainerRef.current ? () => controlPanelContainerRef.current! : undefined;
+
   const { viewerChannelSettings, showControls, hasImage } = props;
 
   // TODO key is a number, but MenuInfo assumes keys will always be strings
@@ -56,7 +59,7 @@ export default function ControlPanel(props: ControlPanelProps): React.ReactEleme
       onClick: makeTurnOnPresetFn,
     };
     return (
-      <Dropdown trigger={["click"]} menu={dropDownMenuProps}>
+      <Dropdown trigger={["click"]} menu={dropDownMenuProps} getPopupContainer={getDropdownContainer}>
         <Button>
           <div style={{ display: "flex", flexDirection: "row", alignItems: "center", gap: "4px" }}>
             Apply palette
@@ -80,7 +83,7 @@ export default function ControlPanel(props: ControlPanelProps): React.ReactEleme
   );
 
   return (
-    <div className="control-panel-col-container">
+    <div className="control-panel-col-container" ref={controlPanelContainerRef}>
       <div className="control-panel-tab-col" style={{ flex: "0 0 50px" }}>
         <Button
           className={"ant-btn-icon-only btn-collapse" + (props.collapsed ? " btn-collapse-collapsed" : "")}
--- a/src/aics-image-viewer/components/StyleProvider/index.tsx
+++ b/src/aics-image-viewer/components/StyleProvider/index.tsx
@@ -110,7 +110,10 @@ const theme = {
       split: palette.white,
     },
     menu: {
-      selectedBg: palette.ltPurple,
+      hoverText: palette.white,
+      hoverBg: palette.ltPurple,
+      selectedText: palette.white,
+      selectedBg: palette.medGrey,
       textPlaceholder: palette.ltPurple,
     },
     tooltip: {
@@ -138,6 +141,8 @@ const CssProvider = styled.div<{ $theme: AppTheme }>`
     return css`
       /* Component and color variables. */
       // TODO: Fix inconsistent use of border vs. outline
+      // TODO: Remove variables that aren't used in other CSS files. These should be set directly
+      // from the $theme object to reduce unnecessary variables.
       --color-text-link: ${$theme.colors.text.link};
       --color-text-header: ${$theme.colors.text.header};
       --color-text-section: ${$theme.colors.text.section};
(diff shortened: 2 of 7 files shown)


src/aics-image-viewer/components/ControlPanel/index.tsx
```diff
@@ -41,6 +41,9 @@ const ControlTabNames = {
 export default function ControlPanel(props: ControlPanelProps): React.ReactElement {
   const [tab, setTab] = React.useState(ControlTab.Channels);
 
+  const controlPanelContainerRef = React.useRef<HTMLDivElement>(null);
+  const getDropdownContainer = controlPanelContainerRef.current ? () => controlPanelContainerRef.current! : undefined;
+
```
ShrimpCryptid, line 46: This pattern is used in a few places, but it prevents Ant from directly appending menus and popups to the `body` tag. Instead, they'll be anchored in the HTML tree to wherever the `containerRef` is, allowing them to receive style overrides from `StyleProvider`.

src/aics-image-viewer/components/StyleProvider/index.tsx
```diff
@@ -65,6 +65,11 @@ const theme = {
         text: palette.medPurple,
       },
       secondary: {
+        bg: "transparent",
+        text: palette.medPurple,
+        outline: palette.medPurple,
+      },
+      tertiary: {
```
ShrimpCryptid, line 72: I'd made a mistake when I set up these CSS variables and mixed up the secondary and tertiary button styles. Some of the line changes here fix that by renaming variables accordingly.

image

src/aics-image-viewer/components/StyleProvider/index.tsx
```diff
@@ -155,18 +166,23 @@ const CssProvider = styled.div<{ $theme: AppTheme }>`
 
       --color-button-link-text: ${$theme.colors.button.link.text};
 
-      --color-button-default-bg: transparent;
-      --color-button-default-text: ${$theme.colors.button.secondary.text};
-      --color-button-default-outline: ${$theme.colors.button.secondary.outline};
-      --color-button-default-hover-outline: ${$theme.colors.button.secondary.hoverOutline};
-      --color-button-default-hover-text: ${$theme.colors.button.secondary.hoverText};
-      --color-button-default-active-outline: ${$theme.colors.button.secondary.activeOutline};
-      --color-button-default-active-text: ${$theme.colors.button.secondary.hoverText};
```
ShrimpCryptid, line 164: `default` -> `tertiary`

interim17 (approved):

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -58,6 +59,7 @@ export default class Toolbar extends React.Component<ToolbarProps, ToolbarState>
       scrollBtnLeft: false,
       scrollBtnRight: false,
     };
+    this.containerRef = React.createRef();
```
frasercl, line 62: If you can believe it, I converted this to a function component at one point, then back into a class component because I thought it was more convenient! Seeing the old `createRef` makes me think I ought to come back again and switch it back now that I'm more familiar with React's new functional preference.

src/aics-image-viewer/components/ControlPanel/index.tsx
frasercl: This solution is low-impact enough to be acceptable, but thinking about doing this every time we need a dropdown does make me pause and wonder if there's a nicer place to hold our global styles long-term. Should the CSS applied to `StyleProvider` live in a class that we can name and apply to Ant's container via `overlayClassName`? If so, would we be able to build the theme in from JavaScript as easily as `StyleProvider` does now (maybe with another build step)? Would it be as easy as it is now to maintain a single source of truth for theme colors (probably not)? Food for thought as we think about how to style this app going forward.

src/aics-image-viewer/components/StyleProvider/index.tsx
frasercl: Why are there so much fewer variables for `secondary` than `tertiary`? Do secondary buttons just deviate less from the global button styling below?

website/components/HelpDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+          return (
+            <Button
+              onClick={() => {
+                setShowVersionModal(false);
+              }}
+            >
+              Close
+            </Button>
+          );
+        }}
```
frasercl, line 94: I'm a weirdo who likes things small and this is a 100% arbitrary stylistic preference that you're within your right to ignore, but: this can be shrunk considerably

```suggestion
        onCancel={() => setShowVersionModal(false)}
        footer={<Button onClick={() => setShowVersionModal(false)}>Close</Button>}
```

could even give `() => setShowVersionModal(false)` an alias above for simplicity:


```suggestion
        onCancel={closeVersionModal}
        footer={<Button onClick={closeVersionModal}>Close</Button>}
```

frasercl (approved):
Looks good! Left some questions and thoughts and one non-blocking suggestion.

src/aics-image-viewer/components/StyleProvider/index.tsx
ShrimpCryptid: Secondary derives almost entirely from primary, yes! They behave similarly when hovered and active.

image

src/aics-image-viewer/components/ControlPanel/index.tsx
ShrimpCryptid: In colorizer this is handled by `AccessibleDropdown`! Making a component that does the container management is a good way to handle it.

Using `overlayClassName` would require us to attach CSS rules to the body or use `:root`, both of which could potentially lead to conflicts with other styles when we embed the viewer... But I see what you're saying and yes, let's keep thinking about style management.

src/aics-image-viewer/components/StyleProvider/index.tsx
ShrimpCryptid: Oops, thought I'd responded to this but evidently not.
Secondary buttons derive mostly from primary buttons, except when at rest or disabled.

image

Issue #235, opened 2024-05-09
Refactor viewer app to allow direct state access when sharing


ShrimpCryptid:
## Description

 The page header's share button has no way to access state from inside of ImageViewerApp to include in a shareable URL. This feeds into a larger discussion about its design and architecture.

New suggested architecture, from Dan:
```
// website / standalone app
<LandingPage>
    <Header></Header>
</LandingPage>

<AppWrapper>
    <AppStateManager> // <=== "APP"
        <Header>
        </Header>
        <ImageViewerApp>
            // also gets embedded
        </ImageViewerApp>
    </AppStateManager>
</AppWrapper>

// embedded
<AppStateManager>
    <Header> // omit for embedding
    </Header>
    <ImageViewerApp>
        // also gets embedded
    </ImageViewerApp>
</AppStateManager>

// renaming
<ImageViewerApp>
    <Header>
    </Header>
    <ImageViewerComponent>
    </ImageViewerComponent>
</ImageViewerApp>
```

Based on a discussion with @frasercl and @toloudis on 4/29/24. Lmk if I'm missing any details!

ShrimpCryptid:
Redux????

ShrimpCryptid:
- Context to reduce the amount of prop drilling?
- Better to have lots of small atomic props instead of a SuperObject(tm)
- Need to define/document that props are initial state only

PR #236, opened 2024-05-09
fix: Fix initial URL parameters being ignored

ShrimpCryptid:
Fixes a bug where initial URL parameters were not being loaded correctly. 

I realized that certain view settings weren't being pulled correctly from the URL-- this was both because I was accidentally overwriting them (oops) but also because of a more complex issue with `ImageViewerApp`.

It turns out that the `ImageViewerApp` only uses passed props as initial values, and ignores all future updates to them. To fix this, I initialized the viewer args + settings as `null` in `AppWrapper` and block rendering of the `ImageViewerApp` until they are loaded for the first time. This way the args + settings are complete the first time they are passed into `ImageViewerApp.`

## Testing
1. Clone the branch and start the dev server using `npm run start`.
2. Open the following link in your browser: http://localhost:9020/viewer?url=https%3A%2F%2Fs3-us-west-2.amazonaws.com%2Fbisque.allencell.org%2Fv1.4.0%2FCell-Viewer_Thumbnails%2FAICS-22%2FAICS-22_8319_2025_atlas.json&view=Z
3. You should see the XY view mode be selected.


website/components/AppWrapper.tsx
```diff
@@ -110,17 +118,19 @@ export default function AppWrapper(inputProps: AppWrapperProps): ReactElement {
...
-        viewerSettings={viewerSettings}
-      />
+      {viewerArgs && viewerSettings && (
+        <ImageViewerApp
+          {...viewerArgs}
+          appHeight={`calc(100vh - ${HEADER_HEIGHT_PX}px)`}
+          canvasMargin="0 0 0 0"
+          viewerSettings={viewerSettings}
+        />
+      )}
```
ShrimpCryptid, line 133: Conditional rendering, waits until the args have been initialized.

website/components/AppWrapper.tsx
```diff
@@ -61,17 +61,25 @@ export default function AppWrapper(inputProps: AppWrapperProps): ReactElement {
...
+        console.log("URL viewer settings", urlViewerSettings);
+        setViewerArgs({ ...DEFAULT_APP_PROPS, ...props.viewerArgs, ...urlArgs, ...locationArgs });
+        setViewerSettings({ ...DEFAULT_VIEWER_SETTINGS, ...props.viewerSettings, ...urlViewerSettings });
+      },
+      (reason) => {
+        console.warn("Failed to parse URL parameters: ", reason);
+        setViewerArgs({ ...DEFAULT_APP_PROPS, ...props.viewerArgs, ...locationArgs });
+        setViewerSettings({ ...DEFAULT_VIEWER_SETTINGS, ...props.viewerSettings });
+      }
+    );
```
ShrimpCryptid, line 82: Second part here triggers if the `getArgsFromParams` operation fails. That way the viewer args + settings will always be initialized.

website/components/AppWrapper.tsx
meganrm: I understand you made this change based on "It turns out that the ImageViewerApp only uses passed props as initial values, and ignores all future updates to them. " but that seems like a weird behavior in a react app? 

Also to make the meaning clearer, I might just make a flag called `processedUrlParamsComplete` or something instead of this double exists

website/components/AppWrapper.tsx
```diff
@@ -62,17 +62,24 @@ export default function AppWrapper(inputProps: AppWrapperProps): ReactElement {
   const location = useLocation();
   const navigation = useNavigate();
 
-  const [viewerSettings, setViewerSettings] = useState<Partial<GlobalViewerSettings>>(props.viewerSettings);
-  const [viewerArgs, setViewerArgs] = useState<AppDataProps>(props.viewerArgs);
+  const [viewerSettings, setViewerSettings] = useState<Partial<GlobalViewerSettings> | null>(null);
+  const [viewerArgs, setViewerArgs] = useState<AppDataProps | null>(null);
```
meganrm, line 66: it's confusing to me that there are props.viewerArgs and this state.viewerArgs. Where are the props.viewerArgs coming from? Also since you have viewerSettings, maybe `viewerProps` is a better name for `viewerArgs`? 

If viewerArgs and viewerSettings are getting derived once from a combination of props and url params, it seems like you don't need `useState` at all since they're never getting changed

website/components/AppWrapper.tsx
ShrimpCryptid: That makes sense, yeah! And yes, I agree the behavior is weird. There's an issue ticket here for a potential refactor to move state handling out of the main ImageViewerApp: #235

website/components/AppWrapper.tsx
ShrimpCryptid: That rename makes sense to me, especially since the underlying type for `viewerArgs` is `AppDataProps`. `viewerArgs` is usually being derived from the browser's location (data passed as part of the navigation state) or from the URL. I think I'll remove the `prop` for it since `AppWrapper` is only used on the website and doesn't really have parent components that would need to pass in prop data.

I think I would still need a `useState` since the derivation from the URL is asynchronous... Can you let me know if you see another way to do it?

meganrm: one thought is just to use `useRef` for the actual data and then store a single `requestHasFinished` flag in state instead of two separate state updates. But since these weren't added by you in this PR and you're fixing a bug I think it makes sense to leave them as is for now. But a refactor where you're not storing props in state and then exporting them as props again I think would be good

meganrm (approved):

website/components/AppWrapper.tsx
```diff
@@ -111,17 +107,19 @@ export default function AppWrapper(inputProps: AppWrapperProps): ReactElement {
...
           <HelpDropdown />
         </FlexRowAlignCenter>
       </Header>
-      <ImageViewerApp
-        {...viewerArgs}
-        appHeight={`calc(100vh - ${HEADER_HEIGHT_PX}px)`}
-        canvasMargin="0 0 0 0"
-        viewerSettings={viewerSettings}
-      />
+      {viewerProps && viewerSettings && (
```
rugeli, line 115: this check before rendering `ImageViewerApp` looks like a good practice!

rugeli (approved):
I was able to open the localhost and saw the `XY` mode was selected.

PR #237, opened 2024-05-10
fix: Updated histogram on T/Z change

### LINKED ISSUE #201: histogram does not update when z or t changes
## Description

Open channel settings.  Play through T.  Lookup table graph changes but histogram does not.

## Expected Behavior

Expect histogram to reflect the actual intensities that are displayed.

ShrimpCryptid:
Closes #201, a bug where the histogram would not update when either T or Z values changed.

*Estimated review size: tiny, < 5 minutes*

## Changes
- `TfEditor` now triggers histogram redraw when volume data has changed.
- Histogram redraw uses the already-calculated Channel histogram data instead of recalculating bins; this previously took about 250 ms per redraw.

**Video demo:**

https://github.com/allen-cell-animated/website-3d-cell-viewer/assets/30200665/2b839e87-e7a8-4298-be6d-11c0a84c0f19


src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -356,24 +356,37 @@ export default class MyTfEditor extends React.Component<MyTfEditorProps, MyTfEdi
     svg.selectAll<SVGGElement, unknown>(".axis.axis--x").call(axis);
   }
 
+  private getBinLengths(): { binLengths: number[]; max: number } {
+    const histogram = this.props.channelData.histogram;
+    const binLengths = [];
+    let max = Number.NEGATIVE_INFINITY;
+    for (let i = 0; i < histogram.getNumBins(); i++) {
+      const binLength = histogram.getBin(i);
+      binLengths.push(binLength);
```
toloudis, line 365: I wonder if this could be even faster. It looks like it's basically just copying the value of each bin one by one, so could probably do something like histogram.bins.slice() to get binLengths (or even just use it directly without copying if we're sure it won't be modified).  And the max value might already be stored inside the histogram class too. (It ought to be, if it isn't)

Whether it's worth doing this depends on the new timings.  You mentioned 250ms before, but what is the after? (And is this function still the biggest part)

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -356,24 +356,37 @@ export default class MyTfEditor extends React.Component<MyTfEditorProps, MyTfEdi
...
         .attr("class", "bar")
-        .attr("transform", (d) => `translate(${this.xScale(d.x0!)},${this.binScale(d.length)})`);
+        .attr("transform", (length, idx) => `translate(${this.xScale(idx)},${this.binScale(length)})`);
 
       barEnter
         .append("rect")
         .attr("x", 1)
-        .attr("width", (d) => d.x1! - d.x0!)
-        .attr("height", (d) => this.height - this.binScale(d.length));
+        .attr("width", () => 1)
```
toloudis, line 388: this function is awesome. Are you allowed to replace `()=>1` with `1` ?

toloudis (approved):

src/aics-image-viewer/components/TfEditor/index.tsx
ShrimpCryptid: I actually couldn't find this function in the performance graph (though admittedly maybe I wasn't looking hard enough)! 

`histogram.bins` is private so I don't have direct access to it unless I perform some unsavory Javascript crimes.

toloudis: ok - we can do some work on volume-viewer to expose that somehow.  In C++ you can expose it as a read-only reference but in js it's all or nothing, I guess.

src/aics-image-viewer/components/TfEditor/index.tsx
ShrimpCryptid: I am 😆

frasercl (approved):

PR #244, opened 2024-05-15
feat: Replaced Dockerfile

### LINKED ISSUE #169: Dockerfile is out of date
## Description

Our Dockerfile does not build because it's so old and depends on old infrastructure.

## Expected Behavior

A very simple dockerfile that can copy the source code in, run npm build, and serve up the dist/ directory over http

## Reproduction

try building the dockerfile!

ShrimpCryptid:
Closes #169, "Dockerfile is out of date". Adds a new Dockerfile that performs the build and makes the files available via a nginx dev server, and fixes some other bugs along the way with video.

## Changes
- Deletes old Dockerfile and related config files.
- Adds a new Dockerfile that builds the project, then serves the files over http using an nginx server.
- Updates webpack config to fix a bug where videos would be missing during builds. (This *should* also fix the missing banner video on GitHub pages.)

## Verification
Follow the steps specified in the README! Spelling them out here:
1. Install Docker locally: [instructions for Windows](https://docs.docker.com/desktop/install/windows-install/), [instructions for Mac](https://docs.docker.com/desktop/install/mac-install/)
2. Clone the project and navigate to the root.
3. Run `docker build -t 3d-volume-viewer .`
4. Run `docker run --rm -p 8080:80 --name 3d-volume-viewer 3d-volume-viewer`
5. Go to http://localhost:8080. Pages should load in as expected.
6. Refreshing any of the subpages should still show the subpage as expected.


package.json
```diff
@@ -22,6 +22,7 @@
     "gh-build-nightly": "webpack --config webpack.dev.js --env env=production basename=/website-3d-cell-viewer",
     "gh-build-release": "webpack --config webpack.dev.js --env env=production basename=/website-3d-cell-viewer-release",
     "s3-build": "webpack --config webpack.dev.js --env env=production",
+    "docker-build": "webpack --config webpack.dev.js --env env=production",
```
ShrimpCryptid, line 25: Same instruction as `s3-build` but it seemed unsafe to reuse it directly.

toloudis:
Added @amilworks because he has written a dockerfile for this viewer before

docker/dockerrun.sh
```diff
@@ -1,2 +1 @@
-docker run -p 9020:9020 -d -v /allen:/allen:rshared danielt/ace2-3d-viewer
-#docker run -p 9020:9020 -d -v /allen:/allen:rshared -e DEPLOYMENT_ENV=dev $docker_image
+docker run --rm -p 9020:80 --name 3d-volume-viewer 3d-volume-viewer-image
```
toloudis, line 1: want to keep using /allen as a volume mount, to load onprem Vast files?  Or keep relying on my dev server's config to do that...

package.json
toloudis: still, any changes to webpack.dev.js will potentially cause issues because shared across multiple build types.  Might be a good thing to note in one of our readme or CONTRIBUTING type of docs.

website/components/LandingPage/index.tsx
```diff
@@ -323,7 +324,7 @@ export default function LandingPage(): ReactElement {
       <Banner>
         <BannerVideoContainer style={{ zIndex: 1 }}>
           <video autoPlay={allowMotion} loop muted>
-            <source src="/videos/banner-video.mp4" type="video/mp4" />
+            <source src={BannerVideo} type="video/mp4" />
```
toloudis, line 327: is this line still downloading the banner video or is it inlining it as something like a data url?

ShrimpCryptid: It'll link to the banner video file! Doing it like this lets webpack configure the final path to the video asset.

docker/dockerrun.sh
ShrimpCryptid: Leaning towards not, based on a conversation I had with Andrew:
> I am curious if the plan is still to bind-mount /allen into the production container.  That's caused us a lot of problems over the past six years with other containers; this might be a good time to brainstorm alternative approaches to that.

amilworks:
> Added @amilworks because he has written a dockerfile for this viewer before

Yes, I made a very minimal `Dockerfile` that worked with the cvapipe app as a standalone service. I embedded the 3D Cell Viewer using an iFrame. Below is the Dockerfile I used.

```docker
# Official Node.js runtime as a parent image
FROM node:14

# Set working dir inside the container
WORKDIR /app

# Copy the rest of the application code to the container
COPY . .

# Expose the port for the app 
EXPOSE 9020

# Define command to run 3D Viewer application
CMD [ "npm", "start" ]
```

I built a separate service to serve images from AWS which simply uses FastAPI.

docker/dockerrun.sh
amilworks: I too would lean towards not volume mounting into the prod container. Since you can serve files from a more controlled location on a separate service, this will lend a bit more flexibility and security compared to mounting.

amilworks (approved):
This Docker looks better than mine and uses nginx. Nice work!

package.json
ShrimpCryptid: Hm, now that I'm thinking about it most of these build commands likely don't need to use the dev webpack config. I'll test removing it for docker and open an issue for this.

ShrimpCryptid: See #247!

PR #253, opened 2024-05-24
feat: Use new URL parameter schema

### LINKED ISSUE #249: Import channels settings through URL parameters
## Use Case
Preload certain settings per channel for sharing w/ default settings

## Solution

```
?url=url1,url2&enabledChannels=[1,2]&c0=[...]&c3=[...]
```
Major URL parameters:
- URLs, comma separated and encoded
- List of enabled channels
- For each channel, a parameter `c{n}` where `n` is the index of that channel, starting from 0.
  - Value is a comma separated, `key:value` list. (Can also try JSON.)
  - Keys correspond to values in `ChannelState`, copied below:
```
export interface ChannelState {
  name: string;
  volumeEnabled: boolean;
  isosurfaceEnabled: boolean;
  isovalue: number;
  colorizeEnabled: boolean;
  colorizeAlpha: number;
  opacity: number;
  color: ColorArray;
  controlPoints: ControlPoint[];
}
```

Also, all of the advanced render settings:
- Mask cell
- Brightness
- Density
- Levels (3 values)
- Interpolate (boolean)
- Background color
- Bounding box (boolean)
- Bounding box color

ShrimpCryptid:
Closes #249! This is marked as priority so we can get this feature to Matheus.

*Estimated review size: medium, 20-30 minutes (about 250 lines are just unit tests)*

## Changes

Implements the new URL schema, where channels settings are defined as `c{n}=key:value,key:value` where `n` is some integer index of the channel. Key/value pairs are stored as a comma-separated list, with key-values separated by colons. 

- Edit App to allow colorize on/off and alpha to be provided via settings.
- Changed LUT parsing to recognize the `autoij` parameter.
- Parses new per-channel settings.
  - Each channel is defined as a URL parameter `c{n}` where n is an integer index.
  - A string of key-value pairs is parsed into a `ViewerChannelSetting` object.
- Adds unit testing for URL parsing. 

## Testing
1. Clone the branch and start the dev server.
2. Open the following link locally: http://localhost:9020/viewer?url=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Frelease%2Fnuc-morph-data%2Fdata_for_analysis%2Fperturbations%2Fimportazole%2FLaminB1%2Fcontrol%2F20230417_01%2Fraw.ome.zarr%2Chttps%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Frelease%2Fnuc-morph-data%2Fdata_for_analysis%2Fperturbations%2Fimportazole%2FLaminB1%2Fcontrol%2F20230417_01%2Fseg.ome.zarr&c0=ven:0&c1=ven:1,lut:autoij:&c2=ven:1,clz:1&view=Z 

The key part is at the end! Just to make it a little clearer:
```js
c0=ven:0              // disable volume
c1=ven:1,lut:autoij:  // enable volume, use autoij on lut
c2=ven:1,clz:1         // enable volume, enable colorize (cz)
```

3. Test one more link, which tests all of the URL properties: http://localhost:9020/viewer?url=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Frelease%2Fnuc-morph-data%2Fdata_for_analysis%2Fperturbations%2Fimportazole%2FLaminB1%2Fcontrol%2F20230417_01%2Fraw.ome.zarr%2Chttps%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Frelease%2Fnuc-morph-data%2Fdata_for_analysis%2Fperturbations%2Fimportazole%2FLaminB1%2Fcontrol%2F20230417_01%2Fseg.ome.zarr&c0=ven:1,col:ff0000,clz:1,cza:0.55,isa:0.3,isv:28,lut:0:100,ven:1,sen:1
```
volume: enabled
surface: enabled
colorize: enabled
colorize alpha: 0.5
surface isovalue: 28
surface opacity: 0.3 (~76)
color: #ff0000 (red)
lut: control points at 0 and 100
```


public/firebase/configure-firebase.ts
```diff
@@ -25,7 +25,12 @@ const firebaseDevConfig = {
 };
 // Initialize Firebase
 firebase.initializeApp(firebaseConfig);
-firebase.analytics();
+
+analytics.isSupported().then((supported) => {
+  if (supported) {
+    firebase.analytics();
+  }
+});
```
ShrimpCryptid, line 33: Fix a warning when running tests

website/utils/url_utils.ts
```diff
@@ -2,16 +2,47 @@ import FirebaseRequest, { DatasetMetaData } from "../../public/firebase";
...
+   * by a colon, e.g. "0:255", "p50:p90", or "autoij:0".
+   */
+  lut?: string;
+  /** Volume enabled. "1" is enabled. */
+  v?: "1" | "0";
+  /** Isosurface enabled. "1" is enabled. */
+  i?: "1" | "0";
+  /** Isosurface value, in the [0, 255] range. */
+  iv?: string;
+};
```
ShrimpCryptid, line 34: Anything you'd change about this naming?

toloudis: IMO, nothing super important to change - we can totally use this as is - but some ideas. 

Would it look nicer/more consistent if all the keys had the same number of characters?

Also, I believe opacity (kinda the same thing as alpha) is really only used for isosurface so it could get an `i` in front.  Also we could consider that "surface" is the key word and use `s` instead of `i`.
So you could have: `col, clz, cza, isa, lut, ven, sen (or ien), isv`: color, colorize on/off, colorize alpha, isosurface alpha, lut, volume enable, surface enable, isovalue

website/utils/url_utils.ts
```diff
@@ -2,16 +2,47 @@ import FirebaseRequest, { DatasetMetaData } from "../../public/firebase";
...
 type ParamKeys = (typeof paramKeys)[number];
-type Params = { [_ in ParamKeys]?: string };
+type Params = { [_ in ParamKeys | ChannelKey]?: string };
 
-function urlSearchParamsToParams(searchParams: URLSearchParams): Params {
+export function urlSearchParamsToParams(searchParams: URLSearchParams): Params {
   const result: Params = {};
   for (const [key, value] of searchParams.entries()) {
-    if (paramKeys.includes(key)) {
+    if (paramKeys.includes(key) || CHANNEL_STATE_KEY_REGEX.test(key)) {
```
toloudis, line 45: Nit: consider separating `ch, luts, colors` out of paramKeys so we can check against mixing the two types of per-channel settings?    Would that help keep things more organized?

website/utils/url_utils.ts
```diff
@@ -43,6 +74,66 @@ const tryDecodeURLList = (url: string, delim = ","): string[] | undefined => {
...
+): ViewerChannelSetting {
+  const result: ViewerChannelSetting = {
+    match: channelIndex,
+    enabled: jsonState.v === "1",
+    surfaceEnabled: jsonState.i === "1",
+    isovalue: parseFloat(jsonState.iv),
+    surfaceOpacity: parseFloat(jsonState.op),
+    colorizeEnabled: jsonState.cz === "1",
+    colorizeAlpha: parseFloat(jsonState.cza),
+  };
```
toloudis, line 122: might be worth a little comment to say defaults for missing fields are expected to be handled downstream of all this?

website/utils/url_utils.ts
```diff
@@ -139,6 +231,36 @@ export async function getArgsFromParams(urlSearchParams: URLSearchParams): Promi
...
+  const channelIndexToSettings: Map<number, ViewerChannelSetting> = new Map();
+  Object.keys(params).forEach((key) => {
+    if (CHANNEL_STATE_KEY_REGEX.test(key)) {
+      const channelIndex = parseInt(key.slice(1), 10);
+      try {
+        const channelData = parseKeyValueList(params[key]!);
+        const channelSetting = deserializeViewerChannelSetting(channelIndex, channelData as ViewerChannelSettingJson);
+        channelIndexToSettings.set(channelIndex, channelSetting);
+      } catch (e) {
+        console.warn(
```
toloudis, line 247: Not saying we should yet, but could we handle this as an error if we decided it was merited?

ShrimpCryptid: Hmm, currently this lives outside of Cameron's error-handling code and alert banners, since this function is called by the wrapper around our viewer app. We'd have to pass in an error object as a prop or something to make it visible to a user, or copy the error component and have an instance of it live at a wrapper-level?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -403,15 +405,21 @@ const App: React.FC<AppProps> = (props) => {
...
     }
 
     return {
       name: channel || "Channel " + index,
       volumeEnabled: volumeEnabled,
       isosurfaceEnabled: surfaceEnabled,
-      colorizeEnabled: false,
-      colorizeAlpha: 1.0,
+      colorizeEnabled: colorizeEnabled,
+      colorizeAlpha: colorizeAlpha,
```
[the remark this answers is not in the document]
ShrimpCryptid: Good catch, thank you 😅

website/utils/url_utils.ts
ShrimpCryptid: Let me think on this one! I'll at least move them into their own list (like `deprecatedParamKeys`).

website/utils/test/url_utils.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      op: "0.75",
+      v: "1",
+      i: "1",
+      iv: "128",
+      lut: "autoij:",
+    });
+  });
+
+  it("removes whitespace", () => {
+    const data = " key1 : value1 , key2 : value2 ";
```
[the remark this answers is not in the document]
ShrimpCryptid: I'll add a check to make sure it doesn't remove whitespace from the center of values. This only trims trailing or leading whitespace currently

website/utils/url_utils.ts
ShrimpCryptid: Renamed! Thank you for the suggestions!

website/utils/test/url_utils.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    const groups = args.viewerChannelSettings?.groups[0]!;
+    expect(groups.channels).toHaveLength(1);
+    const channelSetting = groups.channels[0];
+
+    expect(channelSetting.match).toEqual(1);
+    expect(channelSetting.enabled).toEqual(true);
+    expect(channelSetting.lut).toEqual(["4", "5"]);
+  });
+
+  it("creates empty default data for bad per-channel setting formats", async () => {
```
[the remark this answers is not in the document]
ShrimpCryptid: I can add a test for this, but if this were to happen it should only have defined settings for c0 and c2 as an output? This would make c1 use the default viewer settings, where both the isosurface and volume are not visible.

ShrimpCryptid:
Noticing a possible bug where the autoIJ or default behavior are in the right relative position, but are offset slightly:
image

After pressing "Reset":
image

Is this a known bug or something new?

website/utils/test/url_utils.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      surfaceOpacity: 0.75,
+      colorizeEnabled: true,
+      colorizeAlpha: 0.5,
+      lut: ["0", "255"],
+    });
+  });
+
+  it("handles expected lut formatting", () => {
+    const luts = [
+      ["autoij:0", ["autoij", "0"]],
```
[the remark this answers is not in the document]
ShrimpCryptid: Added. The verification link in the URL is missing one of the min/max parameters; it seems to default to 0 to 255 if one or both parameters are missing (and it's not using a named settings like `autoij`)

toloudis:
> Is this a known bug or something new?

Not sure exactly how you arrived at the first image, but I think clicking "Reset" should have the result that you posted in the 2nd image so that part looks right.

website/utils/url_utils.ts
```diff
@@ -2,16 +2,60 @@ import FirebaseRequest, { DatasetMetaData } from "../../public/firebase";
...
+
+type AllParamKeys = ParamKeysV0 | ParamKeysV2_9_0;
+
+type Params = { [_ in AllParamKeys]?: string };
+
+const isParamKey = (key: string): key is BaseParamKeys => baseParamKeys.includes(key as BaseParamKeys);
+const isDeprecatedParamKey = (key: string): key is DeprecatedParamKeys =>
+  deprecatedParamKeys.includes(key as DeprecatedParamKeys);
+const isChannelKey = (key: string): key is ChannelKey => CHANNEL_STATE_KEY_REGEX.test(key);
+
```
ShrimpCryptid, line 54: @frasercl made some changes here to try and explicitly type parameters, but I want to get your input on if there's better ways to handle this.

ShrimpCryptid:
> > Is this a known bug or something new?
> 
> Not sure exactly how you arrived at the first image, but I think clicking "Reset" should have the result that you posted in the 2nd image so that part looks right.

First image was from doing an empty `lut::`. I also get a similar bug even with AutoIJ though, where it seems to be in the right relative position but is shifted when AutoIJ is clicked again. You can see it in the test link I included on the PR.

website/utils/url_utils.ts
frasercl: This looks good! The only thing I'd nitpick is whether `ParamKeysV0` and `ParamKeysV2_9_0` are necessary, or whether they could be documentation. Looks like they're both only used in the definition of `AllParamKeys`, which could just be `BaseParamKeys | DeprecatedParamKeys | ChannelKey`.

website/utils/url_utils.ts
```diff
@@ -43,6 +87,67 @@ const tryDecodeURLList = (url: string, delim = ","): string[] | undefined => {
...
+    if (HEX_COLOR_REGEX.test(jsonState.col)) {
+      result.color = jsonState.col;
+    }
+  }
+  if (jsonState.lut) {
+    if (LUT_REGEX.test(jsonState.lut)) {
+      const [min, max] = jsonState.lut.split(":");
+      result.lut = [min.trim(), max.trim()];
+    }
+  }
```
frasercl, line 147: More because I think it's cool than because it produces better or more readable code: TypeScript knows that `&&` is short-circuiting, and will narrow types on the right-hand side based on the left:
```typescript
if (jsonState.col && HEX_COLOR_REGEX.test(jsonState.col)) {
  // ...
}
```

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -383,6 +383,10 @@ const App: React.FC<AppProps> = (props) => {
     let color = defaultColor;
     let volumeEnabled = false;
     let surfaceEnabled = false;
+    let colorizeEnabled = false;
+    let colorizeAlpha = 1.0;
+    let isosurfaceOpacity = 1.0;
+    let isovalue = 128;
```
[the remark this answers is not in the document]
ShrimpCryptid: I really like this change, it cleans up the code quite a bit. Implemented this!

website/utils/url_utils.ts
```diff
@@ -2,16 +2,56 @@ import FirebaseRequest, { DatasetMetaData } from "../../public/firebase";
...
+  /** Color, as a 6-digit hex color */
+  col?: string;
+  /** Colorize. "1" is enabled. */
+  clz?: "1" | "0";
+  /** Colorize alpha, in the [0, 1] range. */
+  cza?: string;
+  /** Opacity, in the [0, 1 range] */
+  isa?: string;
+  /** LUT to map from intensity to opacity. Should be two alphanumeric values separated
+   * by a colon, e.g. "0:255", "p50:p90", or "autoij:0".
```
toloudis, line 25: Is this the right place to describe/document the encodings?  Plain number = intensity, p[Number 0-100] = intensity of percentile, m[Number]=median intensity * number, autoij = override with algorithm to automatically select min and max.

toloudis: For documentation purposes, it would be good to know what the defaults are for everything too. In one place.

website/utils/url_utils.ts
```diff
@@ -2,16 +2,56 @@ import FirebaseRequest, { DatasetMetaData } from "../../public/firebase";
...
+  lut?: string;
+  /** Volume enabled. "1" is enabled. */
+  ven?: "1" | "0";
+  /** Isosurface enabled. "1" is enabled. */
+  sen?: "1" | "0";
+  /** Isosurface value, in the [0, 255] range. */
+  isv?: string;
+};
+
+const baseParamKeys = ["mask", "url", "file", "dataset", "id", "view"] as const;
```
toloudis, line 36: seeing this, I am trying to remember what's the difference between file and url?  (Another documentation issue).
Dataset is an identifier for a Cell Feature Explorer dataset that is stored in our firebase db, and I believe id is an image id that's a member of dataset.

toloudis (approved):
Although I didn't explicitly test, the implementation seems good - my only comments are for documentation/more comments.   Good job adding those unit tests.

website/utils/url_utils.ts
ShrimpCryptid: Included, thanks!

website/utils/url_utils.ts
ShrimpCryptid: It looks like `file` pulls directly from your file server? Currently, it appends the contents of `file` to `http://dev-aics-dtp-001.corp.alleninstitute.org/dan-data/` as a base URL.

The comments near it reads:
```
// quick way to load a atlas.json from a special directory.
//
// ?file=relative-path-to-atlas-on-isilon
```

Should I remove it?

toloudis: Thanks -- yes this seems like a reasonable time to remove it or put it in the deprecated list.  It's just an extra thing to maintain for no reason and there's no good reason why that url should be hardcoded into this app.

website/utils/url_utils.ts
```diff
@@ -13,23 +13,38 @@ const LUT_REGEX = /^[a-z0-9.]*:[ ]*[a-z0-9.]*$/;
...
+  /** Isosurface alpha, in the [0, 1 range]. Set to `1.0` by default.*/
   isa?: string;
   /** LUT to map from intensity to opacity. Should be two alphanumeric values separated
-   * by a colon, e.g. "0:255", "p50:p90", or "autoij:0".
+   * by a colon. The first value is the minimum and the second is the maximum.
+   * Defaults to [0, 255].
+   *
+   * - Plain numbers are treated as direct intensity values.
+   * - `p{n}` represents a percentile, where n is a percentile in the [0, 100].
+   * - `m{n}` represents the median multiplied by n.
```
toloudis, line 30: i think it's really multiplying the median intensity by n/100 (so that the code didn't have to deal with decimals)

toloudis: I just noticed this is also documented in ViewerChannelSetting too.

Issue #80, opened 2022-10-03
Tighten up design of channel settings


frasercl:
## Use Case

The expanded channel settings menu could use a little bit of design polish to clarify margins between elements. In particular:
- The isosurface settings sliders use ant's `Slider` component while the colorize slider uses `Nouislider`.
- Text elements like the colorize slider's tooltip and the opacity slider's title have too little bottom margin and intersect with the elements below them.
- The amount of empty space to the right of controls is inconsistent.

Screenshot:
![image](https://user-images.githubusercontent.com/53030214/193702146-0eee3cc0-a4ef-486b-949b-09953c399e6a.png)

## Solution

Modify the appearance of this component to be more consistent with the design of the rest of the app. Work from guidance and feedback from @lynwilhelm if she has the bandwidth.

Design Link: https://www.figma.com/file/yGTPga2j961NKfQT7xTird/2023-Sept-Settings-Redesign?type=design&node-id=1%3A2&mode=dev

lynwilhelm:
Yes, let me know if you need any input! @toloudis @frasercl Do we want to include the updated designs I have for the checkboxes and settings in this work?

It's shown in the panel settings design file:
https://xd.adobe.com/view/904d6444-a800-48b6-8844-4f6f116d85ab-c51e/

Screenshot:
<img width="677" alt="Screen Shot 2022-10-04 at 9 43 30 AM" src="https://user-images.githubusercontent.com/93954795/193879570-496eab9a-45f8-433d-b738-0463e0848d30.png">

toloudis:
@lynwilhelm  I agree with including the updated designs you mentioned above and we should get this on the ux backlog, just to have those controls cleaned up and/or styled.  This section of the ui gets into some interesting questions of what controls are actually needed, and are they really effective as they are,  but maybe the first pass can just be an improved layout. I don't know.

toloudis:
**Background**

We have done a big overhaul of the styling of the viewer, with an eye toward making it usable as a standalone app, and also embeddable in other environments like CFE or jupyter notebook.
There is one area of the user interface that has not been re-examined yet.

**User needs statement(s)**
The channel settings look less polished than the rest of the user interface.
The controls within the channel settings sometimes have confusing interactions.

**Size**
??

**Requirements**
When channel settings are opened, they have a consistent and clean appearance with the rest of the viewer.

**Deliverables**
[...]

PR #254, opened 2024-05-28
Placeholder for disabled channel settings

### LINKED ISSUE #246: Channel settings are not well ordered in HTML
## Description

Channel settings elements appear out of logical order in HTML, making tab-navigating confusing. Elements are grouped like this, in the order marked:

image

This is especially problematic given that the last element of box 3 controls whether box 2 is accessible. A user that expects to open the settings menu with that control then tab into it must instead counterintuitively backtrack past the visibility checkboxes to get at the *end* of the channel settings.

## Expected Behavior

Elements should appear in the correct order and navigating with the keyboard into the advanced channel settings should be intuitive.

## Reproduction

1. Click the "open channel settings" button on any channel.
2. Try to access and adjust that channel's settings with the keyboard (Tab, Shift+Tab, etc.)

## Environment

Firefox, but this would appear in any environment.

frasercl:
Advances #80: adds placeholders for volume/surface settings when a channel's volume/surface is disabled, and allows opening channel settings when the channel is completely invisible.

Also does some related cobweb-clearing and maintenance:
- Converts `ColorPicker` to a function component
- Following discussion with @lynwilhelm, adds a tighter, lighter border around color picker swatches
- Closes #246: puts elements within a channel row in the expected order
- Cleans out a lot of unused or redundant css and css-in-js

Design: https://www.figma.com/design/yGTPga2j961NKfQT7xTird/2023-Sept-Settings-Redesign?node-id=1-2&t=ndIJ40DdmIMNtKgC-0

### Screenshots

| Design | This branch |
| ----- | ----- |
| image | image |


toloudis (approved):
I love the net negative lines of code too.

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -177,52 +161,45 @@ const ChannelsWidgetRow: React.FC<ChannelsWidgetRowProps> = (props: ChannelsWidg
...
 
-  const rowClass = controlsOpen ? "row-card" : "row-card controls-closed";
+  const rowClass = controlsOpen ? "channel-row" : "channel-row controls-closed";
   return (
-    <List.Item key={index} className={rowClass} extra={renderActions()}>
+    <List.Item key={index} className={rowClass}>
       <List.Item.Meta title={<span style={STYLES.channelName}>{props.name}</span>} avatar={createColorPicker()} />
-      {controlsOpen && renderControls()}
+      {visibilityControls}
+      {controlsOpen && <div style={{ width: "100%" }}>{renderControls()}</div>}
```
interim17, line 191: I like how clean this is, but have been guided away from inline styling in the past.

This is not a request for change, but just curious your thought process in deciding to use inline here.

interim17 (approved):
Nice work :)

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
frasercl: This codebase is on the older side and some components have an odd pattern of doing styling with big JS objects laid out like stylesheets at the bottom of .tsx files. (I ended up having to add a type for these objects, `Styles`, when I converted the project to TS.) Though I don't take a super hard line on inline styling, I agree it's not great practice and am not really pleased to be doing it here. I just didn't try to break this file all the way out of that pattern... for now. I think this particular case was a sort of half-measure for a style record that I was able to trim down pretty far but not completely get rid of.

PR #256, opened 2024-05-30
Reorganize channel settings

frasercl:
Review time: moderate (20-30min)

Advances #80: implements most of @lynwilhelm's newer, cleaner channel settings design except adding "advanced" mode to transfer function editor.

Along the way, makes these more behind-the-scenes changes:
- Created `SliderRow` shared component for everywhere the control panel has a slider with a title (a fair few places!)
- Continued removing and consolidating styling
- When editing the color of a control point in the transfer function editor, the color picker is now quite a bit smarter about where on the screen to appear
- Converted `GlobalVolumeControls` and the `SmarterSlider` utility component to function components

### Screenshots!
| Before | Design | After |
| ----- | ----- | ----- |
| image | image | image |

(Note that a user is very unlikely to want to have both "Volume settings" and "Surface settings" open at once)


toloudis (approved):
Looks good!

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -127,10 +99,9 @@ const ChannelsWidgetRow: React.FC<ChannelsWidgetRowProps> = (props: ChannelsWidg
     return (
       <TfEditor
         id={"TFEditor" + index}
-        index={index}
         fit-to-data={false}
-        width={250}
-        height={150}
+        width={418}
```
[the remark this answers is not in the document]
frasercl: Yeah, this is roughly the width of the control panel's content box (its "actual width" minus padding)

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -144,21 +115,20 @@ const ChannelsWidgetRow: React.FC<ChannelsWidgetRowProps> = (props: ChannelsWidg
...
-    </Col>
+    <div>
+      <SliderRow label="Isovalue" max={255} start={ISOVALUE_DEFAULT} onChange={onIsovalueChange} formatInteger={true} />
+      <SliderRow
+        label="Opacity"
+        max={ISOSURFACE_OPACITY_SLIDER_MAX}
+        start={ISOSURFACE_OPACITY_DEFAULT * ISOSURFACE_OPACITY_SLIDER_MAX}
+        onChange={onOpacityChange}
+        formatInteger={true}
+      />
```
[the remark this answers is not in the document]
frasercl: I generally prefer components to create functions anyways - it's much nicer semantically (and does just about the same thing anyways), as long as the added ceremony of writing a props type doesn't get in the way. Though in this case, the primary motivation was seeing the same pattern (label text + slider) used in three separate components and wanting a single source of truth for what that looks like.

src/aics-image-viewer/components/shared/SliderRow/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const INTEGER_FORMATTER = { to: Math.round, from: Number };
+
+/** A component to ensure a single unified style across the many labeled slider rows in the control panel */
+const SliderRow: React.FC<SliderRowProps> = (props) => (
+  <div className="viewer-control-row">
+    <div className="control-name">{props.label}</div>
+    <div className="control">
+      {props.start === undefined
+        ? props.children
```
ShrimpCryptid, line 26: Is there ever a situation where `props.start` is undefined?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -739,66 +696,46 @@ export default class MyTfEditor extends React.Component<MyTfEditorProps, MyTfEdi
...
-          </Button>
-          <Button id={`auto98-${id}`} className="ant-btn" onClick={this.auto98XF}>
-            Auto_98
-          </Button>
-        </div>
+        )}
+        <svg className="tf-editor-svg" width={width} height={height} ref={this.svgElement} />
+        <SliderRow
+          label={
+            <Checkbox checked={colorizeEnabled} onChange={(e) => this.props.updateColorizeMode(e.target.checked)}>
```
ShrimpCryptid, line 724: It's slick that you can directly put a Checkbox into the label field here. Nice!

ShrimpCryptid (approved):
Looks good! This is really clean :'))

src/aics-image-viewer/components/shared/SliderRow/index.tsx
frasercl: When this component has `children`, they replace the slider that the component otherwise renders itself. I did this so I could use the same component for the "interpolate" control:
image
Without a slider, `start` doesn't make much sense, so it's optional. I null-check it here rather than `children` for type-checking reasons.

src/aics-image-viewer/assets/styles/variables.css
```diff
@@ -14,5 +14,5 @@ $dimgray: #6e6e6e;
 $text-gray: #a0a0a0;
 $header-gray: #4b4b4b;
 $light-gray: #aeacae;
-$border-color: $dimgray;
+/* $border-color: $dimgray; */
```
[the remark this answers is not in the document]
frasercl: Gone! There's a lot of other commented-out variables in here, we're so close to being able to just remove this file

src/aics-image-viewer/components/shared/SliderRow/index.tsx
toloudis: This comment makes it seem like checkbox is just some kind of special-case exception to sliders.  It would be interesting to think of these rows as generic "input rows" where there's a label and an input control that can be anything.

PR #258, opened 2024-05-31
Feature/header style tweaks

### LINKED ISSUE #255: Update minor items in header
## Use Case

There are a few UI items in the header that are either a little off from intended design or could be improved. They are small but I think refines the look a little.

Here are the current followed by the adjusted one, which is pretty hard to tell the difference in this format and will list out.
Screenshot 2024-05-29 at 1 12 53 PM
Screenshot 2024-05-29 at 1 12 37 PM

- Left and right padding on header container should be 20px. It's currently 30px.
- Left and right padding on all buttons globally should be 13px. They are currently 15px.
- Space between Help button and Share button should be 12px. It is currently 15px.
- Would love to suggest the site title color change from purple to white with no hover state except the cursor:pointer. In retrospect since we are only dealing with text, I feel like the large size in combination with the purple is overkill and looks slightly tacky or amateur.
- The site title should be updated to open in a new tab instead of current tab when the user is within the tool. This is consistent with a recent rule I decided on for Public File Explorer where any content page when in the tool should not destroy what they are currently (possibly) engaging with.

frasercl:
Review time: tiny (~5min)

Resolves #255: implements a list from @lynwilhelm of minor style tweaks in the header.


interim17 (approved):
LGTM

toloudis (approved):

PR #262, opened 2024-06-08
Viewer state management refactor

frasercl:
Review time: medium-large

Refactors state representation out of `App` in favor of a new component, `ViewerStateProvider`.

Some background. This repo is, in a sense, concerned primarily with providing a nice UI for manipulating viewer state. By viewer state, I mean any state which is in some sense a setting of volume-viewer - so whether the bounding box is visible _is_ viewer state, but whether the control panel is open is _not_. Currently, this state is stored and managed like this:

- Viewer state lives in a couple big state objects owned by our top-level UI component (the one we export from this package) which is called `App`.
- Alongside all of `App`'s other responsibilities - managing image loading, updating the actual viewer, keeping track of basic UI state - we define several setter functions which keep us a bit disciplined about updating viewer state.
- Viewer state and relevant setters are made available to the rest of the UI by drilling each value and function through a chain of props until it gets to the component that needs it.
- There is no good way to access viewer state from outside `App`, which is a problem for upcoming work on populating URLs with viewer state for sharing.
- There is also no great way to set viewer state from outside `App`. While `App` has props for setting viewer state, they only set initial state and do nothing when the component is mounted.

The changes in this PR move us to managing state like this:
- State and setter functions are provided by a new component, `ViewerStateProvider`.
- This component provides state via a React context, and therefore can be lifted to anywhere in the component tree above `App`, allowing access to viewer state outside the viewer component.
- This component also accepts props for viewer settings, which are inspected and synced to state on every render.
  - It currently does _not_ accept settings which apply to specific channels... for reasons. We can discuss if this should be added.
- Components may use a new higher-order component, `connectToViewerState`, to "subscribe" to the props they care about directly, cutting down on prop drilling.


ShrimpCryptid:
Could you clarify who you wanted to review this PR? (I'm planning to because I'll be working off of these changes, but I saw there are four(!) people listed as requested 😆)

src/aics-image-viewer/components/App/savedState.ts
```diff
@@ -1,43 +0,0 @@
...
-import { ShowControls, GlobalViewerSettings } from "./types";
-
-/** Global viewer settings that are not currently controlled by any UI */
-// TODO double-check that this is exhaustive
-interface View3dInternalGlobalSettings {
-  pixelSamplingRate: number;
-  stepSizePrimaryRayVoxels: number;
-  stepSizeSecondaryRayVoxels: number;
-  flipVolume: [number, number, number];
-}
```
ShrimpCryptid, line 11: Might have missed this during my review, but what happened to these settings? Did they get moved or just deleted?

ShrimpCryptid (commented):
Left some minor comments!

src/aics-image-viewer/components/App/savedState.ts
frasercl: This is a pretty old pass I took at viewer state types geared towards some sort of serializable format. I think I included it, maybe accidentally, in a commit a long time ago but never actually used it.

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+  changeMultipleChannelSettings: nullfn,
+  applyColorPresets: nullfn,
+};
+
+const DEFAULT_VIEWER_CONTEXT_OUTER = { ref: { current: DEFAULT_VIEWER_CONTEXT } };
+
+type NoNull<T> = { [K in keyof T]: NonNullable<T[K]> };
+type ContextRefType = NoNull<React.MutableRefObject<ViewerContextType>>;
+
+export const ViewerStateContext = React.createContext<{ ref: ContextRefType }>(DEFAULT_VIEWER_CONTEXT_OUTER);
```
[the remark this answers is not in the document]
frasercl: I did this because, as I learned, refs always include `null` in their value type, since they're originally designed to pull DOM nodes out of the committed component and those nodes might not exist.

website/components/AppWrapper.tsx
```diff
@@ -6,12 +6,12 @@ import LoadModal from "./LoadModal";
...
 import ShareModal from "./ShareModal";
-import { ImageViewerApp, RenderMode, ViewMode } from "../../src";
-import { GlobalViewerSettings } from "../../src/aics-image-viewer/components/App/types";
+import { ImageViewerApp, RenderMode, ViewMode, ViewerStateProvider } from "../../src";
+import { ViewerState } from "../../src/aics-image-viewer/components/ViewerStateProvider/types";
 import { AppDataProps } from "../types";
 import { getArgsFromParams } from "../utils/url_utils";
 
-const DEFAULT_VIEWER_SETTINGS: Partial<GlobalViewerSettings> = {
+const DEFAULT_VIEWER_SETTINGS: Partial<ViewerState> = {
```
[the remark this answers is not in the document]
frasercl: It does appear to be, though the version in `ViewerStateProvider` is more complete (no `Partial`). This one might actually not be needed at all, given that defaults are already applied to viewer settings within the `App` (now the state manager)

rugeli:
> Could you clarify who you wanted to review this PR? (I'm planning to because I'll be working off of these changes, but I saw there are four(!) people listed as requested 😆)

No worries, I haven't started the review and can opt out!

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -15,23 +15,14 @@ import {
...
-import {
-  AppProps,
-  ShowControls,
-  GlobalViewerSettings,
-  ViewerSettingUpdater,
-  ViewerSettingChangeHandlers,
-  UseImageEffectType,
-} from "./types";
+import type { AppProps, ShowControls, UseImageEffectType } from "./types";
+import type { ViewerState, ChannelState } from "../ViewerStateProvider/types";
```
toloudis, line 19: probably should either put a newline after these imports or else group them with the ones below here that don't come from "shared".

toloudis, line 19: probably should either put a newline after these imports or else group them with the ones below here that don't come from "shared".  or just remove the newline after the last "shared" one.  I guess the main distinctions are third party/dependencies, things coming from "shared", and everything else?

toloudis: I think prettier has some way of auto-sorting the imports by rules...

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -782,7 +689,14 @@ const App: React.FC<AppProps> = (props) => {
 
   // Rendering ////////////////////////////////////////////////////////////////
 
-  const showControls = { ...defaultShownControls, ...props.showControls };
+  const showControls = useMemo(
+    (): ShowControls => ({ ...defaultShownControls, ...props.showControls }),
+    [props.showControls]
+  );
```
toloudis, line 695: Why did this become a useMemo of a function?

src/aics-image-viewer/components/App/types.ts
```diff
@@ -1,8 +1,8 @@
-import { View3d, Volume, RawArrayInfo, RawArrayData } from "@aics/volume-viewer";
-import { ImageType, RenderMode, ViewMode } from "../../shared/enums";
-import { PerAxis, MetadataRecord } from "../../shared/types";
-import { ColorArray } from "../../shared/utils/colorRepresentations";
-import { ChannelGrouping, ChannelState, ViewerChannelSettings } from "../../shared/utils/viewerChannelSettings";
+import type { RawArrayData, RawArrayInfo, Volume } from "@aics/volume-viewer";
+
+import type { MetadataRecord } from "../../shared/types";
+import type { ViewerChannelSettings } from "../../shared/utils/viewerChannelSettings";
+import type { ViewerState } from "../ViewerStateProvider/types";
```
toloudis, line 5: thumbs up for using `import type` when it's a type... do we have tsconfig rules to catch this?

src/aics-image-viewer/components/ControlPanel/index.tsx
```diff
@@ -14,7 +14,12 @@ import "./styles.css";
 import ViewerIcon from "../shared/ViewerIcon";
 import { MetadataRecord } from "../../shared/types";
 
-interface ControlPanelProps extends ChannelsWidgetProps, GlobalVolumeControlsProps, CustomizeWidgetProps {
+type PropsOf<T> = T extends React.ComponentType<infer P> ? P : never;
```
toloudis, line 17: Nit: this looks so general I wonder if it could go in a general types module

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+  BOUNDING_BOX_COLOR_DEFAULT,
+  BRIGHTNESS_SLIDER_LEVEL_DEFAULT,
+  DENSITY_SLIDER_LEVEL_DEFAULT,
+  INTERPOLATION_ENABLED_DEFAULT,
+  LEVELS_SLIDER_DEFAULT,
+} from "../../shared/constants";
+import { ColorArray } from "../../shared/utils/colorRepresentations";
+
+const isObject = <T,>(val: T): val is Extract<T, Record<string, unknown>> =>
+  typeof val === "object" && val !== null && !Array.isArray(val);
```
toloudis, line 26: are we really still needing to write our own isObject function in 2024?

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -0,0 +1,83 @@
...
+  value: ChannelState[K]
+) => void;
+
+export type ViewerContextType = ViewerState & {
+  channelSettings: ChannelState[];
+  changeViewerSetting: ViewerSettingUpdater;
+  setChannelSettings: React.Dispatch<React.SetStateAction<ChannelState[]>>;
+  changeChannelSetting: ChannelSettingUpdater;
+  changeMultipleChannelSettings: MultipleChannelSettingsUpdater;
+  applyColorPresets: (presets: ColorArray[]) => void;
```
toloudis, line 82: I want to discuss about what this ViewerContextType is

website/components/AppWrapper.tsx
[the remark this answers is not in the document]
toloudis: I had this same question - would be great to remove if possible, or rename at least.

src/aics-image-viewer/components/App/index.tsx
frasercl: Discussed in a meeting, but so it's here: the profiler revealed that this prop was the only reason that some components continued to render, because this object was being regenerated on every render in order to apply defaults. The `useMemo` ensures we only make a new object when its value might actually change.

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
frasercl: JavaScript 🙃

src/aics-image-viewer/components/App/index.tsx
frasercl: @ShrimpCryptid added a package to nucmorph-colorizer to do this I believe?

src/aics-image-viewer/components/App/types.ts
frasercl: not that I know of, but looks like it can be done: https://stackoverflow.com/questions/71080256/is-there-a-way-to-automatically-fix-import-type-errors-on-typescript-when-usin

website/components/AppWrapper.tsx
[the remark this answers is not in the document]
frasercl: Looks like everything works fine with this removed

src/aics-image-viewer/components/ViewerStateProvider/types.ts
frasercl: Also discussed offline: this represents all the values that go in `ViewerStateContext`, which is all items that this PR moves from being managed by `App` and passed to children via props to being managed by `ViewerStateProvider` and passed via context and the `connectToViewerState` function. Since we discussed that the name was a bit inconsistent and confusing, I just now changed it to `ViewerStateContextType`.

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
ShrimpCryptid: Yup, it's `@trivago/prettier-plugin-sort-imports` and `eslint-plugin-simple-import-sort.` Note that they occasionally disagree, though I've mostly configured them to work the same.

Package.json: https://github.com/allen-cell-animated/timelapse-colorizer/blob/main/package.json#L22
Prettier config: https://github.com/allen-cell-animated/timelapse-colorizer/blob/main/.prettierrc
Eslint config: https://github.com/allen-cell-animated/timelapse-colorizer/blob/main/.eslintrc

ShrimpCryptid (approved):

PR #267, opened 2024-06-11
Fix sliders not updating during playback.

### LINKED ISSUE #264: Time slider does not update on playback
## Description

_A clear description of the bug_

## Expected Behavior

_What did you expect to happen instead?_

## Reproduction

## Environment

frasercl:
Review time: tiny (<5min)

Resolves #264: sliders update correctly on playback again.

`SmarterSlider` is a utility component which wraps the (old) React wrapper around nouislider that we use, preventing it from updating while the slider is sliding. This was done to fix a performance issue (which I was unable to replicate?).

In #256, I converted `SmarterSlider` to a function component, which meant moving from the class component way to prevent updates (`shouldComponentUpdate`) to the function component one (`React.memo`). But I missed when I did this that the function passed to `React.memo` has the opposite expectation of `shouldComponentUpdate`: it should return `true` when the component should _skip_ rendering. So I flipped some booleans around and renamed a couple variables and this component seems to be working as normal again.

### DIFF
--- a/src/aics-image-viewer/components/shared/SmarterSlider.tsx
+++ b/src/aics-image-viewer/components/shared/SmarterSlider.tsx
@@ -4,23 +4,23 @@ import Nouislider, { NouisliderProps } from "nouislider-react";
 type CallbackArgs = Parameters<NonNullable<NouisliderProps["onStart"]>>;
 
 const MemoedNouislider = React.memo(
-  Nouislider as React.ComponentType<NouisliderProps & { shouldUpdate: boolean }>,
-  ({ shouldUpdate }) => shouldUpdate
+  Nouislider as React.ComponentType<NouisliderProps & { noUpdate: boolean }>,
+  ({ noUpdate }) => noUpdate
 );
 
 /** A wrapper around `Nouislider` that prevents updates while the slider is being dragged. */
 const SmarterSlider: React.FC<NouisliderProps> = (props) => {
-  const [shouldUpdate, setShouldUpdate] = React.useState(true);
+  const [noUpdate, setNoUpdate] = React.useState(false);
   const wrapEventHandler = (shouldUpdate: boolean, handler?: (...args: CallbackArgs) => void) => {
     return (...args: CallbackArgs) => {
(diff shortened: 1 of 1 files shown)


ShrimpCryptid (approved):

src/aics-image-viewer/components/shared/SmarterSlider.tsx
```diff
@@ -4,23 +4,23 @@ import Nouislider, { NouisliderProps } from "nouislider-react";
...
 );
 
 /** A wrapper around `Nouislider` that prevents updates while the slider is being dragged. */
 const SmarterSlider: React.FC<NouisliderProps> = (props) => {
-  const [shouldUpdate, setShouldUpdate] = React.useState(true);
+  const [noUpdate, setNoUpdate] = React.useState(false);
   const wrapEventHandler = (shouldUpdate: boolean, handler?: (...args: CallbackArgs) => void) => {
     return (...args: CallbackArgs) => {
-      setShouldUpdate(shouldUpdate);
+      setNoUpdate(shouldUpdate);
```
toloudis, line 16: should the argument here be called `shouldNotUpdate` then?  As in `setNoUpdate(shouldNotUpdate);` ?

toloudis (approved):

toloudis:
Also, does this have implications on update while dragging the slider?  Because currently I think that behavior is correct, before this fix

src/aics-image-viewer/components/shared/SmarterSlider.tsx
frasercl: Knew I was gonna miss something

frasercl:
> Also, does this have implications on update while dragging the slider? Because currently I think that behavior is correct, before this fix

Event handlers that update state will still run and the handle of the slider will still move correctly. That's all handled in nouislider, not nouislider-react. What this prevents is rerenders of the nouislider-react component when the value of the slider changes in props. Those renders basically just imperatively change the position of the slider handle again, even though it was already changed by nouislider itself in response to the user's dragging.

PR #276, opened 2024-06-26
Convert transfer function editor to function component

frasercl:
Review time: medium-large (45-60min). Advances #80.

The remaining work on implementing @lynwilhelm's designs for an updated channel settings interface (available [here](https://www.figma.com/design/yGTPga2j961NKfQT7xTird/2023-Sept-Settings-Redesign?node-id=0-1&t=4kuQqaB30PuGOqYf-1)) involves adding new visual elements and functionality to the transfer function editor. This component is currently implemented as a React class component, though I believe it has been adapted from an older framework at least once, and draws its plot imperatively with D3 outside of React's normal render/commit cycle. Owing to the many lives this component has lived, it now contains a fair bit of unused or contradictory code and tangled imperative updates that are difficult to follow or reason about. Facing the possibility of adding to this tangle, and encouraged by [D3's own example](https://d3js.org/getting-started#d3-in-react) of how the library can integrate more naturally with React's declarative style, I decided it would be worth the time to adapt the component to this style first.

The new component contains about half the code of the old and should be a near-match visually and functionally, though I am aware of at least a few changes:
- Control points no longer animate in on component mount (and in a few other circumstances where new points might appear)
- Control points are no longer clipped off when very near the top of the plot
- The mouse cursor no longer flashes between "move" and the regular pointer while dragging a control point
- I switched from using mouse events to pointer events to handle editing control points. This seems to have made interacting with control points much more performant, though it has also introduced a bug in Firefox where clicking on the plot, dragging off the plot, and releasing over a transfer function preset button is interpreted as a click on that button, causing changes to be overwritten by a generated transfer function. I have no idea why this happens.

### Steps to test:
- `npm i` to update D3
- `npm start` and open an image to test with. I recommend http://localhost:9020/viewer?url=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Frelease%2Fnuc-morph-data%2Fdata_for_analysis%2Fbaseline_colonies%2F20200323_09_small%2Fraw.ome.zarr
- Open at least one channel's settings menu by clicking the icon of 3 sliders to the right of the channel's name
- Verify that everything still works, especially mouse interactions with the plot, which include:
  - Clicking on an existing control point and dragging it around
  - Clicking not on a control point to create a new one (and dragging it around)
  - Right-clicking on a control point to pick its color


ShrimpCryptid:
image

FYI I'm getting some text selection behavior when I click and drag on a point in Chrome. If text is selected, dragging the control point again causes the 🚫 mouse icon to appear.

ShrimpCryptid (approved):
Looks good and feels great to interact with. Left a comment about some potentially unwanted selection behavior that might be worth fixing (though I am seeing it on the original viewer too).

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -13,56 +13,39 @@ import {
...
+import { useRefWithSetter } from "../../shared/utils/hooks";
 
 export const TFEDITOR_DEFAULT_COLOR: ColorArray = [255, 255, 255];
-const COLOR_PICKER_MARGIN = 2;
-
-type Pair = [number, number];
+const TFEDITOR_COLOR_PICKER_MARGIN = 2;
+const TFEDITOR_GRADIENT_MAX_OPACITY = 0.9;
+const TFEDITOR_NUM_TICKS = 4;
+const TFEDITOR_MAX_BIN = 255;
```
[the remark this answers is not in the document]
frasercl: I'm actually not sure - it's possible that full dynamic range data could still be binned down into 256 bins. @toloudis may have a better sense though.

frasercl:
> FYI I'm getting some text selection behavior when I click and drag on a point in Chrome. If text is selected, dragging the control point again causes the 🚫 mouse icon to appear.

I saw this as well and tried to resolve it with various combos of `stopPropagation`s, `preventDefault`s, and handlers for new events (`onMouseDown`, `onClick`, etc.) which just contained `stopPropagation`s and `preventDefault`s. With the current combo I'm unable to replicate the highlighting issue on Chrome, so if you're on the most up-to-date version of this branch and still seeing the issue I may want to investigate it further on your machine.

frasercl:
Following up on the above: turns out I was using a slightly older version of Chrome (122 I believe). Upgrading to Chrome 126 introduced the text highlighting issue for me as well. The CSS changes added in my most recent commit address the issue.

src/aics-image-viewer/components/TfEditor/index.tsx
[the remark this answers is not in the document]
toloudis: Just replying to the discussion but not really adding any value to this PR:  Number of bins is not the same as max intensity.  This is one of the things we have to be careful about when moving to full dynamic range data.  So often you end up with having N bins, and if the bins are equally spaced then you have the first bin containing intensities in the set `[min, (min + (max-min)/N) )` and the second bin spans `[(min + (max-min)/N), (min + 2*(max-min)/N) )`,  and so on up to `[max- (max-min)/N), max)`. and because I wasn't precise enough, you have to know whether to use max or max+1, due to the open-ended upper bound on each bin. Then you have to decide how the lookup table maps to intensities based on the histogram binning bla bla bla

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -82,669 +67,313 @@ const TF_GENERATORS: Record<string, (histogram: Histogram) => Lut> = {
...
-
-  componentDidUpdate(prevProps: MyTfEditorProps): void {
-    const { volumeData } = this.props;
-
-    this.redraw();
-    if (prevProps.volumeData !== volumeData) {
-      this.redrawHistogram();
-    }
-  }
+/** Defines an SVG gradient with id `id` based on the provided `controlPoints` */
```
toloudis, line 70: awesome

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -82,669 +67,313 @@ const TF_GENERATORS: Record<string, (histogram: Histogram) => Lut> = {
...
-  private canvasContext(): CanvasRenderingContext2D | null {
-    let canvas_element = this.canvas.current;
-    if (canvas_element) {
-      return canvas_element.getContext("2d");
-    }
-    return canvas_element;
-  }
+  // Either `null` when the control panel is closed, or an x offset into the plot to position the color picker.
+  // Positive: offset right from the left edge of the plot; negative: offset left from the right edge of the plot.
+  const [colorPickerPosition, setColorPickerPosition] = useState<number | null>(null);
```
toloudis, line 115: does this also need an "above"/"below" designation too?  For when the tf editor is near the bottom or top of the webpage?  Or does the color picker component already have that logic?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -82,669 +67,313 @@ const TF_GENERATORS: Record<string, (histogram: Histogram) => Lut> = {
...
+      .x((d) => xScale(d.x))
+      .y0((d) => yScale(d.opacity))
+      .y1(innerHeight)
+      .curve(d3.curveLinear);
+    return areaGenerator(props.controlPoints) ?? undefined;
+  }, [props.controlPoints, xScale, yScale, innerHeight]);
+
+  // The below `useCallback`s are used as "ref callbacks" - passed as the `ref` prop of SVG elements in order to render
+  // these elements' content using D3. They are called when the ref'd component mounts and unmounts, and whenever their
+  // identity changes (i.e. whenever their dependencies change).
```
toloudis, line 243: cool

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -82,669 +67,313 @@ const TF_GENERATORS: Record<string, (histogram: Histogram) => Lut> = {
...
+
+  const yAxisRef = useCallback(
+    (el: SVGGElement) => d3.select(el).call(d3.axisLeft(yScale).ticks(TFEDITOR_NUM_TICKS)),
+    [yScale]
+  );
+
+  const histogramRef = useCallback(
+    (el: SVGGElement) => {
+      if (el === null) {
+        return;
```
toloudis, line 262: when is it null and when it becomes non-null does this code get re-run?

toloudis (commented):
couple of questions.  I am assuming that this is more performant than the previous implementation due to how it plays better with pure Reactness. I'll download the code and give it a try. 

 It would be good to have tickets and/or a checklist for the next implementation details to clean up the remaining known issues.

src/aics-image-viewer/components/TfEditor/index.tsx
frasercl: The behavior of ref callbacks is that the function will be called with `null` whenever the component unmounts - in this case, whenever volume settings are closed. When the component remounts, they get called with an element again.

src/aics-image-viewer/components/TfEditor/index.tsx
frasercl: Currently, the color picker appears at a fixed Y position relative to the plot... which a quick test shows is a problem when the plot is near the bottom of the screen. I can take a pass at fixing that.

src/aics-image-viewer/shared/utils/hooks.ts
```diff
@@ -12,6 +12,28 @@ export function useStateWithGetter<T>(initialState: T | (() => T)): [T, (value:
   return [state, wrappedSetState, getState];
 }
 
+/**
+ * Wraps a setter function and keeps a ref updated to follow the set value. Useful for making the most up-to-date value
+ * of some state accessible to a closure that might be called after the value is updated.
+ */
+// TODO should this replace `useStateWithGetter`?
+export function useRefWithSetter<T>(setter: (value: T) => void): [MutableRefObject<T | undefined>, (value: T) => void];
+export function useRefWithSetter<T>(setter: (value: T) => void, init: T): [MutableRefObject<T>, (value: T) => void];
```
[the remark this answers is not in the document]
frasercl: These are how TypeScript handles overloads - you can't have JavaScript pick between multiple implementations based on types that it doesn't know about, but you can provide multiple signatures followed by a single implementation that encapsulates all of them and handles the specified cases.

src/aics-image-viewer/components/TfEditor/index.tsx
toloudis: I think we already have that logic for the channel colors.

toloudis (approved):

PR #287, opened 2024-07-09
Use reducers for viewer state

frasercl:
Review time: small (10-15min)

## Problem

Currently, if we try to update multiple viewer or channel settings synchronously, only the last update will be applied:

```typescript
changeViewerSetting("maskAlpha", 0.5); // overwritten by next update
changeViewerSetting("autorotate", true); // only this update is applied
```

This is due to the fact that setters like `changeViewerSetting` produce a new value for `viewerSettings` by pulling and updating the current value in state. But since that state value only gets updated on every render, multiple synchronous updates will end up pulling and modifying the same stale state value. So e.g. above, the first call produces a new value for `viewerSettings` by copying the current value and updating `maskAlpha`, then the second call pulls the same value from _before the `maskAlpha` update_ and ends up overwriting that pending update!

## Solution

`useReducer` solves this problem. Rather than pulling and updating state ourselves, `useReducer` lets us define a function which accepts the current state and an action to perform and returns a modified state value. Since the current state value gets passed in from within the bowels of React rather than from the static state value React provides us for a single render, multiple updates can be applied successfully between renders.

### DIFF
--- a/src/aics-image-viewer/components/ViewerStateProvider/index.tsx
+++ b/src/aics-image-viewer/components/ViewerStateProvider/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useContext, useMemo, useReducer, useRef, useState } from "react";
+import React, { useCallback, useContext, useMemo, useReducer, useRef } from "react";
 
 import type {
   ViewerStateContextType,
@@ -7,7 +7,6 @@ import type {
   ViewerSettingUpdater,
   ChannelSettingUpdater,
   ChannelState,
-  MultipleChannelSettingsUpdater,
   PartialIfObject,
 } from "./types";
 import { ImageType, RenderMode, ViewMode } from "../../shared/enums";
@@ -97,6 +96,35 @@ const viewerSettingsReducer = <K extends keyof ViewerState>(
   }
 };
 
+type ChannelStateAction<K extends keyof ChannelState> =
+  | {
+      index: number | number[];
+      key: K;
+      value: PartialIfObject<ChannelState[K]>;
+    }
+  | {
+      index?: undefined;
+      key: K;
+      value: PartialIfObject<ChannelState[K]>[];
+    };
+
+const channelSettingsReducer = <K extends keyof ChannelState>(
+  channelSettings: ChannelState[],
+  { index, key, value }: ChannelStateAction<K>
+): ChannelState[] => {
+  if (index === undefined) {
+    return channelSettings.map((channel, idx) => {
+      return value[idx] ? { ...channel, [key]: value[idx] } : channel;
+    });
+  } else if (Array.isArray(index)) {
+    return channelSettings.map((channel, idx) => (index.includes(idx) ? { ...channel, [key]: value } : channel));
+  } else {
+    const newSettings = channelSettings.slice();
+    newSettings[index] = { ...newSettings[index], [key]: value };
+    return newSettings;
+  }
+};
+
 const nullfn = (): void => {};
 
 const DEFAULT_VIEWER_CONTEXT: ViewerStateContextType = {
@@ -105,7 +133,6 @@ const DEFAULT_VIEWER_CONTEXT: ViewerStateContextType = {
   changeViewerSetting: nullfn,
   setChannelSettings: nullfn,
   changeChannelSetting: nullfn,
-  changeMultipleChannelSettings: nullfn,
   applyColorPresets: nullfn,
 };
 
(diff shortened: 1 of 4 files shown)


src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -105,6 +95,49 @@ const applyChangeToViewerSettings = <K extends keyof ViewerState>(
...
+      index?: undefined;
+      key: K;
+      value: ChannelState[K][];
+    }
+  | {
+      // Initialize channel settings
+      index?: undefined;
+      key?: undefined;
+      value: ChannelState[];
+    };
```
toloudis, line 116: This is sort of in the Nit category but see what you think:
What I'm used to seeing in redux land is that you'd actually name these actions like `InitializeChannelSettingsAction` for the last one, and then call `channelDispatch(new InitializeChannelSettingsAction({args or payload}));`.  I might suggest that the code is a bit more readable at the point of call when you don't have to overload one single type with so many variants.

Then the reducer could check the type of the action it receives (handwaving? my memory is bit vague on this)... 

You could still have some notion of a ChannelStateAction as a base type which has a index,key,and value.

frasercl: Definitely worth making the type definition clearer by splitting action types into named variants - I've just pushed a commit that does this, check it out - but I'm not sure about how much benefit we might gain from a readability standpoint with the other changes.

As I have it set up, we're missing two major parts of the redux pattern: actions with named types at runtime, and action creators. The main thing you're asking for above is action creators, though you've written it as a class constructor when I believe they're generally implemented as plain old functions. To work with an example, using action creators would turn this definition:

```typescript
const applyColorPresets = useCallback((value: ColorArray[]): void => channelDispatch({ key: "color", value }), []);
```
...into something like...
```typescript
const applyColorPresets = useCallback((value: ColorArray[]): void => channelDispatch(createChannelSettingArrayUpdateAction("color", value)));
```

Trouble is, this pattern starts introducing some of the extra ceremony that people love to grumble about in vanilla redux. Next to the type definition for `ChannelSettingArrayUpdateAction` (added in my last commit), we'd need a definition for `createChannelSettingArrayUpdateAction`:

```typescript
type ChannelSettingArrayUpdateAction<K extends keyof ChannelState> = {
  key: K;
  value: ChannelState[K][];
};
const createChannelSettingArrayUpdateAction<K extends keyof ChannelState> = (
  key: K,
  value: ChannelState[K][],
): ChannelSettingArrayUpdateAction<K> = ({ key, value });
```

That's a lot of redundant type info for a function that does nothing more than bundle its arguments into an object.

The other thing we're missing, which your mention of "the reducer could check the type of the action it receives" above gets at, is explicit action types at runtime. In redux, all actions are required to have a `type` field which names the type as a string literal. React lets you use any old type you want for actions (the react type definitions file actually explicitly mentions this difference with redux). I decided not to use a `type` field in the set of channel state actions, again because it felt unnecessary. The set of actions we want to take with channel settings isn't particularly complex: basically just "initialize" and a couple variants of "update a single setting." The reducer currently does type-check between different kinds of action here, it just does it based on missing properties. E.g. if the action is missing a `key` property, then it's not trying to update a single setting, it's trying to replace the entire state with a new initial value. One of my goals here was treating redux-flavored code not as a pattern the whole app would adopt, but as a necessary bit of complexity in the internal workings of just this one component, so it made some sense to me to take this less explicit approach. The new types layout tries to makes it clearer that the reducer is type-checking between multiple variants of `ChannelStateAction` - let me know if it works.

The way I see it, I'm happy with a nicer types definition, but we could decide to do a couple more things for readability:

1. Just bite the bullet and add a `type` field. This could make things clearer in lieu of action creators, since you'd at least be naming the kind of action you're taking at the `dispatch` call site and naming the type you're checking for in the reducer explicitly rather than in a comment.
2. Your use of a class constructor above actually suggests some syntax we could maybe use to unify action definition and action creator definition, e.g.:
```typescript
class ChannelSettingArrayUpdateActionClass<K extends keyof ChannelState> {
  index?: never;
  constructor(public key: K, public value: ChannelState[K][]) {}
}
```
Then we can use `new` syntax as you suggested above. It's not necessarily traditional redux, but again, react lets us do whatever we want with this pattern in a way redux doesn't.

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -105,6 +95,53 @@ const applyChangeToViewerSettings = <K extends keyof ViewerState>(
...
+type ChannelSettingInitAction = {
+  value: ChannelState[];
+};
+
+type ChannelStateAction<K extends keyof ChannelState> =
+  | ChannelSettingUniformUpdateAction<K>
+  | WithExplicitlyUndefined<"index", ChannelSettingArrayUpdateAction<K>>
+  | WithExplicitlyUndefined<"index" | "key", ChannelSettingInitAction>;
+
+const channelSettingsReducer = <K extends keyof ChannelState>(
```
rugeli, line 122: I like how the function manages different actions based on the `key` and `index` properties, effectively covering various updates. 

nit: are there potential cases where the index could be out of bounds or the key might be invalid?

rugeli (approved):
I learned another React hook today! The use of `userReducer` to handle state updates for viewer and channel settings is a good approach! LGTM

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
toloudis: Thanks for this great reply.  I feel like maybe the class thing you mention at the end might help readability a tiny bit without introducing too much silly boilerplate.  Just having names for the three variants does help but I think it would help even more at the point where you first produce the data values, in the dispatch calls.  But at the same time I think I would be fine with the code in its original state too.

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
frasercl: TypeScript should keep us from ever using an invalid key, but it's definitely possible to try to index the channel settings array out of bounds. I added a check!

Issue #281, opened 2024-07-03
Moving a control point causes other control point to jump


toloudis:
## Description

when clicking to drag a control point, another control point will shift its position

## Expected Behavior

expect that moving a control point will not affect the displayed positions of any others

## Reproduction

https://dev-aics-dtp-001.int.allencell.org/website-3d-cell-viewer/viewer?url=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Fholistic%2Fendos%2Ffeasibility%2Fcdh5.ome.zarr

go to frame 62.  
click IJ Auto
play a few frames until you see the right most control point move left so its x coordinate is less than 255.
pause
click the rightmost controlpoint to drag it and you will see the other control point jump to a different position.

## Environment
Windows, Chrome

frasercl:
Resolved by #294

Issue #131, opened 2023-04-05
idea for redoing the transfer function editor


toloudis:
### Background
#79 
We have per-channel controls that may be excessively complicated for our users.  

### User needs statement(s)
As a microscopist or cell biologist who needs to inspect raw images, I need quick and easy contrast adjustment controls.

### Size
m

### Requirements
I have a specific idea in mind that I want to put down in writing here.
User research has shown that users typically only need to set the "min" and "max" for intensity remapping and rarely (possibly never) will need anything more complicated.  I propose making the histogram control smaller and simpler:
* always only 2 control points showing and users can move only those points. 
* (perhaps an "advanced" mode where user can add control points?  not sure if needed)
* add a menu of color maps, and the ability to create a custom color map.  The color map would be applied between max and min.  A color map is basically just a color gradient. (example: https://docs.paraview.org/en/v5.10.0/_images/ColorMapComboBox.png)
* we still want to preserve "colorize" mode but this might be achievable with just a color map.
* does applying a color map override the single-color per channel?  currently this is used as a multiplier, with an alpha slider to control how much it contributes

### Deliverables
TBD

### Stakeholder(s)
@toloudis @frasercl @lynwilhelm (should add at least one imaging user)

### Timeline
TBD

frasercl:
Captured by/included in #80 - resolve when it resolves

frasercl:
@toloudis confirming: #294 resolves this? There's at least one feature idea here (a color map menu) that isn't included in that.

toloudis:
> @toloudis confirming: #294 resolves this? There's at least one feature idea here (a color map menu) that isn't included in that.

Yes, we can just call this completed.  My intention was to somehow preserve an "easy" way to make a color gradient from min to max.  We can make a new "idea/suggestion" ticket for it or something

frasercl:
Remaining unimplemented suggestion captured in #315. Closing this now.

PR #294, opened 2024-07-11
Basic/advanced mode transfer function editor

### LINKED ISSUE #80: Tighten up design of channel settings

frasercl:
Review time: medium-large (~45min)

Resolves #80 and potentially #131: implements the final piece of @lynwilhelm's [channel settings editor redesign](https://www.figma.com/design/yGTPga2j961NKfQT7xTird/2023-Sept-Settings-Redesign?node-id=1-2&t=RzwAQxJTjkZeTBxI-1) by adding a "basic" mode to the transfer function editor, in which only a min and max of a ramp may be set, and hiding the current control points editor as "advanced" mode.

- Adds two new fields to `ViewerState`: `ramp: [number, number]` for the min and max of the ramp set by "basic" mode, and `useControlPoints: boolean` to indicate whether to set the transfer function based on `ramp` or the existing `controlPoints` field.
- Adds an "advanced" checkbox to the transfer function editor, which controls `useControlPoints`.
- In `TfEditor`'s new "basic" mode:
  - Control points are replaced by two purple sliders for the min and max
  - The horizontal axis labels get pushed down out of the way of the min slider
  - Two additional numeric inputs appear for the min and max
- When a time or region change causes the transfer function to be remapped, the viewer will also remap whichever state is _not_ currently being used to set the transfer function (`controlPoints` in "basic" mode, `ramp` in "advanced" mode) so that it doesn't get out of sync.

### Screenshots

image
image


src/aics-image-viewer/components/App/ChannelUpdater.tsx
```diff
@@ -60,6 +59,20 @@ const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, channelState, vi
     [color]
   );
 
+  const { controlPoints, ramp, useControlPoints } = channelState;
+  useImageEffect(
+    (currentImage) => {
+      if (useControlPoints && controlPoints.length < 2) {
```
toloudis, line 65: what is this if condition checking for? is it for before any channel data has arrived, or something like that? I see it was there in the previous implementation too

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -128,6 +134,8 @@ const DEFAULT_CHANNEL_STATE: ChannelState = {
   isovalue: 128,
   opacity: 1.0,
   color: [226, 205, 179] as ColorArray,
+  useControlPoints: false,
+  ramp: [0, TFEDITOR_MAX_BIN],
```
toloudis, line 138: Not sure if this would make anything cleaner but maybe we could have the control points default to the equivalent of [[0, 0], [TFEDITOR_MAX_BIN, 1]]

toloudis: Also, rather than a bin index, we probably should think of the "x" domain of the transfer function (aka LUT) as intensities.  It's a function of input intensity to output intensity. 

If that incurs a crazy amount of changes then we should save it for a different PR and possibly wait for the full dynamic range modifications. ( also related to putting intensity numbers along the x axis in the UI.)

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -51,14 +48,9 @@ const ChannelsWidgetRow: React.FC<ChannelsWidgetRowProps> = (props: ChannelsWidg
...
-    return (newValue: ChannelState[K]) => changeChannelSetting(index, settingKey, newValue);
-  };
-
-  const _onIsovalueChange = createChannelSettingHandler("isovalue");
-  const onIsovalueChange = ([newValue]: number[]): void => _onIsovalueChange(newValue);
-  const _onOpacityChange = createChannelSettingHandler("opacity");
-  const onOpacityChange = ([newValue]: number[]): void => _onOpacityChange(newValue / ISOSURFACE_OPACITY_SLIDER_MAX);
+  const onIsovalueChange = ([newValue]: number[]): void => changeSettingForThisChannel("isovalue", newValue);
+  const onOpacityChange = ([newValue]: number[]): void =>
+    changeSettingForThisChannel("opacity", newValue / ISOSURFACE_OPACITY_SLIDER_MAX);
```
toloudis, line 53: should this replace all the other usages of `props.changeChannelSetting` in this component, lie the ones above for volumeEnabled and isosurfaceEnabled?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -72,14 +79,36 @@ const TF_GENERATORS: Record<string, (histogram: Histogram) => Lut> = {
   resetXF: (_histo) => new Lut().createFullRange(),
 };
 
+// *---*
+// |   |
+// |   |
+//  \ /
+//   *
+// width: 0.65 * height; height of rectangle: 0.6 * height; height of triangle: 0.4 * height
```
toloudis, line 87: ❤️  love this -- but semi-wondering if a svg does any better here.

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -72,14 +79,36 @@ const TF_GENERATORS: Record<string, (histogram: Histogram) => Lut> = {
...
 /** Defines an SVG gradient with id `id` based on the provided `controlPoints` */
 const ControlPointGradientDef: React.FC<{ controlPoints: ControlPoint[]; id: string }> = ({ controlPoints, id }) => {
   const range = controlPoints[controlPoints.length - 1].x - controlPoints[0].x;
   return (
     <defs>
-      <linearGradient id={id} gradientUnits="objectBoundingBox" spreadMethod="pad" x1="0%" y1="0%" x2="100%" y2="0%">
+      <linearGradient id={id} gradientUnits="objectBoundingBox" spreadMethod="pad" x2="100%">
         {controlPoints.map((cp, i) => {
-          const offset = "" + ((cp.x - controlPoints[0].x) / range) * 100 + "%";
+          const offset = ((cp.x - controlPoints[0].x) / range) * 100 + "%";
```
toloudis, line 111: how about a backtick formatted string here, to defeat the evil "number plus string" expression?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -143,68 +179,93 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
...
+      newControlPoints.splice(draggedIdx + 1, idxRight - draggedIdx - 1);
+    }
+
+    setControlPoints(newControlPoints);
+  };
+
+  const dragRampSlider = (handle: TfEditorRampSliderHandle, x: number): void => {
+    if (handle === TfEditorRampSliderHandle.Min) {
+      const max = rampRef.current[1];
+      setRamp([Math.min(x, max), max]);
```
toloudis, line 210: can you drag the min handle past max and vice versa?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -143,68 +179,93 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
...
-    const idxRight = bisector.right(newControlPoints, draggedPoint);
 
-    if (idxLeft < draggedIdx) {
-      const numPointsToRemove = draggedIdx - idxLeft; // should almost always be 1
-      newControlPoints.splice(idxLeft, numPointsToRemove);
-      draggedPointIdxRef.current -= numPointsToRemove;
-      setSelectedPointIdx(draggedPointIdxRef.current);
-    } else if (idxRight > draggedIdx + 1) {
-      newControlPoints.splice(draggedIdx + 1, idxRight - draggedIdx - 1);
+    if (typeof draggedPointIdxRef.current === "number") {
```
toloudis, line 261: what is the other type that's not a "number"?  should you just check props.useControlPoints here instead?  or maybe `draggedPointIdxRef.current as number` vs `draggedPointIdxRef.current as minmaxsliderenumthingy` in the lines below - for readability

src/aics-image-viewer/components/TfEditor/index.tsx
frasercl: This is an svg! The context works like a canvas, but D3 turns it into an svg data string.

src/aics-image-viewer/components/TfEditor/index.tsx
frasercl: nope - this line ensures that each slider stops at the other slider's position rather than dragging past it.

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -71,3 +71,35 @@ export function initializeLut(
...
 }
+
+export function controlPointsToRamp(controlPoints: ControlPoint[]): [number, number] {
+  if (controlPoints.length < 3) {
+    return [0, TFEDITOR_MAX_BIN];
+  }
+  return [controlPoints[1].x, controlPoints[2].x];
+}
+
+export function rampToControlPoints([min, max]: [number, number]): ControlPoint[] {
```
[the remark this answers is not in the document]
frasercl: good point, it's possible for the ramp sliders to be remapped outside that range

frasercl:
The above 3 commits additionally resolve #281. The fix is quick, and that area of code is changed enough by this PR that fixing on main and merging would have been way more work.

src/aics-image-viewer/components/TfEditor/index.tsx
frasercl: I agree that `props.useControlPoints` would make more semantic sense here. But it also wouldn't cause TypeScript to narrow this type for the following lines, where we either pass `draggedPointIdxRef.current` to a function that takes a `number` or one that takes a `TfEditorRampSliderHandle`.

src/aics-image-viewer/components/App/ChannelUpdater.tsx
frasercl: Looks like this triggers only when the user drags a control point all the way to the left, then all the way to the right, thereby deleting all but one point. The effect is that the transfer function is left in its last valid state rather than getting zeroed out when there's only one control point. Does this behavior make sense?

toloudis: I guess?  It kinda sounds like something that could be handled inside the tfeditor but at least I get the scenario where it could happen.

toloudis (approved):

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -71,3 +71,35 @@ export function initializeLut(
...
+  controlPoints: ControlPoint[],
+  oldRange: [number, number] | undefined,
+  { rawMin, rawMax }: Channel
+): ControlPoint[] {
+  if (oldRange === undefined) {
+    return controlPoints;
+  }
+
+  // TODO: this creates a redundant Uint8Array and algorithmically fills it twice. Can we avoid this?
+  const remapLut = new Lut().createFromControlPoints(controlPoints);
```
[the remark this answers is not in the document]
frasercl: Not currently exported from volume-viewer in a form we can use, but will be once allen-cell-animated/website-3d-cell-viewer#227 merges.

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -71,3 +71,35 @@ export function initializeLut(
   aimg.setLut(channelIndex, lutObject);
   return newControlPoints;
 }
+
+export function controlPointsToRamp(controlPoints: ControlPoint[]): [number, number] {
+  if (controlPoints.length < 3) {
+    return [0, TFEDITOR_MAX_BIN];
+  }
+  return [controlPoints[1].x, controlPoints[2].x];
+}
```
ShrimpCryptid, line 80: What does it mean if there are only two (or fewer) control points? Is one control point always at (0, 0.0), and the other at (255, 1.0)?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -143,68 +179,96 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
...
+
+        const newControlPoints = [...props.controlPoints];
+        newControlPoints.splice(index, 0, point);
+        setControlPoints(newControlPoints);
+      } else {
+        // this click is on an existing point - update current points to ref (may have been remapped since last edit)
+        controlPointsRef.current = props.controlPoints;
+      }
+
+      if (typeof draggedPointIdxRef.current !== "string") {
```
ShrimpCryptid, line 237: When would this be a string value?

ShrimpCryptid (approved):
Left a few questions but otherwise LGTM!

src/aics-image-viewer/components/TfEditor/index.tsx
frasercl: This can either be a number, representing a control point, in "advanced" mode; or either `"min"` or `"max"` representing a slider handle in basic mode. See my response to @toloudis, above:

> ...`props.useControlPoints` would make more semantic sense here. But it also wouldn't cause TypeScript to narrow this type for the following lines...

Since this has now tripped up two people, it might need a comment

toloudis:
> The above 3 commits additionally resolve #281. The fix is quick, and that area of code is changed enough by this PR that fixing on main and merging would have been way more work.

For future, the way to do it with a pr is with a branch off of this branch, and a PR that uses this branch as its base.  Then you just have to be careful not to merge the new pr before the older one.

frasercl:
> For future, the way to do it with a pr is with a branch off of this branch, and a PR that uses this branch as its base. Then you just have to be careful not to merge the new pr before the older one.

This occurred to me mere moments after pushing the commits onto this branch 🙃

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
frasercl: Took this callout as an opportunity to make this function more robust (I knew I wasn't quite dotting all my i's the first time around) but yeah: all the lut generator functions in volume-viewer add those two points at the ends.

PR #300, opened 2024-07-26
Use control point remap function

frasercl:
Review time: tiny (<5min)

#294 introduced the need to remap a set of control points manually, independent of volume-viewer's normal update cycle. Currently, this is accomplished by creating an instance of volume-viewer's `Lut` object and remapping it. This is both unnecessarily verbose and inefficient, since this causes the `Lut` to automatically create a `Uint8Array` and fill it in multiple times with lookup table values that will never be used.

allen-cell-animated/volume-viewer#227 exported the underlying function which `Lut` uses to remap control points. This PR picks up that change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "2.9.1",
       "license": "ISC",
       "dependencies": {
-        "@aics/volume-viewer": "^3.9.1",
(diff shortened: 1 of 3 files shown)


rugeli (approved):

ShrimpCryptid (approved):
LGTM!

PR #301, opened 2024-08-03
Lower scale level during playback

### LINKED ISSUE #286: Dynamic adaptive resolution setting
Example: 
User starts playing through time or Z, and resolution (texture atlas size) goes down by 4x until playback is stopped, and then goes back up to what it was before.

Recommend a user setting in the advanced dialog for how much to reduce the resolution (or not at all?) 

Also see https://github.com/allen-cell-animated/volume-viewer/issues/115

frasercl:
Review time: small (5min)

Resolves #286: implements a naive version of dynamic resolution. During playback, load one scale level lower than "ideal" if available.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -278,7 +278,14 @@ const App: React.FC<AppProps> = (props) => {
     loader.current?.setPrefetchPriority(axis ? [axisToLoaderPriority[axis]] : []);
     loader.current?.syncMultichannelLoading(axis ? true : false);
     if (image) {
-      view3d.setScaleLevelBias(image, axis ? 1 : 0);
+      if (axis === null) {
+        // Playback has stopped - reset scale level bias
+        view3d.setScaleLevelBias(image, 0);
+      } else {
+        // Playback has started - decide whether to lower the scale level to speed up loading
+        const shouldDownlevel = axis === "t" || numSlices[axis] !== numSlicesLoaded[axis];
+        view3d.setScaleLevelBias(image, shouldDownlevel ? 1 : 0);
+      }
     }
     setPlayingAxis(axis);
   };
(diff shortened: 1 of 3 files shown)


ShrimpCryptid (approved):

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -206,7 +207,7 @@ const AxisClipSliders: React.FC<AxisClipSlidersProps> = (props) => {
   };
 
   // Pause when view mode or volume size has changed
-  useEffect(() => props.playControls.pause(), [props.mode, ...Object.values(props.numSlices)]);
+  useEffect(() => props.playControls.pause(), [props.mode, props.imageName]);
```
toloudis, line 210: the comment still says "volume size".  how is imageName a proxy for that?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -277,6 +277,16 @@ const App: React.FC<AppProps> = (props) => {
   playControls.onPlayingAxisChanged = (axis) => {
     loader.current?.setPrefetchPriority(axis ? [axisToLoaderPriority[axis]] : []);
     loader.current?.syncMultichannelLoading(axis ? true : false);
+    if (image) {
+      if (axis === null) {
+        // Playback has stopped - reset scale level bias
+        view3d.setScaleLevelBias(image, 0);
+      } else {
+        // Playback has started - decide whether to lower the scale level to speed up loading
+        const shouldDownlevel = axis === "t" || numSlices[axis] !== numSlicesLoaded[axis];
```
toloudis, line 286: so we should downlevel if the dimension of the playback axis changed?  Not sure I understand when that condition arises. Maybe the comment should say why, too...

toloudis (commented):
I love the clean changes for down-leveling but don't understand a couple of the conditions you have set up.

src/aics-image-viewer/components/App/index.tsx
frasercl: "if the axis we're playing through is time, we should always downlevel; if it's spatial, we should only downlevel if we don't already have the entire dimension in memory." This is to prevent an unnecessary load if we're playing through x or y, e.g. I'll add that as a comment.

src/aics-image-viewer/components/AxisClipSliders/index.tsx
toloudis: I see - could it even be the actual Volume object ref too? that's ultimately the identity of the image

frasercl: ...it could be 🤔

PR #312, opened 2024-08-16
fix: Shorten ramp and slice parameters

ShrimpCryptid:
Missed float formatting/URL shortening for two parameters, the region and slice! 

This change adds formatting to both and a unit test to validate the result.

*Estimated review size: tiny, 5 minutes*

### DIFF
--- a/website/utils/test/url_utils.test.ts
+++ b/website/utils/test/url_utils.test.ts
@@ -483,16 +483,15 @@ describe("Viewer state serialization", () => {
       expect(serializeViewerState(CUSTOM_VIEWER_STATE, false)).toEqual(SERIALIZED_CUSTOM_VIEWER_STATE);
     });
 
-    it("deserializes partial camera settings", () => {
-      const state: Partial<ViewerState> = {
-        cameraState: {
-          position: [1.0, -1.4, 45],
-          up: [0, 1, 0],
-          fov: 43.5,
-        },
+    it("shortens long numbers in the slice and region parameters", () => {
+      // Floats should be rounded to 5 significant digits or less
+      let state: Partial<ViewerState> = {
+        region: { x: [0.4566666666, 0.8667332], y: [0.49999999, 0.8999999], z: [0.3000000001, 0.16467883] },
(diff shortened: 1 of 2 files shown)


ascibisz (approved):

website/utils/test/url_utils.test.ts
```diff
@@ -489,16 +489,15 @@ describe("Viewer state", () => {
...
+      // Floats should be rounded to 5 significant digits or less
+      let state: Partial<ViewerState> = {
+        region: { x: [0.4566666666, 0.8667332], y: [0.49999999, 0.8999999], z: [0.3000000001, 0.16467883] },
+        slice: { x: 0.41111186, y: 0.49999999, z: 0.677402 },
       };
-      const serializedState = "pos:1:-1.4:45,up:0:1:0,fov:43.5";
-      expect(deserializeViewerState({ cam: serializedState })).toEqual(state);
+      let serializedState = serializeViewerState(state, true);
+      expect(serializedState.reg).toEqual("0.45667:0.86673,0.5:0.9,0.3:0.16468");
+      expect(serializedState.slice).toEqual("0.41111,0.5,0.6774");
```
meganrm, line 500: I think I understand this, but a question: why store this many sig figs in state at all? instead of rounding them at the serialized part. If reloading the app with the url params is going to shorten them, it seems like you should store them as shorted numbers everywhere

meganrm (approved):
Looks good: I did have one question that may lead to a suggestion, but I don't think it's necessary for this PR

website/utils/test/url_utils.test.ts
ShrimpCryptid: Hmm, good question. For the regions or slice values for example, it's a 0 to 1 value in increments of `1 / # slices`, and the values we're storing are direct results of that computation. It's relative instead of by slice # in case we load in higher resolution versions of the data. 

I'd need to dig more into the code to see what happens if we reduce sig figs? I'm not totally sure how precise the math needs to be...

PR #320, opened 2024-08-31
display absolute intensities on the transfer function editor graph

toloudis:
start by just translating all the 0-255 numbers to channel min and max


frasercl:
Review time: small! Taking this out of draft because my plot range work builds on it and this seems like a reasonable unit to get reviewed first.

# What's going on here?

Originally, this viewer only supported a data format with 8-bit values &mdash; 0 to 255. This made things easy when it came to generating a histogram and plotting it in the transfer function editor component: just gather 255 bins, and plot them on an axis with a fixed range.

But now we have zarr, and most of our zarrs are 16-bit. Binning every possible value in a histogram would mean a very sparse histogram with 65535 bins! And to make matters worse, a number of our datasets only have values in a very narrow band of that range &mdash; something like 30 to 210. So currently, we *still generate histograms with 255 bins*, but values are remapped so the smallest value that the original dataset uses is 0 and the largest is 255.

Currently these histograms are still appearing on an axis from 0 to 255. This is useless to scientists because they have no information about what the actual original intensities are. So this PR makes sure the numbers they see in the transfer function editor reflect the actual intensities in the underlying data. The "plot range work" I alluded to above is a future PR that gives users a bit more control over the range of the plot, rather than sticking the ends to the min and max of the underlying data.

ShrimpCryptid:
@frasercl  code changes look good but I am seeing some bugs? 

If you load one of the Nucmorph datasets from the landing page and enable Brightfield, the range is correct but the plot will suddenly become empty:

image

Also D3 sometimes doesn't show the minimum domain label, but I assume that's a D3 issue?

image

frasercl:
@ShrimpCryptid try it now? I think a simple fix I brought back from my newer branch solves this.

As for the tick marks issue, it's solved on that branch!

ShrimpCryptid (approved):
That fixed it! Thanks, @frasercl!

image
Just a note also that the number input boxes could probably be a bit wider, since sometimes for longer range values the spinbox handles cover over the text.

frasercl (approved):

PR #321, opened 2024-09-04
feature: Improved reset to defaults behavior

### LINKED ISSUE #174: need a "reset to defaults" button
## Use Case

After users have dialed in a variety of settings, sometimes they just want to restore the default view/appearance

## Solution

A single button to accomplish this

## Alternatives

perhaps view/camera resets are separate from appearance/channel settings resets

ShrimpCryptid:
THIS PR IS WIP!

Closes #174. This change removes some of the first attempts at making a reset to defaults button, and updates it based on the new UX-agreed-upon spec. 

## Changes
- Removes old reset to defaults buttons/behavior in the advanced settings area, channels widget, and channel settings menu.
- Updates `ViewerStateContext` to include additional state:
  - `resetToSavedViewerState`: Resets to the initial view on viewer startup, determined by the input parameters passed to the viewer.
  - `resetToDefaultViewerState`: Resets to the global viewer default.
  - `setSavedChannelState`/`getSavedChannelState`: For each channel, gets or sets a saved state that can be returned to via the reset behavior.
  - `onChannelLoaded`: callback that should be invoked when a channel is loaded.
- Adds a reset button to the viewport toolbar that resets to the initial view seen on load (`resetToSavedViewerState`)
- Adds a reset button ("Clear all settings") in the advanced rendering settings menu that resets to the global viewer default.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -42,8 +42,8 @@ import {
   CACHE_MAX_SIZE,
   QUEUE_MAX_SIZE,
   QUEUE_MAX_LOW_PRIORITY_SIZE,
-  getDefaultChannelState,
-  getDefaultViewerState,
+  getEmptyChannelState,
+  getEmptyViewerState,
 } from "../../shared/constants";
 import PlayControls from "../../shared/utils/playControls";
 
@@ -111,7 +111,7 @@ const defaultProps: AppProps = {
 
   appHeight: "100vh",
   visibleControls: defaultVisibleControls,
-  viewerSettings: getDefaultViewerState(),
+  viewerSettings: getEmptyViewerState(),
   cellId: "",
   imageDownloadHref: "",
   parentImageDownloadHref: "",
@@ -132,7 +132,7 @@ const initializeOneChannelSetting = (
   index: number,
   defaultColor: ColorArray,
   viewerChannelSettings?: ViewerChannelSettings,
-  defaultChannelState = getDefaultChannelState()
+  defaultChannelState = getEmptyChannelState()
 ): ChannelState => {
   let initSettings = {} as Partial<ViewerChannelSetting>;
   if (viewerChannelSettings) {
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -612,12 +612,14 @@ const App: React.FC<AppProps> = (props) => {
     [viewerSettings.viewMode]
   );
 
-  useImageEffect((_currentImage) => {
-    // Set camera transform on initial load only
-    if (viewerSettings.cameraState) {
-      view3d.setCameraState(viewerSettings.cameraState);
-    }
-  }, []);
+  useImageEffect(
+    (_currentImage) => {
+      if (viewerSettings.cameraState) {
+        view3d.setCameraState(viewerSettings.cameraState);
+      }
+    },
+    [viewerSettings.cameraState]
+  );
 
   useImageEffect((_currentImage) => view3d.setAutoRotate(viewerSettings.autorotate), [viewerSettings.autorotate]);
 
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -192,8 +192,8 @@ const App: React.FC<AppProps> = (props) => {
     changeViewerSetting,
     changeChannelSetting,
     applyColorPresets,
-    setDefaultChannelState,
-    getDefaultChannelState,
+    setSavedChannelState,
+    getSavedChannelState,
   } = viewerState.current;
 
   const view3d = useConstructor(() => new View3d());
@@ -323,9 +323,10 @@ const App: React.FC<AppProps> = (props) => {
       }
     }
 
+    // TODO: Move to onVolumeLoaded callback in ViewerStateProvider?
     // Check for 3D vs. 2D and only save control points/channel state if the volume is the
     // right type
-    if (getDefaultChannelState(channelIndex) === undefined && viewerSettings.time === aimg.loadSpec.time) {
+    if (getSavedChannelState(channelIndex) === undefined && viewerSettings.time === aimg.loadSpec.time) {
       // TODO: Does this fail if data is chunked in a way that does not allow for subregions where z=1? Is that possible?
       const isXyAndLoadingXy = viewerSettings.viewMode === ViewMode.xy && aimg.imageInfo.subregionSize.z === 1;
       const is3dAndLoading3d = viewerSettings.viewMode === ViewMode.threeD && aimg.imageInfo.subregionSize.z > 1;
@@ -341,7 +342,7 @@ const App: React.FC<AppProps> = (props) => {
           " with control points ",
           newState.controlPoints.map((cp) => cp.x)
         );
-        setDefaultChannelState(channelIndex, newState);
+        setSavedChannelState(channelIndex, newState);
       }
     }
 
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -66,6 +66,7 @@ import {
 import { ColorArray, colorArrayToFloats } from "../../shared/utils/colorRepresentations";
 
 import "./styles.css";
+import { doesVolumeMatchViewMode } from "../../shared/utils/viewerState";
 
 const { Sider, Content } = Layout;
 
@@ -194,6 +195,7 @@ const App: React.FC<AppProps> = (props) => {
     applyColorPresets,
     setSavedChannelState,
     getSavedChannelState,
+    onChannelLoaded,
   } = viewerState.current;
 
   const view3d = useConstructor(() => new View3d());
@@ -327,10 +329,7 @@ const App: React.FC<AppProps> = (props) => {
     // Check for 3D vs. 2D and only save control points/channel state if the volume is the
     // right type
     if (getSavedChannelState(channelIndex) === undefined && viewerSettings.time === aimg.loadSpec.time) {
-      // TODO: Does this fail if data is chunked in a way that does not allow for subregions where z=1? Is that possible?
-      const isXyAndLoadingXy = viewerSettings.viewMode === ViewMode.xy && aimg.imageInfo.subregionSize.z === 1;
-      const is3dAndLoading3d = viewerSettings.viewMode === ViewMode.threeD && aimg.imageInfo.subregionSize.z > 1;
-      if (isXyAndLoadingXy || is3dAndLoading3d) {
+      if (doesVolumeMatchViewMode(viewerSettings.viewMode, aimg)) {
         const newState = {
           ...thisChannelsSettings,
           ramp: controlPointsToRamp(currentControlPoints),
@@ -345,6 +344,7 @@ const App: React.FC<AppProps> = (props) => {
         setSavedChannelState(channelIndex, newState);
       }
     }
+    onChannelLoaded(aimg, channelIndex);
 
     // save the channel's new range for remapping next time
     channelRangesRef.current[channelIndex] = [thisChannel.rawMin, thisChannel.rawMax];
@@ -425,6 +425,7 @@ const App: React.FC<AppProps> = (props) => {
   };
 
   const openImage = async (): Promise<void> => {
+    console.log("openImage");
     const { imageUrl, parentImageUrl, rawData, rawDims } = props;
     const showParentImage = viewerState.current.imageType === ImageType.fullField && parentImageUrl !== undefined;
     const path = showParentImage ? parentImageUrl : imageUrl;
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -325,13 +325,13 @@ const App: React.FC<AppProps> = (props) => {
       }
     }
 
-    // TODO: Move to onVolumeLoaded callback in ViewerStateProvider?
-    // Check for 3D vs. 2D and only save control points/channel state if the volume is the
-    // right type
-    if (getSavedChannelState(channelIndex) === undefined && viewerSettings.time === aimg.loadSpec.time) {
-      if (doesVolumeMatchViewMode(viewerSettings.viewMode, aimg)) {
+    // If we haven't initialized the control points for this channel yet and the view mode
+    // and time match, update the saved state.
+    const savedChannelState = getSavedChannelState(channelIndex);
+    if (savedChannelState && savedChannelState.controlPoints.length === 0) {
+      if (viewerSettings.time === aimg.loadSpec.time && doesVolumeMatchViewMode(viewerSettings.viewMode, aimg)) {
         const newState = {
-          ...thisChannelsSettings,
+          ...savedChannelState,
           ramp: controlPointsToRamp(currentControlPoints),
           controlPoints: currentControlPoints,
         };
@@ -344,6 +344,7 @@ const App: React.FC<AppProps> = (props) => {
         setSavedChannelState(channelIndex, newState);
       }
     }
+    // Callback to notify the viewer state that a channel has loaded.
     onChannelLoaded(aimg, channelIndex);
 
     // save the channel's new range for remapping next time
@@ -379,7 +380,10 @@ const App: React.FC<AppProps> = (props) => {
 
     const newChannelSettings = channelNames.map((channel, index) => {
       const color = (INIT_COLORS[index] ? INIT_COLORS[index].slice() : [226, 205, 179]) as ColorArray;
-      return initializeOneChannelSetting(channel, index, color, props.viewerChannelSettings);
+      const channelState = initializeOneChannelSetting(channel, index, color, props.viewerChannelSettings);
+      // Initialize saved channel state, but don't include control points to flag that the LUT has not yet been set.
+      setSavedChannelState(index, { ...channelState, controlPoints: [] });
+      return channelState;
     });
     setChannelSettings(newChannelSettings);
     return newChannelSettings;
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -193,6 +193,7 @@ const App: React.FC<AppProps> = (props) => {
     changeChannelSetting,
     applyColorPresets,
     setDefaultChannelState,
+    getDefaultChannelState,
   } = viewerState.current;
 
   const view3d = useConstructor(() => new View3d());
@@ -284,12 +285,16 @@ const App: React.FC<AppProps> = (props) => {
   const onChannelDataLoaded = (aimg: Volume, thisChannelsSettings: ChannelState, channelIndex: number): void => {
     const thisChannel = aimg.getChannel(channelIndex);
 
-    // If this is the first load of this image, auto-generate initial LUTs
+    if (aimg.loadSpec.time === 0) {
+      console.log("Volume loaded", channelIndex, aimg.imageInfo.subregionSize);
+    }
+    let currentControlPoints: ControlPoint[] = [];
     if (initialLoadRef.current || !thisChannelsSettings.controlPoints || !thisChannelsSettings.ramp) {
+      // If this is the first load of this image, auto-generate initial LUTs
       const { ramp, controlPoints } = initializeLut(aimg, channelIndex, props.viewerChannelSettings);
+      currentControlPoints = controlPoints;
       changeChannelSetting(channelIndex, "controlPoints", controlPoints);
       changeChannelSetting(channelIndex, "ramp", controlPointsToRamp(ramp));
-      setDefaultChannelState(channelIndex, { ...thisChannelsSettings, ramp: controlPointsToRamp(ramp), controlPoints });
     } else {
       // try not to update lut from here if we are in play mode
       // if (playingAxis !== null) {
@@ -304,6 +309,7 @@ const App: React.FC<AppProps> = (props) => {
         // now manually remap ramp using the channel's old range
         const controlPoints = rampToControlPoints(thisChannelsSettings.ramp);
         const newControlPoints = remapControlPointsForChannel(controlPoints, oldRange, thisChannel);
+        currentControlPoints = newControlPoints;
         changeChannelSetting(channelIndex, "ramp", controlPointsToRamp(newControlPoints));
       } else {
         // ramp was just automatically remapped - update in state
@@ -312,9 +318,33 @@ const App: React.FC<AppProps> = (props) => {
         // now manually remap control points using the channel's old range
         const { controlPoints } = thisChannelsSettings;
         const newControlPoints = remapControlPointsForChannel(controlPoints, oldRange, thisChannel);
+        currentControlPoints = newControlPoints;
         changeChannelSetting(channelIndex, "controlPoints", newControlPoints);
       }
     }
+
+    // Check for 3D vs. 2D and only save control points/channel state if the volume is the
+    // right type
+    if (getDefaultChannelState(channelIndex) === undefined && viewerSettings.time === aimg.loadSpec.time) {
+      // TODO: Does this fail if data is chunked in a way that does not allow for subregions where z=1? Is that possible?
+      const isXyAndLoadingXy = viewerSettings.viewMode === ViewMode.xy && aimg.imageInfo.subregionSize.z === 1;
+      const is3dAndLoading3d = viewerSettings.viewMode === ViewMode.threeD && aimg.imageInfo.subregionSize.z > 1;
+      if (isXyAndLoadingXy || is3dAndLoading3d) {
+        const newState = {
+          ...thisChannelsSettings,
+          ramp: controlPointsToRamp(currentControlPoints),
+          controlPoints: currentControlPoints,
+        };
+        console.log(
+          "Saving default channel state ",
+          channelIndex,
+          " with control points ",
+          newState.controlPoints.map((cp) => cp.x)
+        );
+        setDefaultChannelState(channelIndex, newState);
+      }
+    }
+
     // save the channel's new range for remapping next time
     channelRangesRef.current[channelIndex] = [thisChannel.rawMin, thisChannel.rawMax];
 
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -13,6 +13,7 @@ import {
   IVolumeLoader,
   PrefetchDirection,
   VolumeFileFormat,
+  ControlPoint,
 } from "@aics/volume-viewer";
 
 import type { AppProps, ControlVisibilityFlags, UseImageEffectType } from "./types";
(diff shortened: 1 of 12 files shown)


src/aics-image-viewer/components/ChannelsWidget.tsx
ShrimpCryptid, line 1: Deleted reset behavior

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -171,20 +184,26 @@ const ViewerStateProvider: React.FC<{ viewerSettings?: Partial<ViewerState> }> =
 
   const context = useMemo(() => {
     ref.current = {
+      ...ref.current,
```
ShrimpCryptid, line 187: Prevents updates to the ref from overriding the reset-related callbacks.

src/aics-image-viewer/components/ControlPanel/index.tsx
ShrimpCryptid, line 1: image

Adds the "Clear all settings" button in the Advanced tab.

src/aics-image-viewer/shared/constants.ts
ShrimpCryptid, line 1: `getDefaultCameraState` now takes in a view direction to fix an issue where the previous default would cause flipping or bad panning behavior after a reset.

src/aics-image-viewer/shared/constants.ts
```diff
@@ -98,19 +98,37 @@ export const PRESET_COLOR_MAP = Object.freeze([
...
-  position: [0, 0, 5],
+export const getDefaultCameraState = (viewMode: ViewMode = ViewMode.threeD): CameraState => ({
+  // Default position varies by view mode
+  position: viewModeToDefaultCameraPosition[viewMode],
   target: [0, 0, 0],
-  up: [0, 1, 0],
+  up: viewModeToDefaultCameraUp[viewMode],
   fov: 20,
   orthoScale: 0.5,
 });
```
ShrimpCryptid, line 127: `getDefaultCameraState` now takes in a view direction to fix an issue where the previous default would cause flipping or bad panning behavior after a reset.

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -66,6 +68,10 @@ export interface ChannelState {
   ramp: [number, number];
   useControlPoints: boolean;
   controlPoints: ControlPoint[];
+  /** Internal only. If set, flags that this channel should be reset to the default LUT when the
+   * next volume is loaded or when the viewer is updated.
+   */
+  needsDefaultLut?: boolean;
```
ShrimpCryptid, line 74: Had to add this to make the "Clear all settings" button work. (Alternatively I also could have overloaded the control points so `controlPoints=[]` indicates that it should be set to the defaults, but I worried that this might be too ambiguous/packing extra non-intuitive behavior. Open to feedback here.

ShrimpCryptid, line 74: Had to add this to make the "Clear all settings" button work. (Alternatively I also could have overloaded the control points so `controlPoints=[]` indicates that it should be set to the defaults, but I worried that this might be too ambiguous/packing extra non-intuitive behavior.) Open to feedback here.

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -126,7 +126,7 @@ export function initializeLut(
 }
 
 export function controlPointsToRamp(controlPoints: ControlPoint[]): [number, number] {
-  if (controlPoints.length === 1 || controlPoints.length === 3) {
+  if (controlPoints.length < 1 || controlPoints.length === 3) {
```
ShrimpCryptid, line 129: handle crash if empty control points are passed to this method

src/aics-image-viewer/shared/utils/viewerState.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export function matchesSavedSubregion(subregion: Vector3 | null, savedSubregion: Vector3 | null): boolean {
+  // Fixes a bug where the initial saved subregion has a z=1 slice, but the the actual volume has slices chunked
+  // in z=2+ slices.
+  return (
+    savedSubregion !== null &&
+    subregion !== null &&
+    subregion.x === savedSubregion.x &&
+    subregion.y === savedSubregion.y
+  );
```
ShrimpCryptid, line 59: In `App/index.tsx`, we are now applying a `requiredLoadSpec` on the initial volume load. In 2D mode, we specifically set the subregion to the Z-slice we are interested in. This causes the initial loaded slice to always have a Z-thickness of 1, and we save the subregion size as (x, y, 1).

HOWEVER, in volumes where Z-slices are chunked as 2 or more, this causes the reset behaviors to fail. There's an operation that waits until the channels for the saved subregion are loaded in, but it does a comparison between (x, y, 2) and (x, y, 1). To fix this, we skip comparing the z-dimension here.

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -29,11 +30,17 @@ const VIEWER_SETTINGS_CHANGE_HANDLERS: ViewerSettingChangeHandlers = {
...
-  renderMode: (prevSettings, renderMode) => ({
-    ...prevSettings,
-    renderMode,
-    autorotate: renderMode === RenderMode.pathTrace ? false : prevSettings.autorotate,
-  }),
+  // Also, do not allow pathtrace mode in any mode other than 3D.
+  renderMode: (prevSettings, renderMode) => {
+    if (renderMode === RenderMode.pathTrace && prevSettings.viewMode !== ViewMode.threeD) {
+      return { ...prevSettings };
+    }
```
ShrimpCryptid, line 37: This was previously enforced in the UI by disabling the path trace option, but it's possible to have conflicting URL parameters.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -304,9 +329,35 @@ const App: React.FC<AppProps> = (props) => {
...
         const newControlPoints = remapControlPointsForChannel(controlPoints, oldRange, thisChannel);
+        currentControlPoints = newControlPoints;
         changeChannelSetting(channelIndex, "controlPoints", newControlPoints);
       }
     }
+
+    // Save the first loaded channel's volume subregion as our reference for resetting channel states.
+    if (!getSavedSubregionSize()) {
+      setSavedSubregionSize(aimg.imageInfo.subregionSize.clone());
+    }
```
ShrimpCryptid, line 340: This and other API changes are explained in `ViewerStateProvider/types.ts` and `ViewerStateProvider/ResetSTateProvider.tsx`.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -292,21 +315,49 @@ const App: React.FC<AppProps> = (props) => {
...
+      const newState = {
+        ...thisChannelsSettings,
+        ramp: newRamp,
+        controlPoints: newControlPoints,
+        needsDefaultLut: false,
+      };
+      setSavedChannelState(channelIndex, newState);
+    }
+    // Callback to notify the viewer state that a channel has loaded.
+    onChannelLoaded(aimg, channelIndex);
```
ShrimpCryptid, line 359: Debating whether this should be renamed; this callback actually gives the ResetStateProvider the chance to check for and reset the channel's control points and ramp if it's currently pending reset. Something like `checkForChannelResetOnLoad`?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -418,7 +482,7 @@ const App: React.FC<AppProps> = (props) => {
       aimg = await loader.current.createVolume(loadSpec, (v, channelIndex) => {
         // NOTE: this callback runs *after* `onNewVolumeCreated` below, for every loaded channel
         // TODO is this search by name necessary or will the `channelIndex` passed to the callback always match state?
-        const thisChannelSettings = getOneChannelSetting(v.imageInfo.channelNames[channelIndex]);
+        const thisChannelSettings = viewerState.current.channelSettings[channelIndex];
```
ShrimpCryptid, line 485: Had to change channels from looking up the settings by name to using the channel's index in a few places to fix functionality in CFE; I suspect this was caused by the change to `ViewerStateProvider` and the consolidation of the channel states into an array.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -431,15 +495,23 @@ const App: React.FC<AppProps> = (props) => {
     setAllChannelsUnloaded(channelNames.length);
     channelRangesRef.current = new Array(channelNames.length).fill(undefined);
 
+    const requiredLoadspec = new LoadSpec();
+    requiredLoadspec.time = viewerState.current.time;
+
+    if (viewerSettings.viewMode === ViewMode.xy) {
+      const slice = viewerSettings.slice;
+      requiredLoadspec.subregion = new Box3(new Vector3(0, 0, slice.z), new Vector3(1, 1, slice.z));
+    }
```
ShrimpCryptid, line 504: Will we ever load a higher resolution slice for axes other than Z? @frasercl

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -431,15 +492,23 @@ const App: React.FC<AppProps> = (props) => {
     setAllChannelsUnloaded(channelNames.length);
     channelRangesRef.current = new Array(channelNames.length).fill(undefined);
 
+    const requiredLoadspec = new LoadSpec();
+    requiredLoadspec.time = viewerState.current.time;
+
+    if (viewerSettings.viewMode === ViewMode.xy) {
```
ShrimpCryptid, line 498: Needs comment

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -171,20 +185,26 @@ const ViewerStateProvider: React.FC<{ viewerSettings?: Partial<ViewerState> }> =
...
 
     // `ref` is wrapped in another object to ensure that the context updates when state does.
     // (`ref` on its own would always compare equal to itself and the context would never update.)
     return { ref };
   }, [viewerSettings, channelSettings]);
 
-  return <ViewerStateContext.Provider value={context}>{props.children}</ViewerStateContext.Provider>;
+  return (
+    <ViewerStateContext.Provider value={context}>
+      <ResetStateProvider viewerStateInputProps={props.viewerSettings}>{props.children}</ResetStateProvider>
```
toloudis, line 205: what is in props.children?

src/aics-image-viewer/shared/constants.ts
```diff
@@ -98,19 +98,37 @@ export const PRESET_COLOR_MAP = Object.freeze([
...
+  [ViewMode.xz]: [0, 2, 0],
+  [ViewMode.yz]: [2, 0, 0],
+};
+
+const viewModeToDefaultCameraUp: Record<ViewMode, [number, number, number]> = {
+  [ViewMode.threeD]: [0, 1, 0],
+  [ViewMode.xy]: [0, 1, 0],
+  [ViewMode.xz]: [0, 0, 1],
+  [ViewMode.yz]: [0, 0, 1],
+};
```
toloudis, line 113: these really look like they should be coming from volume-viewer but I suppose they are fine here.

src/aics-image-viewer/shared/constants.ts
```diff
@@ -135,19 +153,29 @@ export const getDefaultViewerState = (): ViewerState => ({
...
-  ],
-});
+/**
+ * Returns the default channel state as a new object. If an index is provided, uses the default
+ * color preset for that index.
+ * @param index Optional channel index to use for the color preset.
+ * @returns a default ChannelState object.
+ */
+export const getDefaultChannelState = (index: number = 0): ChannelState => {
+  const color = PRESET_COLORS_0[index] || PRESET_COLORS_0[0];
```
toloudis, line 163: nit: how about using PRESET_COLORS_0[index % num_of_preset_colors] ?

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
toloudis: did you want <= 1 here? what happens for length=1?

src/aics-image-viewer/shared/utils/viewerState.ts
toloudis: Maybe the above PR comment could be combined with the one in the code to augment it

package.json
```diff
@@ -59,7 +59,8 @@
   "peerDependencies": {
     "antd": "^5.16.2",
     "react": "16.x",
-    "react-dom": "16.x"
+    "react-dom": "16.x",
+    "three": "^0.144.0"
```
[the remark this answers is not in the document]
ShrimpCryptid: I need access to some data types/classes like `Vector3` and `Box3`. They're used in a few places for saving the subregion size and also setting the initial required load spec.

src/aics-image-viewer/shared/constants.ts
ShrimpCryptid: If you load something with a lot of channels, the default behavior is that it doesn't repeat the color sequence. Everything defaults to beige.

toloudis: yeah, the suggestion here is to change that, on the assumption that it would actually be an improvement :)

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
ShrimpCryptid: It's passed in from the `ViewerStateProvider`, so typically this would be the viewer component and the rest of the app. Basically anything `ViewerStateProvider` is wrapping

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -431,15 +492,23 @@ const App: React.FC<AppProps> = (props) => {
...
+      const slice = viewerSettings.slice;
+      requiredLoadspec.subregion = new Box3(new Vector3(0, 0, slice.z), new Vector3(1, 1, slice.z));
```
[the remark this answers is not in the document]
ShrimpCryptid: Slice is in a [0,1] range, yes.

src/aics-image-viewer/components/ViewerStateProvider/ResetStateProvider.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+type ResetStateProviderProps = {
+  viewerStateInputProps?: Partial<ViewerState>;
+};
+
+/**
+ * A wrapper that provides reset functionality in the ViewerStateContext, defining and overriding
+ * reset-related callbacks. Must be used inside a ViewerStateProvider.
+ */
+const ResetStateProvider: React.FC<ResetStateProviderProps> = (props) => {
```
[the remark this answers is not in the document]
ShrimpCryptid: Made this change and I think it looks pretty good now.

src/aics-image-viewer/shared/utils/viewerState.ts
ShrimpCryptid: Noted, thanks!

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -431,15 +492,23 @@ const App: React.FC<AppProps> = (props) => {
...
+    }
+
+    imageUrlRef.current = path;
+    placeImageInViewer(aimg, newChannelSettings);
```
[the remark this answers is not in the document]
ShrimpCryptid: Tested undoing this change and I couldn't recreate the bug that originally prompted it. Reverted it 👌

src/aics-image-viewer/components/App/index.tsx
ShrimpCryptid: I'm honestly not sure, I think there's only one place that's still using `getOneChannelSetting` remaining after this PR. @frasercl can you weigh in? I'm seeing a TODO immediately above this on line 420 that is asking this as well.

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
ShrimpCryptid: Yup, good catch

src/aics-image-viewer/components/ViewerStateProvider/index.tsx
ShrimpCryptid: Marking as resolved now that ResetStateProvider is a class

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -66,6 +68,11 @@ export interface ChannelState {
   ramp: [number, number];
   useControlPoints: boolean;
   controlPoints: ControlPoint[];
+  /**
+   * Internal only. If set, flags that this channel should be reset to the default LUT when the
+   * next volume is loaded or when the viewer is updated.
+   */
+  needsDefaultLut?: boolean;
```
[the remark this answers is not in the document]
ShrimpCryptid: Was able to delete along with the simplification change!

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -56,30 +58,12 @@ import StyleProvider from "../StyleProvider";
...
-function colorHexToArray(hex: string): ColorArray | null {
-  // hex is a xxxxxx string. split it into array of rgb ints
-  const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
-  if (result) {
-    return [parseInt(result[1], 16), parseInt(result[2], 16), parseInt(result[3], 16)];
-  } else {
-    return null;
-  }
-}
-
```
ShrimpCryptid, line 82: Moved to `utils/viewerState`

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -127,34 +111,6 @@ const axisToLoaderPriority: Record<AxisName | "t", PrefetchDirection> = {
...
-    colorizeAlpha: initSettings.colorizeAlpha ?? defaultChannelState.colorizeAlpha,
-    isovalue: initSettings.isovalue ?? defaultChannelState.isovalue,
-    opacity: initSettings.surfaceOpacity ?? defaultChannelState.opacity,
-    color: colorHexToArray(initSettings.color ?? "") ?? defaultColor,
-    useControlPoints: initSettings.controlPointsEnabled ?? defaultChannelState.useControlPoints,
-    controlPoints: initSettings.controlPoints ?? defaultChannelState.controlPoints,
-    ramp: defaultChannelState.ramp,
-  };
-};
-
```
ShrimpCryptid, line 157: Also moved to `utils/viewerState`

ShrimpCryptid:
*Stumbles out of the woods, covered in [code]*

I have managed to make this PR a lot simpler by changing how we're saving reset states and also making ResetStateProvider a class.

Before, we were saving the individual channel states and their control point/ramp values once they were loaded in, which had a lot of strange edge cases and required me to store a lot of additional information like the subregion size.

I've changed this so now we're only saving the `ViewerChannelSettings`, the props that are passed into App that determine the initial view. This value is hoisted up to `ResetStateProvider` and is then accessed by App indirectly through a getter function. Now, whenever a reset occurs, the App is informed whether a channel should be reset and goes through the same initialization steps it did during the initial load.

What's clever is that this also works for the "reset to default" function too! We just swap out what `ViewerChannelSetting` we give to the App and give it a bland default, but still store the initial value to reset to later. 

I'm still doing some cleanup but will mark this as ready for review once I'm done :')

src/aics-image-viewer/components/ChannelsWidget.tsx
```diff
@@ -48,8 +45,7 @@ const ChannelsWidget: React.FC<ChannelsWidgetProps> = (props: ChannelsWidgetProp
     let volChecked: number[] = [];
     let isoChecked: number[] = [];
     channelArray.forEach((channelIndex: number) => {
-      const name = channelDataChannels![channelIndex].name;
-      const channelSetting = find(channelSettings, { name });
+      const channelSetting = channelSettings[channelIndex];
```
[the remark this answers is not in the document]
ShrimpCryptid: Opened as https://github.com/allen-cell-animated/website-3d-cell-viewer/issues/325

src/aics-image-viewer/shared/constants.ts
ShrimpCryptid: Fixed! Change visible here: http://localhost:9020/viewer?url=https%3A%2F%2Fuk1s3.embassy.ebi.ac.uk%2Fidr%2Fzarr%2Fv0.4%2Fidr0076A%2F10501752.zarr

toloudis (commented):
Looks good!  Mainly at this point I suggest a couple of extra fleshed-out comments...

src/aics-image-viewer/components/ViewerStateProvider/ResetStateProvider.ts
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Helper method. Resets the viewer and all channels to the provided state. If new data needs to
+   * be loaded, handles setup so the reset will be applied to each channel as it loads in.
+   */
+  private resetToState(newState: ViewerState, newChannelStates: ChannelState[]): void {
+    const { changeViewerSetting, setChannelSettings, channelSettings, viewMode, time, slice } = this.ref.current;
+
+    // Needs reset on reload if one of the view modes is 2D while the other is 3D,
+    // if the timestamp is different, or if we're on a different z slice.
+    // TODO: Handle stopping playback? Requires playback to be part of ViewerStateContext
```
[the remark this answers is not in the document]
ShrimpCryptid: It'll jump to the original frame and continue playback from it. Though it does seem to stop playback if you switch between 3D and 2D views.

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -82,4 +84,35 @@ export type ViewerStateContextType = ViewerState & {
   changeChannelSetting: ChannelSettingUpdater;
   setChannelSettings: (settings: ChannelState[]) => void;
   applyColorPresets: (presets: ColorArray[]) => void;
+
+  // Reset-related callbacks
+  /**
+   * Sets the initial viewer channel settings that will be applied when calling
+   * `resetToSavedViewerState()`.
```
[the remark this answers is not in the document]
ShrimpCryptid: Yeah, I'll add that.

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -82,4 +84,35 @@ export type ViewerStateContextType = ViewerState & {
...
+  /**
+   * Returns the current viewer channel settings that should be used when resetting
+   * the channel transfer functions (control points and ramp).
+   */
+  getCurrentViewerChannelSettings: () => ViewerChannelSettings | undefined;
+  /** Whether the channel should be immediately reset on next render. */
+  getChannelsAwaitingReset: () => Set<number>;
+  /** Whether the channel should be reset once new data is loaded. */
+  getChannelsAwaitingResetOnLoad: () => Set<number>;
+  /** Callback used to flag when the channel has been reset. */
```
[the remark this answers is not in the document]
ShrimpCryptid: Yup! This removes it from the sets of channels awaiting reset. I'm not sure if there's a better name for this, maybe `markChannelAsReset`?

src/aics-image-viewer/components/ViewerStateProvider/ResetStateProvider.ts
```diff
@@ -0,0 +1,151 @@
...
+    this.getCurrentViewerChannelSettings = this.getCurrentViewerChannelSettings.bind(this);
+  }
+
+  // Setup Callbacks ////////////////////////////////////////////////////////////////////
+
+  public setSavedViewerState(state: Partial<ViewerState>): void {
+    this.savedViewerState = state;
+  }
+
+  public setSavedViewerChannelSettings(settings: ViewerChannelSettings | undefined): void {
```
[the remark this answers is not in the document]
ShrimpCryptid: Undefined is passed in if the viewer is used in embedded mode but there are no initial settings/props passed in. This can also happen when opening a URL with no parameters set, I believe.

src/aics-image-viewer/shared/utils/viewerState.ts
```diff
@@ -0,0 +1,80 @@
...
+/** Sets all fields of the viewer state to the values of the `newState`. */
+export function overrideViewerState(changeViewerSetting: ViewerSettingUpdater, newState: ViewerState): void {
+  for (const key of Object.keys(newState) as (keyof ViewerState)[]) {
+    changeViewerSetting(key, newState[key] as any);
+  }
+
+  // Pathtrace rendering is not allowed in 2D view mode but is in 3D mode.
+  // Since we're not guaranteed on the order of object keys, apply changes to render mode a second time
+  // in case we were previously in 2D view mode and the change was blocked, but we're now in 3D.
+  changeViewerSetting("renderMode", newState.renderMode);
```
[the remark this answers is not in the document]
ShrimpCryptid: I rewrote this so many times so I'm glad!

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -82,4 +84,35 @@ export type ViewerStateContextType = ViewerState & {
   changeChannelSetting: ChannelSettingUpdater;
   setChannelSettings: (settings: ChannelState[]) => void;
   applyColorPresets: (presets: ColorArray[]) => void;
+
+  // Reset-related callbacks
```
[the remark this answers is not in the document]
ShrimpCryptid: Done! Ended up making a type for this :)

ShrimpCryptid:
> still enforced in ui as well?
Don't know why I can't directly reply to this, but yes, it is.

src/aics-image-viewer/components/ControlPanel/index.tsx
```diff
@@ -157,4 +173,4 @@ function ControlPanel(props: ControlPanelProps): React.ReactElement {
   );
 }
 
-export default React.memo(ControlPanel);
+export default React.memo(connectToViewerState(ControlPanel, ["resetToDefaultViewerState"]));
```
frasercl, line 176: `connectToViewerState` already memoizes the passed component:


```suggestion
export default connectToViewerState(ControlPanel, ["resetToDefaultViewerState"]);
```

src/aics-image-viewer/components/ViewerStateProvider/types.ts
```diff
@@ -76,10 +78,46 @@ export type ChannelSettingUpdater = <K extends ChannelStateKey>(
...
+};
+
 export type ViewerStateContextType = ViewerState & {
   channelSettings: ChannelState[];
   changeViewerSetting: ViewerSettingUpdater;
   changeChannelSetting: ChannelSettingUpdater;
   setChannelSettings: (settings: ChannelState[]) => void;
   applyColorPresets: (presets: ColorArray[]) => void;
-};
+} & ResetState;
```
frasercl, line 123: optional nit: move up to live with the other named intersection:


```suggestion
export type ViewerStateContextType = ViewerState & ResetState & {
  channelSettings: ChannelState[];
  changeViewerSetting: ViewerSettingUpdater;
  changeChannelSetting: ChannelSettingUpdater;
  setChannelSettings: (settings: ChannelState[]) => void;
  applyColorPresets: (presets: ColorArray[]) => void;
};
```

frasercl (approved):
This looks solid, and I like `ResetStateProvider` a lot more as a class than as another component in the React tree.

PR #329, opened 2024-10-14
Fix invisible brain image

toloudis:
Default LUT had 3 control points and our code was not handling it well.
In production viewer, this url would load and ultimately reset its luts to a generic fullrange LUT which made the image invisible.
With the changes, the volumes show up as expected.
https://allen-genetic-tools.s3.amazonaws.com/tissuecyte/1298553550/ome-zarr

The code change is just trying to make a reasonable guess at a min/max from the given LUT control points.

### DIFF
--- a/src/aics-image-viewer/shared/utils/controlPointsToLut.ts
+++ b/src/aics-image-viewer/shared/utils/controlPointsToLut.ts
@@ -133,7 +133,8 @@ export function controlPointsToRamp(controlPoints: ControlPoint[]): [number, num
   } else if (controlPoints.length === 3) {
     if (
       controlPoints[0].opacity !== controlPoints[1].opacity &&
-      controlPoints[0].opacity !== controlPoints[2].opacity
+      controlPoints[0].opacity !== controlPoints[2].opacity &&
+      controlPoints[1].opacity !== controlPoints[2].opacity
     ) {
       // if all 3 are unequal, assume a ramp from first to last
       return [controlPoints[0].x, controlPoints[2].x];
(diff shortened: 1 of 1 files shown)


src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -126,10 +126,23 @@ export function initializeLut(
...
   } else if (controlPoints.length === 2) {
     return [controlPoints[0].x, controlPoints[1].x];
+  } else if (controlPoints.length === 3) {
+    if (
+      controlPoints[0].opacity !== controlPoints[1].opacity &&
+      controlPoints[0].opacity !== controlPoints[2].opacity &&
+      controlPoints[1].opacity !== controlPoints[2].opacity
+    ) {
+      // if all 3 are unequal, assume a ramp from first to last
+      return [controlPoints[0].x, controlPoints[2].x];
```
ShrimpCryptid, line 140: Could this be reduced to just a check for `controlPoints[0].opacity !== controlPoints[2].opacity`?

toloudis: if you only check c0 != c2, then c0 could still equal c1, or c1 could be equal to c2.  You still need to make some decision on which are the two best control points to choose.

interim17 (approved):
LGTM

I keep reading the title of the PR as "invisible brain damage" and think I'm reviewing a punk album.

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -126,10 +126,23 @@ export function initializeLut(
 }
 
 export function controlPointsToRamp(controlPoints: ControlPoint[]): [number, number] {
-  if (controlPoints.length <= 1 || controlPoints.length === 3) {
+  if (controlPoints.length === 1) {
```
toloudis, line 129:
```suggestion
  if (controlPoints.length <= 1) {
```

PR #330, opened 2024-10-16
refactor to allow multiple channel settings changes in one

toloudis:
Expected time to review: ~10 min

Problem
=======
Channel state changes go through a lot of react plumbing.  There are some cases where we do multiple consecutive changes and they could be consolidated.

Solution
========
Preliminary refactor to allow the channel state change API to accept batches of state setting changes.
No app behavior is changed.  There will be a followup PR to actually batch up the changes.

## Type of change

Change summary:
---------------
Change the actions to take a Record of key,value pairs instead of one single key and value.

Steps to Verify:
----------------
No changes to functionality expected.

### DIFF
--- a/src/aics-image-viewer/components/ViewerStateProvider/index.tsx
+++ b/src/aics-image-viewer/components/ViewerStateProvider/index.tsx
@@ -12,6 +12,7 @@ import type {
 import { RenderMode, ViewMode } from "../../shared/enums";
 import { ColorArray } from "../../shared/utils/colorRepresentations";
 import { getDefaultViewerState } from "../../shared/constants";
+import type { ChannelStateKey } from "./types";
 
 const isObject = <T,>(val: T): val is Extract<T, Record<string, unknown>> =>
   typeof val === "object" && val !== null && !Array.isArray(val);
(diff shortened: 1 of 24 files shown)


src/aics-image-viewer/components/ViewerStateProvider/index.tsx
```diff
@@ -76,52 +77,61 @@ const viewerSettingsReducer = <K extends keyof ViewerState>(
   }
 };
 
-/** Utility type to explicitly assert that one or more properties will *not* be defined on an object */
-type WithExplicitlyUndefined<K extends keyof any, T> = T & { [key in K]?: never };
+enum ChannelSettingActionType {
+  UniformUpdate = "UniformUpdate",
+  ArrayUpdate = "ArrayUpdate",
+  Init = "Init",
+}
```
toloudis, line 84: this file has the important change.  The actions now each have an action type, and they take `Record<K, ChannelState[K]>` which looks like `{key: value, key2:value2,...}` whereas they used to take only a single  `key:K, value:ChannelState[K]`

ShrimpCryptid (approved):
LGTM! I like how you removed the ambiguity of the action types by adding an explicit type parameter.

ascibisz (approved):

PR #333, opened 2024-10-25
fix: Channels are no longer renamed

### LINKED ISSUE #325: Don't rename channels in `ChannelState`
## Description
If a channel matches a group in `ViewerChannelSettings`, it's currently renamed and the original channel name is not stored anywhere in `ChannelState`. Consider changing the `ChannelState` to have a `displayName` and preserve the original `name` field?

## Expected Behavior

_What did you expect to happen instead?_

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
Closes #325, "don't rename channels in `ChannelState`."

*Estimated review size: tiny, 2 minutes*

Solution
========
- `initializeOneChannelSetting` no longer renames channels unless no value is provided.

## Type of change


toloudis (approved):
looks like the variable `channel` should be called `channelName`

frasercl (approved):

PR #338, opened 2024-12-04
Fix/closed loader

toloudis:
Problem
=======
Fixes https://github.com/allen-cell-animated/cell-feature-explorer/issues/220
in which a "closed" loader was being used to initiate channel loading on an image that should be completely gone.
In conjunction with https://github.com/allen-cell-animated/volume-viewer/pull/266
This also fixes some side-effect bugs with retaining correct enabled channel state, and also properly identifying when two images have the same set of channels.

Solution
========
* Remove volume safely from the viewer before updating channel state and initiating loading on the new volume.
* Because two images in the same dataset could have different channel names that are remapped with a regex to the same display name, add some storage for display names so that display name can be used as a comparison.  This comparison is used as a test for whether to retain per-channel settings when switching images.
* There was one place where the 2d single-slice mode was not being respected.
* The PR in volume-viewer also addresses a condition where the wrong channels might have been loaded in a json image.

with @frasercl and @ShrimpCryptid ... whew this was a deep dive.

## Type of change

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "author": "Megan Riel-Mehan",
   "license": "ISC",
   "dependencies": {
-    "@aics/volume-viewer": "^3.11.2",
+    "@aics/volume-viewer": "^3.12.2",
     "@ant-design/icons": "^5.2.5",
     "@fortawesome/fontawesome-svg-core": "^6.5.2",
     "@fortawesome/free-solid-svg-icons": "^6.5.2",
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -302,7 +302,9 @@ const App: React.FC<AppProps> = (props) => {
     // save the channel's new range for remapping next time
     channelRangesRef.current[channelIndex] = [thisChannel.rawMin, thisChannel.rawMax];
 
+    view3d.updateLuts(aimg);
     view3d.onVolumeData(aimg, [channelIndex]);
+
     view3d.setVolumeChannelEnabled(aimg, channelIndex, thisChannelsSettings.volumeEnabled);
     if (aimg.channelNames[channelIndex] === getCurrentViewerChannelSettings()?.maskChannelName) {
       view3d.setVolumeChannelAsMask(aimg, channelIndex);
@@ -323,17 +325,14 @@ const App: React.FC<AppProps> = (props) => {
     const grouping = makeChannelIndexGrouping(channelNames, getCurrentViewerChannelSettings());
     setChannelGroupedByType(grouping);
 
-    // we really want to match names on the remapped grouped name if possible.
-
-    // compare each channel's new displayName to the old displayNames currently in state
+    // compare each channel's new displayName to the old displayNames currently in state:
+    // same number of channels, and each channel has same displayName
     const allNamesAreEqual = channelNames.every((name, idx) => {
       const displayName = getDisplayName(name, idx, getCurrentViewerChannelSettings());
       return displayName === channelSettings[idx]?.displayName;
     });
 
     if (allNamesAreEqual) {
-      console.log("setChannelStateForNewImage: All channel names are the same; returning current channel settings");
-      //return channelSettings;
       const newChannelSettings = channelNames.map((channel, index) => {
         return { ...channelSettings[index], name: channel };
       });
@@ -341,7 +340,6 @@ const App: React.FC<AppProps> = (props) => {
       return newChannelSettings;
     }
 
-    console.log("setChannelStateForNewImage: Setting up new channel settings");
     const newChannelSettings = channelNames.map((channel, index) => {
       const color = getDefaultChannelColor(index);
       return initializeOneChannelSetting(channel, index, color, getCurrentViewerChannelSettings());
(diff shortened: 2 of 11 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -325,9 +325,19 @@ const App: React.FC<AppProps> = (props) => {
...
+      const displayName = getDisplayName(name, idx, getCurrentViewerChannelSettings());
+      return displayName === channelSettings[idx]?.displayName;
+    });
+
+    if (allNamesAreEqual) {
+      const newChannelSettings = channelNames.map((channel, index) => {
+        return { ...channelSettings[index], name: channel };
+      });
+      setChannelSettings(newChannelSettings);
+      return newChannelSettings;
```
toloudis, line 340: The above lines use a new criterion to test whether to use old or new/defaulted channel settings.  It uses the remapped display name after the getCurrentViewerChannelSettings are applied, instead of the old way of just using the raw native channel names..

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -420,12 +432,32 @@ const App: React.FC<AppProps> = (props) => {
...
+      : [];
+
+    // add mask channel to required channels?
+    const maskChannelName = getCurrentViewerChannelSettings()?.maskChannelName || "";
+    const maskChannelIndex = channelNames.indexOf(maskChannelName);
+    if (maskChannelIndex >= 0 && !requiredChannelsToLoad.includes(maskChannelIndex)) {
+      requiredChannelsToLoad.push(maskChannelIndex);
+    }
+    requiredLoadSpec.channels = requiredChannelsToLoad;
+
```
toloudis, line 460: the above set of lines 447-460 actually explicitly selects channels to load.  I'm not certain of whether this is necessary, will test one more time without it. I think it makes a difference when switching images, especially if we fail to identify the mask channel.

src/aics-image-viewer/components/ErrorAlert/index.tsx
```diff
@@ -140,6 +140,7 @@ export const useErrorAlert = (): [React.ReactNode, (error: unknown) => void] =>
   const [errorCounts, setErrorCounts] = React.useState<number[]>([]);
 
   const addError = React.useCallback((error: unknown) => {
+    console.error(error);
```
toloudis, line 143: this error output to console seemed very useful

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -356,6 +360,9 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
       if (el === null) {
         return;
       }
+      if (props.channelData.histogram.getNumBins() < 1) {
+        return;
+      }
```
toloudis, line 365: the only change in this file is this one which is basically a "null check".. at some point during debugging this, there was a condition where empty channelData was coming in here.

package.json
```diff
@@ -41,7 +41,7 @@
   "author": "Megan Riel-Mehan",
   "license": "ISC",
   "dependencies": {
-    "@aics/volume-viewer": "^3.11.2",
+    "@aics/volume-viewer": "^3.12.2",
```
toloudis, line 44: this should pick up the fixes in volume-viewer

src/aics-image-viewer/components/App/index.tsx
ShrimpCryptid: This might not apply,  but does this mean that the current ramp/control points will be preserved too? Could that cause a volume to look super blown out when switching?

ShrimpCryptid (approved):
LGTM! Left one question but these fixes look great :')

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -420,12 +432,32 @@ const App: React.FC<AppProps> = (props) => {
...
+      ? newChannelSettings.map((channel, index) => (channel.volumeEnabled ? index : -1)).filter((index) => index >= 0)
+      : [];
+
+    // add mask channel to required channels?
+    const maskChannelName = getCurrentViewerChannelSettings()?.maskChannelName || "";
+    const maskChannelIndex = channelNames.indexOf(maskChannelName);
+    if (maskChannelIndex >= 0 && !requiredChannelsToLoad.includes(maskChannelIndex)) {
+      requiredChannelsToLoad.push(maskChannelIndex);
+    }
+    requiredLoadSpec.channels = requiredChannelsToLoad;
```
frasercl, line 459: nit/optional: maybe a silly thing to optimize, but I might put this in one more `if` statement to null-check `maskChannelName` rather than giving it a fallback value `... || "";`

frasercl (approved):
Looks good! Thanks for taking point and diving deep on this!

src/aics-image-viewer/components/App/index.tsx
toloudis: I did test this and I think it makes a minor difference in channel load order but we still get correct results either way.  I am inclined to leave it in.  The only scenario I would worry about is the possibility of loading only a mask channel and no others but it would be hard to unintentionally get a data scenario where that happens.

src/aics-image-viewer/components/App/index.tsx
toloudis: I agree, I'll do that

PR #340, opened 2024-12-06
fix: Changed "In Review" flag styling

### LINKED ISSUE #339: Update tag styling on homepage so it doesn't hurt user's eyes
## Use Case
The high contrast tag background color may hurt user's eyes causing pain and general distrust in the application.
Screenshot 2024-12-05 at 3 07 06 PM

## Solution
The tag should likely avoid purple due to color contrast issues as well as looking too much like a tiny button.
Please update the following:
- font weight and color
- div element's height, border and background
Screenshot 2024-12-05 at 3 40 50 PM
Screenshot 2024-12-05 at 3 41 07 PM

Solid background styling below would be an equally reasonable option but designer slightly prefers the one with the border.
Screenshot 2024-12-05 at 3 55 19 PM

## Alternatives
You might suggest your designer take a visual design 101 refresher course, or encourage them to slow down and try focus on being a little more "detail oriented".

ShrimpCryptid:
Problem
=======
Closes #339, "Update tag styling on homepage so it doesn't hurt user's eyes".

Solution
========
- Updates styling on "In Review" flag to a border and grey text.

Screenshots (optional):
-----------------------
Before:
image

After:
image

### DIFF
--- a/website/components/LandingPage/index.tsx
+++ b/website/components/LandingPage/index.tsx
@@ -209,15 +209,15 @@ const DatasetCard = styled.li`
 const InReviewFlag = styled(FlexRowAlignCenter)`
   border-radius: 4px;
   padding: 1px 6px;
-  background-color: var(--color-statusflag-bg);
-  height: 22px;
+  border: solid 1px var(--color-statusflag-border);
+  height: 23px;
   flex-wrap: wrap;
 
   && > p {
     margin-bottom: 0;
     color: var(--color-statusflag-text);
     font-size: 11px;
-    font-weight: 600;
+    font-weight: 500;
     white-space: nowrap;
   }
 `;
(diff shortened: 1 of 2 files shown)


website/components/LandingPage/index.tsx
```diff
@@ -209,15 +209,15 @@ const DatasetCard = styled.li`
 const InReviewFlag = styled(FlexRowAlignCenter)`
   border-radius: 4px;
   padding: 1px 6px;
-  background-color: var(--color-statusflag-bg);
-  height: 22px;
+  border: solid 1px var(--color-statusflag-border);
```
frasercl, line 212: I thought the standard order for `border` was thickness then style, i.e. `border: 1px solid ...;`, but this also seems to work everywhere I try it? Optionally consider changing for consistency though, the other order is what's used elsewhere in the codebase:

```suggestion
  border: 1px solid var(--color-statusflag-border);
```

frasercl (approved):
👍

interim17 (approved):

PR #344, opened 2024-12-17
fix: Track channel load status individually

### LINKED ISSUE #342: Volumes that are initially disabled are remapped incorrectly when enabled
## Description

image

On one of the EMT volumes, enabling Brightfield will cause it to remap to use negative ramp values (despite the ramp values being set to 108, 127 in the URL).

## Reproduction

Compare:
- https://allen-cell-animated.github.io/website-3d-cell-viewer/viewer?url=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Femt_timelapse_dataset%2Fdata%2F3500005829_4_raw_converted.ome.zarr&mode=maxproject&c0=cps:0:0:1:108:0:1:127:1:1:255:1:1,rmp:108:127&c1=ven:1,col:ffffff,cps:0:0:1:30:0:1:42:1:1:255:1:1,rmp:30:65.495&c2=col:bd10e0,cps:0:0:1:13:0:1:21:1:1:255:1:1,rmp:13:21
- https://volumeviewer.allencell.org/viewer?url=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Femt_timelapse_dataset%2Fdata%2F3500005829_4_raw_converted.ome.zarr&mode=maxproject&c0=cps:0:0:1:108:0:1:127:1:1:255:1:1,rmp:108:127&c1=ven:1,col:ffffff,cps:0:0:1:30:0:1:42:1:1:255:1:1,rmp:30:65.495&c2=col:bd10e0,cps:0:0:1:13:0:1:21:1:1:255:1:1,rmp:13:21

1. Load both volumes.
2. Click to enable to Brightfield channel.
3. The channel will appear all white with negative ramp values in the current build.

## Environment

ShrimpCryptid:
Problem
=======
Fixes #342, "Volumes that are initially disabled are remapped incorrectly when enabled."

The issue seemed to be caused by the `initialLoadRef` only allowing normal initialization of channels for enabled channels were loaded for the first time. Channels that were initially disabled, but then enabled later, would have the initial default ramps or control points remapped to incorrect values.

This change fixes the issue by changing `initialLoadRef` to be a boolean array, and the loaded status of each channel is tracked individually. This way, the normal initialization can occur even when a channel is loaded later.

From testing it seems like the bug was introduced in #338, but I can't tell exactly what changes to the viewer would have broken this behavior. 

Solution
========
- Channel load status is now tracked using `channelLoadedRef`, and checks for initial load are performed against it.
  - This allows channel LUTs to be initialized for channels that are initially disabled and later reenabled.

## Type of change

Steps to Verify:
----------------
1. Open this URL: http://localhost:9020/viewer?url=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Femt_timelapse_dataset%2Fdata%2F3500005829_4_raw_converted.ome.zarr&mode=maxproject&c0=cps:0:0:1:108:0:1:127:1:1:255:1:1,rmp:108:127&c1=ven:1,col:ffffff,cps:0:0:1:30:0:1:42:1:1:255:1:1,rmp:30:65.495&c2=col:bd10e0,cps:0:0:1:13:0:1:21:1:1:255:1:1,rmp:13:21
2. Enable the Brightfield channel. It should have the expected range of 108, 127.

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -187,7 +187,7 @@ const App: React.FC<AppProps> = (props) => {
   // State for image loading/reloading
 
   // `true` when this is the initial load of an image
-  const initialLoadRef = useRef(true);
+  const hasChannelLoadedRef = useRef<boolean[]>([]);
   // `true` when image data has been requested, but no data has been received yet
   const [sendingQueryRequest, setSendingQueryRequest] = useState(false);
   // `true` when all channels of the current image are loaded
@@ -258,7 +258,7 @@ const App: React.FC<AppProps> = (props) => {
 
     // If this is the first load of this image, auto-generate initial LUTs
     if (
-      initialLoadRef.current ||
+      !hasChannelLoadedRef.current[channelIndex] ||
       !thisChannelsSettings.controlPoints ||
       !thisChannelsSettings.ramp ||
       getChannelsAwaitingResetOnLoad().has(channelIndex)
@@ -309,6 +309,7 @@ const App: React.FC<AppProps> = (props) => {
     if (aimg.channelNames[channelIndex] === getCurrentViewerChannelSettings()?.maskChannelName) {
       view3d.setVolumeChannelAsMask(aimg, channelIndex);
     }
+    hasChannelLoadedRef.current[channelIndex] = true;
 
     // when any channel data has arrived:
     setSendingQueryRequest(false);
@@ -316,7 +317,6 @@ const App: React.FC<AppProps> = (props) => {
     if (aimg.isLoaded()) {
       view3d.updateActiveChannels(aimg);
       setImageLoaded(true);
-      initialLoadRef.current = false;
       playControls.onImageLoaded();
     }
   };
@@ -400,7 +400,7 @@ const App: React.FC<AppProps> = (props) => {
 
     setSendingQueryRequest(true);
     setImageLoaded(false);
-    initialLoadRef.current = true;
+    hasChannelLoadedRef.current = [];
 
     const loadSpec = new LoadSpec();
     loadSpec.time = viewerState.current.time;
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -186,7 +186,7 @@ const App: React.FC<AppProps> = (props) => {
 
   // State for image loading/reloading
 
-  // `true` when this is the initial load of an image
+  /** `true` when a channel's data has been loaded for the current image. */
   const hasChannelLoadedRef = useRef<boolean[]>([]);
   // `true` when image data has been requested, but no data has been received yet
   const [sendingQueryRequest, setSendingQueryRequest] = useState(false);


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -187,7 +187,7 @@ const App: React.FC<AppProps> = (props) => {
   // State for image loading/reloading
 
   // `true` when this is the initial load of an image
-  const initialLoadRef = useRef(true);
+  const hasChannelLoadedRef = useRef<boolean[]>([]);
```
toloudis, line 190: Maybe the comment above this should say "per-channel"

toloudis (approved):
Very clean! Ship it!

frasercl (approved):
Thinking out loud here: I wonder if, in the spirit of not duplicating state, `hasChannelLoadedRef` should be merged with `channelVersions` below now that it's also responsible for tracking per-channel load states? e.g. an initial load would be version `-1`? I'm okay letting this fix in as-is but might return to this in the future

ShrimpCryptid:
> Thinking out loud here: I wonder if, in the spirit of not duplicating state, `hasChannelLoadedRef` should be merged with `channelVersions` below now that it's also responsible for tracking per-channel load states? e.g. an initial load would be version `-1`? I'm okay letting this fix in as-is but might return to this in the future

I tested this and it seems to work great! I think it's initialized to version `0` by default which I can test against. Are there any concerns with not using a `ref` value?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -232,7 +232,7 @@ const App: React.FC<AppProps> = (props) => {
   };
 
   const setAllChannelsUnloaded = (numberOfChannels: number): void => {
-    setChannelVersions(new Array(numberOfChannels).fill(0));
+    setChannelVersions(new Array(numberOfChannels).fill(INITIAL_CHANNEL_VERSION));
   };
 
   const setOneChannelLoaded = (index: number): void => {
```
ShrimpCryptid, line 238: Channel versions are incremented once they're loaded once!

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: Looks like channel transfer functions are now getting recomputed rather than remapped on initial load. This is why I suggested `-1` for `INITIAL_CHANNEL_VERSION` initially. Though it does add some additional complexity: `setOneChannelLoaded` has to go from simply incrementing channel versions to something like `Math.max(version + 1, 1)`, and we need two different versions of `setAllChannelsUnloaded` for when channels are unloaded due to an image switch vs. due to loading new data from the same image, where the second does something like `setChannelVersions(getChannelVersions.map((version) => Math.min(version, 0)));`.

Not sure whether the additional complexity is worth the unification; I'm leaning towards yes, but I'm on the fence. If so, the representation of `channelVersions` should definitely get a few lines of documentation. (And honestly, it should already have documentation - the current representation just confused me, and I wrote it. The rationale for representing loaded state as a number rather than a boolean is described all the way back in #156.)

ShrimpCryptid: Moved this to a new branch because of some complexity that was worth checking a little more thoroughly! It's now on `refactor/remove-channel-loaded-ref`.

toloudis: I'm all for any change that both makes things work correctly and improves the readability, maintainability, and simplicity of the codebase.  Do you really think we can get rid of the channelversions?

frasercl (approved):

PR #357, opened 2025-01-23
doc: Added URL parameter documentation

### LINKED ISSUE #358: Document URL parameters
## Use Case

Users who want to view volumes with certain settings in our standalone viewer do not currently have documentation on what our URL parameters do.

## Solution

Create a Markdown document in our repo that explains the URL parameter spec and what the values control.

## Alternatives

ShrimpCryptid:
Problem
=======
Closes #358, "Document URL parameters." This change adds a markdown document describing the URL query parameters accepted by Vol-E.

*Estimated review size: medium, 20-30 minutes*

Solution
========
-  Adds URL query parameter documentation as a markdown document.

Steps to Verify:
----------------
1. Review the formatted spec here: https://github.com/allen-cell-animated/website-3d-cell-viewer/blob/doc/urls/documentation/URL_SPEC.md


documentation/URL_SPEC.md
ShrimpCryptid, line 1: Formatted version here: https://github.com/allen-cell-animated/website-3d-cell-viewer/blob/doc/urls/documentation/URL_SPEC.md

I'll leave some specific questions in the diff here as well.

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,157 @@
...
+as the base address. If you are running Vol-E locally, you can substitute this
+for a `localhost` address.
+
+## Common Examples
+
+| Query Parameters                            | Description                                                                                                                               |
+| ------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `?url={url}&c0=ven:1&c2=ven:1,clz:1&view=Z` | Enable the first and third channel volumes and apply colorizing to the third. View in XY / Z-slice mode.                                  |
+| `?url={url}&c0=ven:1,rmp:15:180`            | Enables the volume of the first channel and ramps opacity for volume intensity values from 0% at histogram bin `15` to 100% at bin `180`. |
+| `?url={url}&reg=0:1,0:1,0.25:0.75,t=40`     | Clips along the Z-axis to the middle 50% of the volume, and sets the time to frame 40.                                                    |
```
ShrimpCryptid, line 21: Any other common examples I should add? I use this mostly as a cookbook for users who don't want to read the whole spec.

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,157 @@
...
+
+For example, the above control points are at bin indices `0`, `53`, `136`, and
+`255`. Depending on the range of the channel's intensity values, these may
+represent different raw intensity values.
+
+| Min raw intensity | Max raw intensity | Bin `0` intensity | Bin `53` intensity | Bin `136` intensity | Bin `255` intensity |
+| ----------------- | ----------------- | ----------------- | ------------------ | ------------------- | ------------------- |
+| 0                 | 255               | 0                 | 53                 | 136                 | 255                 |
+| 50                | 200               | 50                | 103                | 166                 | 200                 |
+| 0                 | 1                 | 0                 | 0.207              | 0.533               | 1                   |
```
ShrimpCryptid, line 133: Is this an accurate description of how the binning works?

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,157 @@
...
+| -------- | ----------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- | -------------------------------------- | ----------------------------------- |
+| `ven`    | Volume enabled.                                                                           | `1` (enabled) or `0` (disabled)                                                                | `0` (`1` for the first three channels) | `?c0=ven:1`                         |
+| `iso`    | Isosurface enabled.                                                                       | `1` (enabled) or `0` (disabled)                                                                | `0`                                    | `?c0=iso:1`                         |
+| `isv`    | Isosurface value, the intensity value at which to generate the isosurface.                | Number in the range `[0, 255]`                                                                 | `128`                                  | `?c0=isv:195`                       |
```
ShrimpCryptid, line 103: From a quick look I *think* this is using raw intensity values.

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+
+For example, the above control points are at bin indices `0`, `53`, `136`, and
+`255`. Depending on the range of the channel's values, these may
+represent different raw intensity values.
+
+| Min raw intensity | Max raw intensity | Bin `0` intensity | Bin `53` intensity | Bin `136` intensity | Bin `255` intensity |
+| ----------------- | ----------------- | ----------------- | ------------------ | ------------------- | ------------------- |
+| 0                 | 255               | 0                 | 53                 | 136                 | 255                 |
+| 50                | 200               | 50                | 103                | 166                 | 200                 |
+| 0                 | 1                 | 0                 | 0.207              | 0.533               | 1                   |
```
ShrimpCryptid, line 135: @frasercl @toloudis is this an accurate description of how the binning works?

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+| Query Parameters                | Description                               | Example                                                                                      |
+| ------------------------------- | ----------------------------------------- | -------------------------------------------------------------------------------------------- |
+| `?url={url}`                    | Load a volume from the specified URL.     | `?url=https://example.com/data/example1.ome.zarr`                                            |
+| `?url={url1},{url2},{url3},...` | Load multiple volumes from multiple URLs. | `?url=https://example.com/data/example1.ome.zarr,https://example.com/data/example2.ome.zarr` |
+
+## View settings
+
+General global settings for the viewer.
+
+| Query parameter | Description                                                                                                             | Expected Type                                                       | Default       | Example                         |
```
frasercl, line 45: "Expected Values" might be clearer here - I'd expect "type" to read to people like "string" or "number," thinking of types as subsets beneath those "classical" types (as TypeScript lets you do with e.g. literal types) is relatively niche.

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+represented by a `min:max` pair. Values between the `min` and `max` ramp
+linearly from an opacity of `0` to `1`. The `min` and `max` can match any of the
+following:
+
+- Plain numbers are treated as bin indices, in a `[0, 255]` range.
+- `p{n}` represents a percentile, where `n` is a percentile in the `[0, 100]` range.
+- `m{n}` represents the volume's median intensity multiplied by `n / 100`.
+- `autoij` in either the min or max fields will use the ["auto" algorithm from
+  ImageJ](https://github.com/imagej/ImageJ/blob/7746fcb0f5744a7a7758244c5dcd2193459e6e0e/ij/plugin/frame/ContrastAdjuster.java#L816)
+  to select the min and max.
```
frasercl, line 149: Unsure if we can claim to _exactly_ implement ImageJ's algorithm, though @toloudis would know more

frasercl (approved):
Clear, detailed, readable!

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+| `?url={url}&reg=0:1,0:1,0.25:0.75,t=40`     | Clips along the Z-axis to the middle 50% of the volume, and sets the time to frame 40.                                                    |
+
+## Data source (`url`)
+
+The `url` parameter specifies the HTTPS URL of one or more volume to be loaded.
+Supported formats include OME-Zarr and OME-TIFF files.
+
+Multiple volumes can be loaded by including commas between each URL. This
+requires all volumes to have **some resolution/scale where the dimensions
+match.** The viewer will throw an error if there is no match possible.
```
toloudis, line 30: You need to say that the volumes will show up combined as channels

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+
+URLs containing special characters (`?`, `#`, `&`, or `,`) must be first encoded
+using
+[`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).
+
+| Query Parameters                | Description                               | Example                                                                                      |
+| ------------------------------- | ----------------------------------------- | -------------------------------------------------------------------------------------------- |
+| `?url={url}`                    | Load a volume from the specified URL.     | `?url=https://example.com/data/example1.ome.zarr`                                            |
+| `?url={url1},{url2},{url3},...` | Load multiple volumes from multiple URLs. | `?url=https://example.com/data/example1.ome.zarr,https://example.com/data/example2.ome.zarr` |
```
toloudis, line 39: maybe add here that the volume channels will be appended together in the viewer, too

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+| `view`          | Initial view axis.                                                                                                      | `X`, `Y`, `Z`, or `3D`                                              | `3D`          | `?view=Z`                       |
+| `mode`          | Rendering mode to use.                                                                                                  | `volumetric`, `maxproject`, or `pathtrace`                          | `volumetric`  | `?mode=maxproject`              |
+| `mask`          | The opacity of masked areas of the volume. Used to fade areas outside a segmentation, must be used with mask channel.   | Number in range `[0, 100]`                                          | `50`          | `?mask=75`                      |
+| `image`         | Which of two images to display, for JSON-based volume data that provides two volumes for the cell and wider colony FOV. | `cell` or `fov`                                                     | `cell`        | `?image=fov`                    |
```
toloudis, line 50: do we even use this?  I think `image` could probably be removed from this doc

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+| `rot`           | Whether to autorotate the view, which will spin it slowly on the vertical axis. Only available in 3D view.              | `1` (enabled) or `0` (disabled)                                     | `0`           | `?rot=1`                        |
+| `bright`        | The brightness of the image.                                                                                            | Number in the range `[0, 100]`                                      | `70`          | `?bright=80`                    |
+| `dens`          | Density modifier for the volume`. Higher densities make the volume appear more opaque.                                  | Number in the range `[0, 100]`                                      | `50`          | `?dens=60`                      |
+| `lvl`           | Low, medium, and high levels for image intensity adjustment                                                             | Three numbers values separated by commas                            | `35,140,255`  | `?lvl=0,128,255`                |
```
toloudis, line 58: expects 0-255 range

documentation/URL_SPEC.md
```diff
@@ -0,0 +1,159 @@
...
+| `cpe`    | Use and show the control points instead of the ramp on load.                              | `1` (enabled) or `0` (disabled)                                                                | `0`                                    | `?c0=cpe:1`                         |
+| `cps`    | Control points for the transfer function. If provided, overrides the `lut` field.         | List of `bin_index:opacity:color` triplets, separated by a colon ([_see 'Binning'_](#binning)) | `0:0:ffffff:255:1:ffffff`              | `?c0=cps:0:0:ff0000:150:0.5:ffff00` |
+| `rmp`    | Raw ramp values. If provided, overrides the `lut` field when calculating the ramp values. | Two float bin indices, formatted as `min:max` ([_see 'Binning'_](#binning))                    | `0:255`                                | `?c0=rmp:0:255`                     |
+| `lut`    | Lookup table (LUT) to map from volume intensity values to opacity.                        | [_See 'Lookup Table'_](#lookup-table-lut)                                                      | `0:255`                                | `?c0=lut:0:255`                     |
```
toloudis, line 113: do lut and rmp map the same values - a min and a max?  I think the only difference is that lut accepts the shorthand / prefixed values?

documentation/URL_SPEC.md
toloudis: You might want to say something to remind users that the bins represent a range of values (like any binned histogram).  So if the range of intensities is greater than 256, then the bins span over multiple intensities.  Pixels in an intensity range go into  the bins

documentation/URL_SPEC.md
ShrimpCryptid: Yes! There is some different behavior though, since setting `lut` will set both the control points and the ramp, unless one or both are overridden.

ShrimpCryptid: When we send share links from VV it's always with `rmp` and `cps` and we don't include `lut`.

documentation/URL_SPEC.md
ShrimpCryptid: Changed the last row to have a domain of greater than 256 values, and updated the bins to be ranges.

documentation/URL_SPEC.md
toloudis: Maybe we should remove lut then!

ShrimpCryptid: I think it may still be a good idea to keep it? This way a user can still say, "load any volume with autoij"

Issue #275, opened 2024-06-24
Update landing page for NucMorph publication


ShrimpCryptid:
## Use Case

Update the landing page for the NucMorph publication to match TFE + include volume data links.

## Solution

- Copy over the language from TFE about the landing page (once finalized by Julie)
- Copy over Lyndsay's styling from TFE.
- Add links to the three zarr datasets once they're uploaded (needs information from NucMorph teams)

## Alternatives

_Please describe any alternatives you've considered, even if you've dismissed them_

PR #277, opened 2024-06-26
feature: Update landing page

ShrimpCryptid:
Closes #275, updating the landing page for the NucMorph publication.

*Estimated review size: medium,, 15-20 minutes*

## Changes
- Fixes a bug where defining viewer settings in the landing page content would not cause it to update.
- Adjusts some styling to reflect changes made in TFE, based on UX feedback.
- Adds links to Quilt ZARR data for the Nucmorph colony datasets.

image

Loaded dataset:
image

PR #362, opened 2025-02-13
Multi-scene collections

### LINKED ISSUE #351: Multiscene UI
## Use Case

Scientists want to easily switch between scenes of multi-scene data sets.

## Solution

Add a slider to the clipping panel to control the scene in view. Only show the slider if the image has multiple scenes.

frasercl:
Review time: medium (~20-30min)

# Problem

Scientists want a way to view multiple images collected along a "scenes" dimension. This PR gets us to basic support for that.

Resolves #351: adds a new "Scene" slider to the clipping panel.

Additionally, resolves allen-cell-animated/vole-core#275. Most of the work on this ticket was finished in allen-cell-animated/vole-core#279 &mdash; that PR described the remaining work as creating a new data structure or framework to gather multiple data sources into a single multi-scene collection. This PR includes that: the `SceneStore` class, described below.

# Solution

- Add a new `SceneStore` class, which manages creating and swapping out loaders for different scenes, and which replaces the single loader that the `App` component used to hold onto.
- Add new semantics for `App` props to support describing multi-scene collections, and new semantics to URL parameters to parse out those props
- Add a scene slider to the clipping panel, which appears when viewing a multi-scene collection. Because the clipping panel is too short by default to contain an X, Y, Z, time, and scene slider all at once, the panel now grows taller when all these sliders are present.

### Remaining work

This implementation deliberately performs minimal validation. For example, it will fall over when successive scenes have a different number of channels. We consider this acceptable while we're demoing this feature to scientists, since this feature isn't exposed to users in any prominent way. This remaining work is captured in allen-cell-animated/vole-core#277.

# Steps to verify

- `npm install`, `npm start`
- Add a `?url=` parameter to the URL, and join multiple URLs with `+`. For example: http://localhost:9020/viewer?url=https%3A%2F%2Fanimatedcell-test-data.s3.us-west-2.amazonaws.com%2Fvariance%2F1.zarr+https%3A%2F%2Fanimatedcell-test-data.s3.us-west-2.amazonaws.com%2Fvariance%2F2.zarr+https%3A%2F%2Fanimatedcell-test-data.s3.us-west-2.amazonaws.com%2Fvariance%2F3.zarr

# Screenshot

image


src/aics-image-viewer/components/App/types.ts
```diff
@@ -31,14 +31,23 @@ export type ControlVisibilityFlags = { [K in ControlNames]: boolean };
...
-  parentImageUrl?: string | string[];
+
+  // The inner level of array(s), if present, groups multiple sources into a single volume with all sources' channels.
+  // If there is an outer array level, it groups multiple volumes into a single multi-scene collection.
+  // To clarify:
+  // - A bare string is a single volume scene with a single source.
+  // - An array of only strings is interpreted as a single volume with multiple sources, not a multi-scene collection.
+  // - An array of strings *or* string arrays is a multi-scene collection, and all strings within the top-level array
+  //   are treated as if they were string arrays with one element (i.e. volumes with one source).
+  imageUrl: string | (string | string[])[];
```
toloudis, line 49: from the comment above, it's not clear how you can distinguish between "array of only strings" and "array of strings or string arrays"

website/utils/url_utils.ts
```diff
@@ -955,8 +960,12 @@ export async function parseViewerUrlParams(urlSearchParams: URLSearchParams): Pr
 
   // Parse data sources (URL or dataset/id pair)
   if (params.url) {
-    const imageUrls = tryDecodeURLList(params.url) ?? decodeURL(params.url);
-    const firstUrl = Array.isArray(imageUrls) ? imageUrls[0] : imageUrls;
+    const sceneUrls = tryDecodeURLList(params.url, " ") ?? [params.url];
+    const sourceUrls = sceneUrls.map((scene) => tryDecodeURLList(scene) ?? [decodeURL(scene)]);
```
toloudis, line 964: is this supposed to give you every source url of every scene? I guess I don't know what tryDecodeUrlList can return .  I assume decodeURL returns a single string?

website/utils/url_utils.ts
```diff
@@ -955,8 +960,12 @@ export async function parseViewerUrlParams(urlSearchParams: URLSearchParams): Pr
...
+
+    const firstScene = sourceUrls[0];
+    const firstUrl = Array.isArray(firstScene) ? firstScene[0] : firstScene;
+    const imageUrls = sourceUrls.length === 1 ? (sourceUrls[0].length === 1 ? firstScene[0] : firstScene) : sourceUrls;
```
toloudis, line 968: it has become too hard to parse these 5 lines. I think types on the variable declarations would help to understand.  And maybe rename imageUrls and firstUrl to be imageUrlsOfFirstScene and firstUrlOfFirstScene?  if that's even accurate?

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -168,9 +186,9 @@ const App: React.FC<AppProps> = (props) => {
   // install loadContext into view3d
   view3d.loaderContext = loadContext;
 
-  const loader = useRef<IVolumeLoader>();
+  const loader = useRef<SceneStore>();
   const [image, setImage] = useState<Volume | null>(null);
-  const imageUrlRef = useRef<string | string[]>("");
+  const imageUrlRef = useRef<string | (string | string[])[]>("");
```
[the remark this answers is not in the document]
toloudis: I even wonder if there's a way to formalize it with a simple construct like 
```
type Scene {
  sources: string[];
}
```
 instead of carrying these weird arrays around.
 Then the url parsing can just fill an array of Scenes.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -108,17 +112,31 @@ const axisToLoaderPriority: Record<AxisName | "t", PrefetchDirection> = {
...
-  const CLIPPING_PANEL_HEIGHT = 150;
+/** `true` if `p` is not an array that contains another array */
+const notDoublyNested = <T,>(p: T | (T | T[])[]): p is T | T[] => !Array.isArray(p) || !p.some(Array.isArray);
+
+const setIndicatorPositions = (
+  view3d: View3d,
+  panelOpen: boolean,
+  hasTime: boolean,
+  hasScenes: boolean,
+  mode3d: boolean
```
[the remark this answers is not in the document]
ShrimpCryptid: Seconding `isMode3d`

src/aics-image-viewer/components/App/types.ts
```diff
@@ -31,14 +31,23 @@ export type ControlVisibilityFlags = { [K in ControlNames]: boolean };
...
-  imageUrl: string | string[];
-  parentImageUrl?: string | string[];
+
+  // The inner level of array(s), if present, groups multiple sources into a single volume with all sources' channels.
+  // If there is an outer array level, it groups multiple volumes into a single multi-scene collection.
+  // To clarify:
+  // - A bare string is a single volume scene with a single source.
+  // - An array of only strings is interpreted as a single volume with multiple sources, not a multi-scene collection.
+  // - An array of strings *or* string arrays is a multi-scene collection, and all strings within the top-level array
+  //   are treated as if they were string arrays with one element (i.e. volumes with one source).
```
ShrimpCryptid, line 48:
```suggestion
  // - An array where one or more elements are string *arrays* is a multi-scene collection.
  //   - String elements are treated as if they were string arrays with one element (i.e. volumes with one source).
  //   - String arrays are single volumes with multiple sources.
```
Consider breaking this down into sub-bullets?

src/aics-image-viewer/components/App/types.ts
ShrimpCryptid, line 49:
```suggestion
  imageUrl: string | string[] | (string | string[])[];
```
I know this is redundant but it might make it easier to parse this in combination with your comment above

website/utils/datatype_utils.ts
```diff
@@ -1,5 +1,8 @@
 import { isEqual } from "lodash";
 
+/** `true` if `p` is not an array that contains another array */
+export const notDoublyNested = <T>(p: T | (T | T[])[]): p is T | T[] => !Array.isArray(p) || !p.some(Array.isArray);
+
```
ShrimpCryptid, line 5: Nit: Convention is usually to make a function name a positive assertion since it makes it easier to reason about.

e.g. `isDoubleNested` or `isStringOrFlatArray` or `isSingleLevelArray` etc.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -108,17 +112,31 @@ const axisToLoaderPriority: Record<AxisName | "t", PrefetchDirection> = {
   x: PrefetchDirection.X_PLUS,
 };
 
-const setIndicatorPositions = (view3d: View3d, panelOpen: boolean, hasTime: boolean): void => {
-  const CLIPPING_PANEL_HEIGHT = 150;
+/** `true` if `p` is not an array that contains another array */
+const notDoublyNested = <T,>(p: T | (T | T[])[]): p is T | T[] => !Array.isArray(p) || !p.some(Array.isArray);
```
ShrimpCryptid, line 116: Duplicated with `datatype_utils`?

website/utils/datatype_utils.ts
ShrimpCryptid: `isMultiSceneArray`?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -69,28 +72,34 @@ const ViewerWrapper: React.FC<ViewerWrapperProps> = (props) => {
     return noImageText || spinner;
   };
 
-  const { appHeight, changeViewerSetting, visibleControls, numSlices, numTimesteps, viewMode, region, slice, time } =
+  const { appHeight, changeViewerSetting, visibleControls, numTimesteps, numScenes, viewMode, region, slice, time } =
     props;
+  const clippingPanelTall = numTimesteps > 1 && numScenes > 1 && viewMode === ViewMode.threeD;
```
ShrimpCryptid, line 77: This check happens in two places... For DRY, should it just be a single flag that's passed in as a prop?

src/aics-image-viewer/components/App/index.tsx
frasercl: removed by most recent commit

website/utils/datatype_utils.ts
frasercl: removed

src/aics-image-viewer/components/App/index.tsx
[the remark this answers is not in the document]
frasercl: did roughly the second thing above

src/aics-image-viewer/components/App/types.ts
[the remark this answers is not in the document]
frasercl: removed comment

website/utils/url_utils.ts
frasercl: My little refactor didn't make this much better, but I did add some docs and a type signature. Let me know if there's more I can do?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: My thought here was that, since this property is derivable from other passed properties, it was better not to pass it in explicitly and risk an "invalid" value. Though this value has less of a strong notion of "invalid" than, say, a `string` along with an `isAllCaps: boolean` value. So maybe it would be better?

website/components/LandingPage/index.tsx
```diff
@@ -253,16 +254,11 @@ export default function LandingPage(): ReactElement {
...
-        state: appProps,
-      });
-    } else {
-      navigation(`/viewer?url=${encodeURIComponent(url)}`, {
-        state: appProps,
-      });
-    }
+    const urls = (appProps.imageUrl as MultisceneUrls).scenes ?? [appProps.imageUrl];
+    const sceneUrlsUnencoded = urls.map((scene) => (Array.isArray(scene) ? scene.join(",") : scene));
+    navigation(`/viewer?url=${encodeURIComponent(sceneUrlsUnencoded.join(" "))}`, {
```
ShrimpCryptid, line 259: If this code is copied with `AppWrapper` maybe it's worth putting in a function?

ShrimpCryptid (approved):
Left some final questions but I like the new way of representing multiscenes.

src/aics-image-viewer/components/App/types.ts
```diff
@@ -29,16 +29,26 @@ type ControlNames =
 /** Show/hide different elements of the UI */
 export type ControlVisibilityFlags = { [K in ControlNames]: boolean };
 
+export type MultisceneUrls = { scenes: (string | string[])[] };
```
toloudis, line 32: why not
```
{ 
  scenes: 
  {
    sources :string[]
  } []
}
```
 ?

toloudis: or to be even more explicit and readable, something like:
export type Scene = { sources: string[] };
export type MultiSceneUrls = Scene[];

I'm still not sure why you need to have `string | string[]`

src/aics-image-viewer/components/App/types.ts
```diff
@@ -29,16 +29,26 @@ type ControlNames =
...
-  parentImageUrl?: string | string[];
+
+  /**
+   * URL(s) from which to fetch the image. You can pass a `string` to load from a single data source, or get fancier:
+   * - Pass an array of strings to assemble a single volume with all sources' channels, in order.
+   * - Pass an object with a key `scenes: (string | string[])[]` to load multiple volumes as a *multi-scene collection*.
+   *   Each string or string array within the `scenes` array is treated as a single volume with one or more sources.
+   */
+  imageUrl: string | string[] | MultisceneUrls;
+  parentImageUrl?: string | string[] | MultisceneUrls;
```
toloudis, line 51: is it impossible to just have these accept MultisceneUrls?

src/aics-image-viewer/components/App/types.ts
frasercl: My thought here was mainly backwards compatibility... though I suppose the multi-source images we have out in the wild are expressed as URLs and so we can just choose to parse them into a new props type?

src/aics-image-viewer/components/App/types.ts
frasercl: Same thing - would be a breaking change that we'd have to propagate to NBVV, CFE, etc. Or are you asking to include `string | string[]` in the `MultisceneUrls` type?

src/aics-image-viewer/components/App/types.ts
toloudis: that was my thought, use the url parsing code to transform them into the props type that we want.  Maybe it's ok that the props are still general to string or array of string or multiscene struct, but as early as possible it should all get converted into a multiscene struct for the rest of the app to use.

src/aics-image-viewer/components/App/types.ts
toloudis: I understand that here it's a compatibility problem.  PArt of me wants to say we can break older versions and accept that nbvv and cfe need updates.  Or, since we are still sort of prototyping, I would really even consider a new named prop called multisceneImageUrl : MultisceneUrls but I'm not psyched about these unions of types overloading the same variable names.

frasercl:
@toloudis I've simplified the props type for image URLs a bit, from `string | string[] | MultisceneUrls` to `string | MultisceneUrls`. Note that this change required changing the content of the landing page too. The further changes I might make would be:

- Removing `string` and just having `MultisceneUrls`. I think it's reasonable to have an easy fallback for just looking at one image from one source, it's simple to parse it into a `MultisceneUrls` record for consistency, and removing it will definitely require changes in CFE, nbvv, etc.
- Adding a new layer of record in `MultisceneUrls` &mdash; i.e. turning `{ scenes: (string | string[])[] }` into `{ scenes: { sources: string[] }[] }`. I think this is unnecessarily complicated, especially given that we don't really treat bare URLs differently from URL arrays at this level &mdash; we just pass either one down to a loader in vole-core, and it does the work of assembling a multi-source volume from there.

toloudis (approved):
LGTM

PR #369, opened 2025-04-08
Usability: open control panel on tab change; make tf editor sliders more draggable

frasercl:
Review time: tiny <5min

# Problem

Two minor, easy usability wins:
- Currently, when the control panel is collapsed, the buttons to change the panel to a different tab appear to do nothing:
image
- Currently, in the "basic mode" transfer function editor, only the purple slider handles poking off the top and bottom of the plot are interactable; the purple lines are not:
image

# Solution

- Now, clicking a tab while the control panel is collapsed also un-collapses the control panel
- The lines on the transfer function editor sliders are dashed and quite slim, so rather than make them clickable there's now a second invisible line on top of them with a heavier, solid stroke. Clicking on it has the same behavior as the current slider handles.
- Per @lynwilhelm's feedback, both the slider handles and the lines have a different cursor (this thing: <->)

### Steps to verify:

- `npm start`
- Try it!

### DIFF
--- a/src/aics-image-viewer/components/ControlPanel/index.tsx
+++ b/src/aics-image-viewer/components/ControlPanel/index.tsx
@@ -1,19 +1,18 @@
+import { Button, Collapse, CollapseProps, Dropdown, Flex, MenuProps, Tooltip } from "antd";
+import { MenuInfo } from "rc-menu/lib/interface";
 import React from "react";
 
-import { Button, Dropdown, Tooltip, MenuProps, Collapse, CollapseProps, Flex } from "antd";
-import { MenuInfo } from "rc-menu/lib/interface";
+import { PRESET_COLOR_MAP } from "../../shared/constants";
+import { MetadataRecord } from "../../shared/types";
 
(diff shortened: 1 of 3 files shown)


src/aics-image-viewer/components/ControlPanel/index.tsx
```diff
@@ -1,20 +1,19 @@
...
-
-import { PRESET_COLOR_MAP } from "../../shared/constants";
-
-import "./styles.css";
 import ViewerIcon from "../shared/ViewerIcon";
-import { MetadataRecord } from "../../shared/types";
 import { connectToViewerState } from "../ViewerStateProvider";
 
+import "./styles.css";
+
```
frasercl, line 16: This is all just automatic import reordering!

ShrimpCryptid (approved):
Works great! A part of me wishes that clicking on the panel buttons again would collapse the panel...

BTW I have PR previews set up on vole-app now, so you can test in the browser here: https://allen-cell-animated.github.io/vole-app/pr-preview/pr-369/

interim17 (approved):

PR #370, opened 2025-04-08
Adjustable range in transfer function editor

### LINKED ISSUE #373: Adjustable range in transfer function editor
## Use Case

Scientists want channel histograms in the transfer function editor to be displayed in a consistent, configurable range. The current behavior, where the ends of the plot are fixed to the min and max of the data per-frame, is unacceptable.

## Solution

The option should be available in the UI to fix the left side of the plot at absolute 0 and to configure the upper range of the plot.

frasercl:
Review size: medium

# Problem

Closes #373: scientists want channel histograms in the transfer function editor to be displayed in a consistent, configurable range. The current behavior, where the ends of the plot are fixed to the min and max of the data per-frame, is unacceptable.

# Solution

- Add new UI elements which afford a minimal amount of control over the range of the plot. These controls may later be upgraded with designs that incorporate feedback from this first iteration, but this first iteration should at least be sufficient to meet scientists' needs.
- Now that the range of the plot can be changed, the potential for elements in the plot to render outside the bounds of the plot is much higher. A lot of the code changes here are about ensuring those elements (histogram bars, control points, the line and gradient connecting control points, etc.) clip properly against the min and max of the plot.
- There's also a bit of extra polish in here, like making tick marks on the X axis a bit smarter.
- State for the range of the plot is currently held by the `TfEditor` component which contains the plot itself. This is problematic because that state is reset whenever the component is unmounted and mounted again, including when the channel controls are collapsed or the user navigates to another tab of the control panel. A future PR (#381) will lift this state into the `ChannelState` type which is persistent while the `App` component is mounted.

### Screenshots

image

### DIFF
--- a/src/aics-image-viewer/components/TfEditor/index.tsx
+++ b/src/aics-image-viewer/components/TfEditor/index.tsx
@@ -581,23 +581,27 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
           {/* plot axes */}
           <g ref={xAxisRef} className="axis" transform={`translate(0,${innerHeight})`} />
           <g ref={yAxisRef} className="axis" />
-          {/* control points */}
+          {/* "advanced mode" control points */}
           {controlPointCircles}
           {/* "basic mode" sliders */}
           {!props.useControlPoints && (
             <g className="ramp-sliders">
-              <g transform={`translate(${xScale(u8ToAbsolute(props.ramp[0], props.channelData))})`}>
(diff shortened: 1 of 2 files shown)


frasercl:
Closes #373

ShrimpCryptid (approved):
Tested these changes out in the preview and the behavior feels sensible! This was really well-thought out :>

ascibisz (approved):
Looks good to me!

PR #379, opened 2025-04-18
Store multiscene collections in local storage

frasercl:
Review time: small-medium

# Problem

Scientists want to be able to open a selection of files in BioFile Finder as a multi-scene collection in Vol-E. Currently, collections of scenes to open are encoded as a delimited list of URLs in a query parameter in the app URL. This presents a couple problems for the BFF use case:
- Since users can select an arbitrary number of images in BFF, they could very easily create unworkably long Vol-E URLs.
- We want to bring metadata for each scene along from BFF as well. We don't currently have a way to parse metadata out of a query parameter, and even if we did, this could create unworkably long URLs from even a single scene!

# Solution

For this use case, we've decided to put scene URLs and metadata in Vol-E's *local storage* instead. This presents a new wrinkle: sites can't write to a different site's local storage. But they can send a message and have the site write it themselves. So the whole process looks like:
- BFF `iframe`s in Vol-E
- BFF calls `postMessage` to transfer scenes and metadata
- Vol-E receives that message, drops the scenes and metadata in its own local storage, and responds that it's received the message
- BFF opens a new Vol-E window with a query parameter telling Vol-E to open scenes from local storage
- Vol-E does that

So this change implements the Vol-E side of that procedure: receiving messages and writing/reading local storage.

## Change summary

- Add a new route to Vol-E, pointing to a component that renders nothing but listens to messages from external sites and writes local storage
- Ensure that other routes are lazy-loaded so that, when we access that route, the entire Vol-E viewer doesn't come along for the ride
- Add code to pull URLs and metadata from local storage when the `url` query parameter is set to `storage`
- Tweak how the `metadata` prop is interpreted: if the top-level of the metadata object is an array, assume that it's a list of metadata objects for each scene, not the metadata for a single image.

### DIFF
--- a/public/LocalStorageReceiver.tsx
+++ b/public/LocalStorageReceiver.tsx
@@ -9,7 +9,7 @@ const LocalStorageReceiver: React.FC = () => {
         return;
       }
 
-      window.localStorage.setItem("scenes", encodeImageUrlProp(e.data));
+      window.localStorage.setItem("url", encodeImageUrlProp(e.data));
       (e.source as Window)?.postMessage("SUCCESS", e.origin);
     };
 
--- a/public/LocalStorageReceiver.tsx
+++ b/public/LocalStorageReceiver.tsx
@@ -3,7 +3,12 @@ import React from "react";
 const LocalStorageReceiver: React.FC = () => {
   React.useLayoutEffect(() => {
     const receiveMessage = (e: MessageEvent) => {
-      console.log(e.data);
+      if (e.origin === window.location.origin) {
+        return;
+      }
+
+      window.localStorage.setItem("scenes", e.data);
+      (e.source as Window)?.postMessage("SUCCESS", e.origin);
     };
 
     window.addEventListener("message", receiveMessage);
--- a/public/LocalStorageReceiver.tsx
+++ b/public/LocalStorageReceiver.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 
 const LocalStorageReceiver: React.FC = () => {
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     const receiveMessage = (e: MessageEvent) => {
       console.log(e.data);
     };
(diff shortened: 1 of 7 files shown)


public/index.tsx
```diff
@@ -31,15 +29,21 @@ if (locationUrl.hash !== "" || isEncodedPathUrl(locationUrl)) {
   window.history.replaceState(null, "", newRelativePath);
 }
 
-const routes = [
+// TODO these components are now lazy loaded. Should they get `Suspense`s around them? What should we fall back to?
+const routes: RouteObject[] = [
   {
     path: "/",
-    element: <LandingPage />,
+    lazy: async () => ({ Component: (await import("../website/components/LandingPage")).default }),
```
ShrimpCryptid, line 36: Just curious, were you able to test at all that the lazy loading worked and the bundle size was kept smaller?

ShrimpCryptid (approved):
LGTM! I think this is a really clever solution :>

rugeli (approved):
LGTM! I agree on Peyton's opinion on error messaging for failed uploads. And it might be helpful to make the success messages more informational by showing the path or number of scenes that were stored.

public/LocalStorageReceiver.tsx
```diff
@@ -0,0 +1,39 @@
...
+        return;
+      }
+
+      window.localStorage.setItem("url", encodeImageUrlProp(message));
+      if (e.data.meta !== undefined) {
+        window.localStorage.setItem("meta", JSON.stringify(message.meta));
+      } else {
+        window.localStorage.removeItem("meta");
+      }
+      (e.source as Window)?.postMessage("SUCCESS", e.origin);
```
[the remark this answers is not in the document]
frasercl: This and @rugeli's comment building on it below are good ideas. Currently on the BFF branch, nothing really happens if this fails &mdash; it costs us nothing to wait for a success message that never comes. But it could be good to start sending back failure messages for if we want to display them on the BFF side later.

public/index.tsx
frasercl: I definitely confirmed that (`console.log`s I dropped inside of) scripts behind this lazy load don't run until that route actually loads. I did my best to root around the built files to work out where everything went and whether code was getting split properly, but wasn't able to 100% convince myself that way.

ShrimpCryptid: Okay, great! Very nice :>

PR #381, opened 2025-04-28
Persistent transfer function editor range

frasercl:
Review time: small (~5-15min)

# Problem

#370 added the ability to modify the range of the transfer function editor's plot. That initial implementation put the state which determines the plot range in the `TfEditor` component itself. This has a significant drawback: that state is destroyed whenever the `TfEditor` component unmounts, including when the channel is disabled, the channel's settings are collapsed, or the user navigates to another tab in the control panel.

# Solution

Lift the relevant state up into the `ChannelState` type, which is held by `ViewerStateProvider` and persists as long as the `App` component is mounted.

This is a little bit of a strange place to put this information, since that type is otherwise reserved for settings which change the appearance of the channel in the viewport. But (I think) it's the least unnatural place for it to go right now.

### DIFF
--- a/website/utils/test/url_utils.test.ts
+++ b/website/utils/test/url_utils.test.ts
@@ -246,6 +246,8 @@ describe("Channel state serialization", () => {
       { x: 255, opacity: 1.0, color: [255, 255, 255] },
     ],
     ramp: [0, 255],
+    lockPlotToDataRange: true,
+    plotMax: 255,
   };
   const DEFAULT_SERIALIZED_CHANNEL_STATE: ViewerChannelSettingParams = {
     col: "ff0000",
@@ -380,6 +382,8 @@ describe("Channel state serialization", () => {
         useControlPoints: false,
         controlPoints: [],
         ramp: [0, 255],
+        lockPlotToDataRange: true,
+        plotMax: 255,
       };
       const serializedCustomChannelState: Required<Omit<ViewerChannelSettingParams, "lut">> = {
         col: "03ff9d",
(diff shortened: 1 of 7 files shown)


website/utils/test/url_utils.test.ts
```diff
@@ -380,6 +382,8 @@ describe("Channel state serialization", () => {
         useControlPoints: false,
         controlPoints: [],
         ramp: [0, 255],
+        lockPlotToDataRange: true,
+        plotMax: 255,
```
ShrimpCryptid, line 386: Should this be serialized to the URL?

ShrimpCryptid: Maybe worth adding a TODO here at the very least

ShrimpCryptid (approved):
Left a question about saving these settings to the URL, but otherwise LGTM!

website/utils/test/url_utils.test.ts
frasercl: My thought was no, since these are configuration parameters for how a UI element gets drawn, and are never passed down to the vole-core level to modify how the image is actually displayed. Though `TfEditor` may be on its way to being an actual data display in its own right, rather than just a control element. In that case I think it would make more sense to preserve these parameters. I'll give it a TODO and an issue.

rugeli (approved):
LGTM! Comments in `DTYPE_RANGE` really helped clear my questions and changes look clean and reasonable!

website/utils/test/url_utils.test.ts
frasercl: Created #384 for this

PR #383, opened 2025-05-01
1-index clipping panel sliders

### LINKED ISSUE #371: 1-index sliders
## Description

Sliders in the clipping panel (Z, Y, X, time, scene) are currently, confusingly, 0-indexed.

## Expected Behavior

They should be 1-indexed instead, for clarity.

frasercl:
Review time: xsmall (~5min)

## Problem

Resolves #371: sliders in the clipping panel with just one handle (slice in 2d modes, time, and/or scene) are 0-indexed. This is confusing.

## Solution

Keep the underlying state 0-indexed, but modify the component that handles the sliders and spinboxes to display it as 1-indexed. This keeps the code changes relatively limited, but adds some odd complexity in the specific component, since that component also has to handle the case where the slider sets a range between _two_ values.

### Steps to Verify:

1. Open the PR preview link
2. Mess with the sliders and spinboxes in the "Clipping" panel (especially in XY mode)
3. Verify everything works as expected

### DIFF
--- a/src/aics-image-viewer/components/AxisClipSliders/index.tsx
+++ b/src/aics-image-viewer/components/AxisClipSliders/index.tsx
@@ -41,22 +41,25 @@ const SliderRow: React.FC<SliderRowProps> = ({
   onEnd,
 }) => {
   const isRange = vals.length > 1;
-  // If slider is a range, handles represent slice *edges*: the range around only the max slice is [max-1, max], e.g.
-  // If slider is not a range, handle represents 0-indexed slices: last slice is at max-1
-  const inputMax = isRange ? max : max - 1;
+  // If slider is a range, handles represent slice *edges*: the range around only the 1st slice is [0, 1]; the range
+  // around only the last is [max-1, max].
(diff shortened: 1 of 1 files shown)


src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -41,22 +41,25 @@ const SliderRow: React.FC<SliderRowProps> = ({
   onEnd,
 }) => {
   const isRange = vals.length > 1;
-  // If slider is a range, handles represent slice *edges*: the range around only the max slice is [max-1, max], e.g.
-  // If slider is not a range, handle represents 0-indexed slices: last slice is at max-1
-  const inputMax = isRange ? max : max - 1;
+  // If slider is a range, handles represent slice *edges*: the range around only the 1st slice is [0, 1]; the range
+  // around only the last is [max-1, max].
```
ShrimpCryptid, line 45: Maybe it would be clearer to say that the bounds are `[min, max)` or `inclusive min, exclusive max`?

ShrimpCryptid (approved):

ascibisz (approved):

PR #391, opened 2025-06-05
Add React hooks lints

frasercl:
Review time: medium (~25min)

[`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) is an eslint plugin by the React team that enforces the "[Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks)" &mdash; specifically, that hooks must be called from the top-level scope of a React component, and that hooks which contain functions and their dependencies must [include all their dependencies](https://react.dev/learn/removing-effect-dependencies). (That last requirements includes effects, which can cause potentially strange consequences! Though the React team may be working on an [escape hatch](https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event) for this requirement. It's all very confusing.)

Other projects like BFF include this plugin, with generally positive impacts on code quality. I'm looking to add this as an initial step in a larger refactor, which would eventually separate the block of volume loading code in the main `App` component into a self-contained hook and/or provider component. So this PR deliberately excludes `App` from the new lints in anticipation of that refactor, but every other file is brought into compliance.


src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
```diff
@@ -47,13 +47,7 @@ const ViewerWrapper: React.FC<ViewerWrapperProps> = (props) => {
 
   React.useEffect(() => {
     view3dviewerRef.current!.appendChild(props.view3d.getDOMElement());
-    props.view3d.setAutoRotate(props.autorotate);
```
ShrimpCryptid, line 50: Where does this get set now?

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -500,7 +500,7 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
         .attr("y", (len) => binScale(len))
         .attr("height", (len) => innerHeight - binScale(len));
     },
-    [xScale, props.channelData, props.channelData.histogram, innerWidth, innerHeight, plotMinU8, plotMaxU8]
```
ShrimpCryptid, line 503: Will the histogram object reference update?

src/aics-image-viewer/components/CellViewerCanvasWrapper/index.tsx
frasercl: App/index.tsx, ln 676

PR #393, opened 2025-06-11
Fix view mode and reset camera controls

frasercl:
Review time: tiny (~5min)

Fixes two controls which were broken by #391:

- View mode could not be changed. I tracked this to code which synced certain props to viewer state: I broke this code out of a hacky hook in #391, but did a bad job and caused props to be synced to state on _every render_, not just ones on which the props changed.
- Pressing the reset camera button threw an error. This was due to the persistent JS problem of what `this` is - a `bind` call fixed it.

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -832,6 +832,7 @@ const App: React.FC<AppProps> = (props) => {
     (): [number, number, number] => (image ? image.imageInfo.physicalPixelSize.toArray() : [1, 1, 1]),
     [image?.imageInfo.physicalPixelSize]
   );
+  const resetCamera = useMemo(() => view3d.resetCamera.bind(view3d), [view3d]);
 
   return (
     <StyleProvider>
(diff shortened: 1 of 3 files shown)


src/aics-image-viewer/components/App/index.tsx
```diff
@@ -832,6 +832,7 @@ const App: React.FC<AppProps> = (props) => {
     (): [number, number, number] => (image ? image.imageInfo.physicalPixelSize.toArray() : [1, 1, 1]),
     [image?.imageInfo.physicalPixelSize]
   );
+  const resetCamera = useMemo(() => view3d.resetCamera.bind(view3d), [view3d]);
```
toloudis, line 835: makes me wonder if there are more like this one

src/aics-image-viewer/components/Toolbar/ViewModeRadioButtons.tsx
```diff
@@ -1,8 +1,9 @@
-import React from "react";
 import { Radio } from "antd";
 import { RadioChangeEvent } from "antd/lib/radio";
+import React from "react";
```
toloudis, line 3: is this change still needed?

toloudis (approved):

src/aics-image-viewer/components/App/index.tsx
frasercl: This is hopefully relatively isolated, since it's pretty rare to have controls directly attached to vole-core methods. Most controls change React state, then those changes are propagated by effects, and those effects are arrow functions in which `this` behaves sensibly. This required `resetCamera` to never be wrapped in an arrow function between the button's event listener and vole-core.

ShrimpCryptid (approved):

Issue #372, opened 2025-04-14
Prominently show file name in viewer


frasercl:
## Use Case

Scientists frequently want to open multiple images in multiple Vol-E instances, or open a single multi-scene collection, and compare images. This is challenging to impossible without a prominently visible filename to distinguish windows.

## Solution

Filename should be shown somewhere, probably the top bar next to the Vol-E name. As a bonus, include the filename in the browser tab title (e.g. "Vol-E &mdash; my_image.ome.zarr")

As another bonus, we could show names of individual scenes, or some useful metadata about the scene, as a tooltip when the user interacts with the scene slider.

frasercl:
See also #22

PR #324, opened 2024-09-24
React 18

toloudis:
Estimated time to review: 20-30 min? 
  Most of the changes are converting "class components" to "functional components"

Coauthored with @frasercl 
Resolves much of #191 (React lifecycle takes up extra cpu cycles during playback)

* Upgrade to React 18, bump some other dependency versions.

* Convert to functional components everywhere.

* Actually use production mode for production builds.

* Some updates to SharedCheckbox

future TODOs clean up some seemingly harmless console warnings:
1. 
```
Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>. Error Component Stack
    at p (<anonymous>)
    at p (<anonymous>)
    at li (<anonymous>)
    at O (styled-components.br…ser.esm.js:32:23549)
    at ul (<anonymous>)
    at O (styled-components.br…ser.esm.js:32:23549)
    at div (<anonymous>)
    at O (styled-components.br…ser.esm.js:32:23549)
    at div (<anonymous>)
    at LandingPage (index.tsx:235:33)
```
2. (antd Tooltip uses findDOMNode)
```
Warning: findDOMNode is deprecated and will be removed in the next major release. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node Error Component Stack
    at SingleObserver (index.js:23:24)
    at ResizeObserver (index.js:24:24)
    at eval (index.js:61:34)
    at Tooltip (Tooltip.js:23:32)
    at eval (index.js:52:16)
    at div (<anonymous>)
    at div (<anonymous>)
    at div (<anonymous>)
    at Toolbar (index.tsx:61:24)
```
3. (nouislider-react is out of date and still using defaultProps)
```
Warning: v: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead. Error Component Stack
    at v (nouislider-react.umd…ction.min.js:1:3395)
    at SmarterSlider (SmarterSlider.tsx:13:35)
    at span (<anonymous>)
    at span (<anonymous>)
    at SliderRow (index.tsx:34:3)
    at PlaySliderRow (index.tsx:123:54)
```

PR #394, opened 2025-06-24
Move volume loading into a hook

frasercl:
# Problem

The `App` component currently has several somewhat disparate roles:
1. The root component of all our viewer UI
2. Owns the `View3d`, and handles syncing React state to it
3. Owns the `Volume`, and handles (re)loading it

This makes `App` pretty big and pretty tightly coupled, in a way that can get in the way of development. 1. and 2. fit pretty naturally together (both have to do with the UI within `App`'s screen real estate), but 3. is a bit of an outlier. It's not clear that one component needs complete control over both the `Volume` and the means to display it; it makes just as much sense, and is more nicely modular, to load the volume in one place, and pass it down to the component that displays and modifies it.

This work is immediately motivated by #372, which needs access to the image from outside `App`, but the added modularity here is a broader win and should make other issues easier in the future as well.

# Solution

Add a hook, `useVolume`, which accepts a few parameters for how to load an image and returns the image, plus some additional tools for interacting with it in a reactive way.

Also, bring `App` up to speed with newly-added React hooks lints. #391 held off on applying these lints to `App` in anticipation of this refactor to image loading hooks. With that done, the whole project is in compliance.


src/aics-image-viewer/shared/utils/sliderValuesToImageValues.ts
```diff
@@ -32,8 +32,8 @@ export const gammaSliderToImageValues = (sliderValues: [number, number, number])
 // Density and brightness are overloaded for the two supported rendering modes in the volume viewer.
 // These formulae are somewhat ad-hoc based on what will subjectively look reasonable in the viewer, and should be considered tweakable.
 // vole-core expects to see numbers from 0..1 for density, brightness, and alpha.
```
ShrimpCryptid, line 34: Consider adding this as a comment to the methods below (expected slider value and expected output)

src/aics-image-viewer/shared/utils/sliderValuesToImageValues.ts
```diff
@@ -32,8 +32,8 @@ export const gammaSliderToImageValues = (sliderValues: [number, number, number])
...
-export const densitySliderToImageValue = (sliderValues: number, _isPT: boolean): number => +sliderValues / 100.0;
+export const densitySliderToImageValue = (sliderValues: number): number => +sliderValues / 100.0;
```
ShrimpCryptid, line 35: Is the `+{number}` coercion necessary? Are there ever instances where `sliderValues` is not a number?

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -0,0 +1,409 @@
...
+
+export type ReactiveVolume = {
+  image: Volume | null;
+  /**
+   * Indicates the load status of each channel:
+   *
+   * - `-1` indicates the channel has not yet loaded and will get some extra initialization (e.g. LUTs) when it loads.
+   * - `0` indicates the channel has been loaded once, but is currently waiting for new data.
+   * - `1` or greater indicates the channel is loaded. Note this is not exactly `1` to handle multiple simultaneous
+   *   loads: if multiple loads are issued before the first one completes, _incrementing_ rather than _setting_ the
```
ShrimpCryptid, line 55:
```suggestion
   *   loads; if multiple loads are issued before the first one completes, _incrementing_ rather than _setting_ the
```

ShrimpCryptid, line 55:
```suggestion
   *   loads; if multiple loads are issued before the first one completes, _incrementing_ rather than _setting_ the
```

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -0,0 +1,409 @@
...
+   *   version number means we react to each load when it completes, rather than just the first.
+   */
+  channelVersions: number[];
```
ShrimpCryptid, line 58: I wish this was an array of objects containing a load status Enum + a version number...

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -149,247 +116,95 @@ const setIndicatorPositions = (
...
   const view3d = useConstructor(() => new View3d());
   if (props.view3dRef !== undefined) {
     props.view3dRef.current = view3d;
   }
-  const loadContext = useConstructor(
-    () => new VolumeLoaderContext(CACHE_MAX_SIZE, QUEUE_MAX_SIZE, QUEUE_MAX_LOW_PRIORITY_SIZE)
-  );
 
-  // install loadContext into view3d
-  view3d.loaderContext = loadContext;
```
ShrimpCryptid, line 190: note to self check for this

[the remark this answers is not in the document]
frasercl: addressed in allen-cell-animated/vole-core#324

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -0,0 +1,409 @@
...
+    );
+
+    if (allLoaded && inInitialLoadRef.current) {
+      inInitialLoadRef.current = false;
+    }
+
+    return noneLoaded ? ImageLoadStatus.REQUESTED : allLoaded ? ImageLoadStatus.LOADED : ImageLoadStatus.LOADING;
+  }, [channelVersions, channelSettings, maskChannelName, loadThrewError]);
+
+  const setIsLoading = useCallback(() => {
```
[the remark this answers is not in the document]
frasercl: I definitely agree it's odd, since it's meant to modify state specifically to enter into the "loading" state. We're more used to something like `setIsLoading(loading: boolean)`. `updateLoadStatus` sounds more general to me though, as if it's picking up changes that may move into or out of the "loading" state? I also considered `beginLoad`, but that sounds like it's actually beginning the process rather than just reflecting it in state, or `setStateForLoad`, which is both clunky and non-specific. Maybe best of bad options?

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -0,0 +1,409 @@
...
+      const viewerChannelSettings = getCurrentViewerChannelSettings();
+      const grouping = makeChannelIndexGrouping(channelNames, viewerChannelSettings);
+      setChannelGroupedByType(grouping);
+
+      // compare each channel's new displayName to the old displayNames currently in state:
+      // same number of channels, and each channel has same displayName
+      const allNamesAreEqual = channelNames.every((name, idx) => {
+        const displayName = getDisplayName(name, idx, viewerChannelSettings);
+        return displayName === channelSettings[idx]?.displayName;
+      });
```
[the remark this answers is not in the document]
frasercl: Looks like it does. That function is traversing the new image's channel names, not the previous one's; if the previous list is empty that shouldn't pose a problem.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -403,121 +218,23 @@ const App: React.FC<AppProps> = (props) => {
...
-        const thisChannelSettings = viewerState.current.channelSettings[channelIndex];
-        onChannelDataLoaded(v, thisChannelSettings!, channelIndex);
-      });
-    } catch (e) {
-      showError(e);
-      throw e;
-    }
+  const hasRawImage = !!(props.rawData && props.rawDims);
+  const numScenes = hasRawImage ? 1 : ((props.imageUrl as MultisceneUrls).scenes?.length ?? 1);
+  const numSlices: PerAxis<number> = image?.imageInfo.volumeSize ?? { x: 0, y: 0, z: 0 };
```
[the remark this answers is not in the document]
frasercl: I don't think this value gets used if `image` is missing regardless, but 1 does still seem safer than 0.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -149,247 +116,95 @@ const setIndicatorPositions = (
...
-        const ramp = controlPointsToRamp(thisChannel.lut.controlPoints);
-        // now manually remap control points using the channel's old range
-        const { controlPoints } = thisChannelsSettings;
-        const remappedControlPoints = remapControlPointsForChannel(controlPoints, oldRange, thisChannel);
-        changeChannelSetting(channelIndex, { controlPoints: remappedControlPoints, ramp: ramp });
+      const showParentImage = imageType === ImageType.fullField && parentImageUrl !== undefined;
+      const path = showParentImage ? parentImageUrl : imageUrl;
+      // Don't reload if we're already looking at this image
+      if (isEqual(path, imageUrlRef.current)) {
+        return scenes;
```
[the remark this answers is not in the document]
frasercl: Hm, that is weird... I didn't notice it. Nonetheless, it seems to be the only way that we can return an identical value from `useMemo`? So maybe it's fine?

src/aics-image-viewer/components/useVolume.ts
frasercl: I definitely would prefer better typing, but what I like about the bare numbers is that they capture the possible states more tightly, where in the enum-plus-version approach the version field is unused in all but the "loaded" state. Outside of this hook these values are only ever used as a "key" to determine when to run channel update effects, and these semantics are barely used (there's one `> 0` check for "is this channel loaded"). This could be a union with a string-typed enum (`"init"` and `"reload"`?)?

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -0,0 +1,409 @@
...
+    setIsLoading,
+    onChannelDataLoaded,
+  ]);
+  // of the above dependencies, we expect only `sceneLoader` to change.
+
+  const setTime = useCallback(
+    (view3d: View3d, time: number): void => {
+      if (image && !inInitialLoadRef.current) {
+        view3d.setTime(image, time, onChannelDataLoaded).catch(onError);
+        setIsLoading();
```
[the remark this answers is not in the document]
ShrimpCryptid: Would it make sense to reverse the lines here to make that more clear?

ShrimpCryptid (approved):

Issue #374, opened 2025-04-14
Transfer function editor redesign pt. 2 USER TESTING/INTERVIEWS RESEARCH


frasercl:
## Use Case

We've done the basic work to show absolute intensities in the transfer function editor (#320) and to allow the transfer function range to be modified (#370). But there are some lingering questions about the current implementation, both technical and UX.
- X axis minimum can only ever be 0 or the minimum of the data. Is that okay?
- Are the current controls sufficient? Is there another way we can let users control the plot that would be better?
- What if users want to "zoom to current data range" and then lock the plot there?
- Is the gradient over the plot too opaque?
- ZEN includes a checkbox next to its "automatic" level-setting algorithms that, when checked, applies that algorithm on every time change. When that checkbox is checked, the algorithm buttons go from push buttons which apply the algorithm to the transfer function once to radio buttons that select exactly one algorithm to apply whenever the time changes. Is that something users want in Vol-E?
- ZEN's percentile algorithm also has a configurable min/max %. We have a percentile algorithm, "Default," which is fixed at 50% min and 98% max. Is that sufficient, or should we also offer a configurable option?

## Solution

- Test with users: does the current UI make sense? How can we improve it?
- Implement polished UI

lynwilhelm:
Script:
https://docs.google.com/document/d/1egX6Ur1B4qzcFS_ykZieCp_k0uGUK98u7VmyskPTe5M/edit?tab=t.0

Will begin to schedule soon

frasercl:
@lynwilhelm and I interviewed Jess and Serge. Notes are in the document above. 


1. Shrink buttons in the channel settings panel?
2. Currently, numeric inputs will not update their internal state to the maximum if the user inputs a number past the maximum, which can lead to confusion. (E.g. Serge entered 70,000, when the max is 65,535, and he was unable to see that change until he clicked away.)
3. Language change: "Ramp min/max" becomes "Levels min/max?" Or just "Levels?" (Might be overloaded with a "levels" slider in the Advanced Settings panel?)
4. Add a "plot min" numeric input. Give both it and the current plot max input a single shared label ("Plot min/max" or "Histogram min/max") and align everything together, analogous to the current "Ramp min/max" row above the plot.
5. Remove "Lock to data range" checkbox and the progressive disclosure of the "Plot max" input. Rather than having a mode which always adjusts the range when the data changes, we want a push button that sets the range once.
[...]

PR #396, opened 2025-06-26
Transfer function range editing v2

### LINKED ISSUE #395: Implement transfer function editor redesign pt. 2
## Use Case

@lynwilhelm and I conducted a couple user tests of transfer function editor range controls version 1 (implemented in #370), and she's designed a new version off the feedback we received and the discussions we had in #374.

## Solution

Implement new UI as designed.

frasercl:
Review time: small (~5-15min)

Resolves #395: implements @lynwilhelm's design for better transfer function editor range controls from #374.

In the first version of range controls (implemented in #370) the user can either "lock to data range" or freely adjust only the maximum of the plot's domain. In this version, users always have full control over both the min and max, and we're never "locked" to any range. Instead, the user can use new push buttons "Fit to data" and "Full range" to apply the default ranges from the previous "locked" and "unlocked" states once.

Also, buttons are more compact, and the language of a preset transfer function button is changed a bit.

| Before | After |
| ------- | ------- |
| image | image |

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -297,6 +297,7 @@ const App: React.FC<AppProps> = (props) => {
         controlPoints: controlPoints,
         ramp: controlPointsToRamp(ramp),
         // set the default range of the transfer function editor to cover the full range of the data type
+        plotMin: DTYPE_RANGE[thisChannel.dtype].min,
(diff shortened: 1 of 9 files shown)


rugeli (approved):
Tried out the pr preview and looked great! The `fit to data` button is much more easy and straightforward! One small non-blocking thing I noticed: the plotMin/Max inputs allow any text like "asdfg", we might want to restrict them to numbers only.

interim17 (approved):

PR #403, opened 2025-08-01
Fix/raw array loading

toloudis:
Problem
=======
What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

Solution
========
What I/we did to solve this problem

with @pairperson1

## Type of change

* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change requires a documentation update
* This change requires updated or new tests

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/aics-image-viewer/shared/utils/sceneStore.ts
+++ b/src/aics-image-viewer/shared/utils/sceneStore.ts
@@ -33,6 +33,7 @@ export default class SceneStore {
       let options: Partial<CreateLoaderOptions> = {};
       if (typeof path === "object" && !Array.isArray(path)) {
         options.rawArrayOptions = path;
+        options.fileType = VolumeFileFormat.DATA;
         path = "";
       }
 
--- a/src/aics-image-viewer/shared/utils/sceneStore.ts
+++ b/src/aics-image-viewer/shared/utils/sceneStore.ts
@@ -5,6 +5,7 @@ import {
   PrefetchDirection,
   RawArrayLoaderOptions,
   Volume,
+  VolumeFileFormat,
   VolumeLoaderContext,
 } from "@aics/vole-core";
 import { ThreadableVolumeLoader } from "@aics/vole-core/es/types/loaders/IVolumeLoader";


src/aics-image-viewer/components/useVolume.ts
```diff
@@ -175,6 +175,11 @@ const useVolume = (
   const [channelGroupedByType, setChannelGroupedByType] = useState<ChannelGrouping>({});
 
   const onChannelDataLoaded = useCallback(
+    // TODO FIXME this can be called immediately during volume creation, before
+    // the channel settings are set up.  This can happen during RawArrayLoader usage.
+    // The Raw Array Loader does not go across thread boundaries and it doesn't ahve to do
+    // anything asynchronous, so this just gets called immediately.
+    // But the onChannelLoadedRef callback is going to index into the channel state array, which is not set up yet.
```
toloudis, line 182: @frasercl I could use a consult with you on how best to handle this.  In my testing, wrapping the body of this function below in a setTimeout(..., 0) was sufficient to get the raw array to load without error.

frasercl (approved):

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -235,7 +268,6 @@ const App: React.FC<AppProps> = (props) => {
       view3d.updateLuts(image);
       view3d.onVolumeData(image, [channelIndex]);
 
-      view3d.setVolumeChannelEnabled(image, channelIndex, thisChannelSettings.volumeEnabled);
```
ShrimpCryptid, line 238: Is this getting called somewhere else?

toloudis: good catch, not sure why this is removed here.  @frasercl does this look right?

PR #409, opened 2025-08-15
Show image titles

### LINKED ISSUE #372: Prominently show file name in viewer

frasercl:
# Problem

Resolves #372: shows file name in the top bar of the viewer.

# Solution

Viewer gets a new prop `onImageTitleChange` with a callback that gets called whenever we learn the name of the image. `AppWrapper` keeps the title in its state, and shows the title in its `Header` component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The volume shader itself is a heavily modified version of one that has distant o
 ## to use
 
 - `https://vole.allencell.org/?url=path/to/ZARR`
-- for more url parameters, see [`URL_SPEC.md`](documentation\URL_SPEC.md)
+- for more url parameters, see [`URL_SPEC.md`](documentation/URL_SPEC.md)
 
 or as React component:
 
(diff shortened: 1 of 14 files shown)


toloudis:
Suggestion/first thought: 
* add `imageName` or `imageNameOverride` prop (string)
* update landing page to pass that prop with the DataSetEntry.loadParams
* add implementation to use that name if present

Issue #416, opened 2025-08-19
Use zustand for state management


frasercl:
Zustand is a minimal state management library that has been used successfully in TFE. Given that our current state management system adds some additional trouble when adding new state and using it, we could make the same change here.

PR #430, opened 2025-09-02
Zustand 1: setup, viewer state

frasercl:
Part 1 of refactoring state management to use zustand (#416). This PR...

- ...sets up a zustand store,
- moves most *global* (not per-channel) viewer state into the store,
- detaches the components that depend on that viewer state from the old state solution, and instead subscribes them to the store,
- replaces most hooks for syncing global state to the vole-core layer with zustand subscribers.

Still to come:
- (de)serializing to/from URL
- resetting state
- subscribers/syncing for channel settings
- cleanup, organization, and eventual removal of the previous state manager

### DIFF
--- a/src/aics-image-viewer/state/subscribers.ts
+++ b/src/aics-image-viewer/state/subscribers.ts
@@ -77,21 +77,27 @@ const selectAxisClipUpdateInfo = (axis: AxisName): ((store: ViewerStore) => Axis
 export const subscribeImageToState = (store: typeof useViewerState, view3d: View3d, image: Volume): (() => void) => {
   const axisClipUpdater = (axis: AxisName) => {
     return ({ region: [minval, maxval], slice, viewMode }: AxisClipUpdateInfo) => {
-      // Logic to determine axis clipping range, for each of x,y,z,3d slider:
-      // if slider was same as active axis view mode:  [viewerSettings.slice[axis], viewerSettings.slice[axis] + 1.0/volumeSize[axis]]
-      // if in 3d mode: viewerSettings.region[axis]
-      // else: [0,1]
       let isOrthoAxis = false;
       let axismin = 0.0;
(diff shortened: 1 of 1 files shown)


src/aics-image-viewer/state/subscribers.ts
```diff
@@ -0,0 +1,179 @@
...
+      REF_EQ
+    ),
+
+    // interpolation
+    store.subscribe(
+      select("interpolationEnabled"),
+      (enabled) => view3d.setInterpolationEnabled(image, enabled),
+      REF_EQ
+    ),
+
```
ShrimpCryptid, line 169: Maybe easy question, do all of these values get assigned on `view3d` from state at initialization?

src/aics-image-viewer/state/subscribers.ts
```diff
@@ -0,0 +1,179 @@
...
+      let axismin = 0.0;
+      let axismax = 1.0;
+      if (viewMode === ViewMode.threeD) {
+        axismin = minval;
+        axismax = maxval;
+      } else {
+        isOrthoAxis = activeAxisMap[viewMode] === axis;
+        const oneSlice = 1 / image.imageInfo.volumeSize[axis];
+        axismin = isOrthoAxis ? slice : 0.0;
+        axismax = isOrthoAxis ? slice + oneSlice : 1.0;
```
ShrimpCryptid, line 94: Maybe explicitly note here that, in orthographic slice mode, the region is reset to the [0, 1] range for all axes other than the one currently being viewed. The comment on L80-L83 could maybe be clarified too?

ShrimpCryptid (approved):
Left lots of question/comments but nothing is blocking. I can immediately see some benefits to using Zustand selectors for removing boilerplate and exposing unused dependencies!

interim17 (approved):

PR #434, opened 2025-09-16
Zustand 2: URL params

frasercl:
Part 2 of converting state management to use zustand. This PR moves URL (de)serialization to use the zustand store, rather than the old `ViewerStateProvider`. In support of this, I also added a new function to the store, `mergeViewerSettings`, and broke out state validation to be shareable between this new function and the current `changeViewerSetting`.

Still to come:
- state reset
- subscribers/syncing for channel settings
- cleanup, organization, and eventual removal of the previous state manager

### DIFF
--- a/website/components/Modals/ShareModal.tsx
+++ b/website/components/Modals/ShareModal.tsx
@@ -20,11 +20,8 @@ type ShareModalProps = {
 const ModalContainer = styled.div``;
 
 const ShareModal: React.FC<ShareModalProps> = (props: ShareModalProps) => {
-  const selectViewerSettingsShallow = useShallow(selectViewerSettings);
-  const selectChannelSettingsShallow = useShallow((store: ViewerStore) => store.channelSettings);
-
-  const viewerSettings = useViewerState(selectViewerSettingsShallow);
-  const channelSettings = useViewerState(selectChannelSettingsShallow);
+  const viewerSettings = useViewerState(useShallow(selectViewerSettings));
+  const channelSettings = useViewerState(useShallow((store: ViewerStore) => store.channelSettings));
 
   const [showModal, setShowModal] = useState(false);
   const modalContainerRef = useRef<HTMLDivElement>(null);


website/components/Modals/ShareModal.tsx
```diff
@@ -3,27 +3,29 @@ import { ShareAltOutlined } from "@ant-design/icons";
...
+};
 
 const ModalContainer = styled.div``;
 
 const ShareModal: React.FC<ShareModalProps> = (props: ShareModalProps) => {
+  const selectViewerSettingsShallow = useShallow(selectViewerSettings);
+  const selectChannelSettingsShallow = useShallow((store: ViewerStore) => store.channelSettings);
+
+  const viewerSettings = useViewerState(selectViewerSettingsShallow);
+  const channelSettings = useViewerState(selectChannelSettingsShallow);
```
ShrimpCryptid, line 27: Would it make sense to make these one-liners if `selectViewerSettingsShallow` is not being used anywhere? e.g.
`const viewerSettings = useViewerState(useShallow(selectViewerSettings))`

website/components/AppWrapper.tsx
```diff
@@ -76,7 +83,7 @@ export default function AppWrapper(): ReactElement {
 
   return (
     <div>
-      <ViewerStateProvider viewerSettings={viewerSettings}>
```
ShrimpCryptid, line 79: Why did this get removed? (Was it because ShareModal still needed viewer state to be in ViewerStateProvider?)

src/aics-image-viewer/state/store.ts
```diff
@@ -43,8 +43,27 @@ const VIEWER_SETTINGS_CHANGE_HANDLERS: ViewerSettingChangeHandlers = {
...
+): Partial<ViewerState> => {
+  const changeHandler = VIEWER_SETTINGS_CHANGE_HANDLERS[key];
+
+  if (changeHandler) {
+    // some settings have custom change handlers to avoid creating illegal states; if this one has one, call it
+    return changeHandler(value, currentState);
+  } else {
+    // if not, merge the new value to the current one (if applicable) and return
+    const currentValue = currentState[key];
+    const nextValue = isRecord(currentValue) && isRecord(value) ? { ...currentValue, ...value } : value;
```
ShrimpCryptid, line 59: Is it possible for `currentValue` to not be a record while `value` is? Or is the check just for satisfying type checking?

website/components/Modals/ShareModal.tsx
frasercl: Yeah probably! Just felt weird having a hook in a hook like that.

src/aics-image-viewer/state/store.ts
frasercl: Probably mostly for type checking

PR #443, opened 2025-10-02
refactor: Moved FirebaseRequest class into `src/aics-image-viewer`

ShrimpCryptid:
Problem
=======
I was working on https://github.com/allen-cell-animated/cell-feature-explorer/issues/237, which requires that we export the url parsing logic (`parseViewerUrlParams()` in `url_utils`). However, I was running into a compilation bug if I just moved `url_utils` directly.

It turns out that this is because it was importing the firebase configuration from `public`, which caused Typescript to compile `public` as a separate module. This change removes `url_utils`'s dependency on `public` and moves the firebase database initialization to `public/index.tsx`.

Solution
========
What I/we did to solve this problem

with @pairperson1

## Type of change

* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change requires a documentation update
* This change requires updated or new tests

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/aics-image-viewer/shared/utils/firebase/index.ts
+++ b/src/aics-image-viewer/shared/utils/firebase/index.ts
@@ -1,6 +1,7 @@
-import { DocumentReference, QuerySnapshot, DocumentData } from "@firebase/firestore-types";
+import { DocumentData, DocumentReference, FirebaseFirestore, QuerySnapshot } from "@firebase/firestore-types";
 
-import { firestore } from "./configure-firebase";
+// TODO: These types are shared with Cell Feature Explorer. Can they be moved to
+// a shared package?
 
 export interface DatasetMetaData {
   name: string;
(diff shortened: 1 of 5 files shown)


src/aics-image-viewer/shared/utils/firebase/index.ts
```diff
@@ -1,6 +1,13 @@
-import { DocumentReference, QuerySnapshot, DocumentData } from "@firebase/firestore-types";
+import {
+  DocumentData,
+  DocumentReference,
+  DocumentSnapshot,
+  FirebaseFirestore,
+  QuerySnapshot,
+} from "@firebase/firestore-types";
```
ShrimpCryptid, line 7: Had to add return types to functions in here, so a few new types were added.

PR #437, opened 2025-09-22
Move isosurface slider to histogram

frasercl:
Review time: medium (~25min)

Resolves allen-cell-animated/vole-core#430: moves the isosurface slider onto the histogram, and reorders settings to accommodate the histogram as the primary input component for both volume and surface settings. The latter comes off a design from Thao [here](https://www.figma.com/design/EJCCnElvRvObznBhN519sl/Vol-E---CFE?node-id=1459-1513&t=1XF1DP25fjy8R5pW-1).

This issue was previously temporarily addressed in #436, which made the isosurface feature usable again after it was broken by our move to multiple datatype ranges.

| Before | After |
| ------- | ------ |
| image | image |

PR #462, opened 2025-11-06
refactor: Store ramp and control points as raw intensity values (keep range pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of 2 for changes to support https://github.com/allen-cell-animated/cell-feature-explorer/issues/242, ("Keep Vol-E settings when switching between files").

In Vol-E, we often refer to data either by a **raw intensity value** (e.g. the pixel value in data) or by a **histogram bin index** in the range of [0, 255].  When rendering, we bin raw intensity values into one of 256 bins (similar to how a histogram bins data), and then we use a lookup table (or LUT, an array of 256 values) to determine how bright that value should be onscreen. Histogram bin indices were historically used in various places in the codebase because they map nicely to the LUT.  

This included the `ramp` and `controlPoints`, which are user-facing controls. Internally, they store histogram bin indices, but on the UI, they're shown as raw intensity values. The drawback of using histogram bin indices is that they need to be remapped each time new data is loaded in, since they're they're relative to the data (`0` is data min, `255` is data max). Because of float imprecision, this would cause "swimming", where intensity values that were supposed to be static would fluctuate between each frame.

vole-data-swimming-levels

This change updates both the `ramp` and `controlPoints` to store raw intensity values directly, preventing these float imprecision errors!

Solution
========
- Changed `ramp` and `controlPoints` to be stored as raw intensity values.
  - Removed code logic that remapped both on data load.
  - Changed LUT to be derived from `ramp` and `controlPoints` values.
- Updated URL parameters to accept ramp and control point values prefixed with a `v`, to indicate raw values.
  - This keeps backwards-compatibility with legacy behavior.
  - When sharing URLs, ramp and control points will be specified in raw values from now on.

Misc:
- Updated multiscene behavior so channels are reinitialized to defaults when switching between scenes.
- Changed plot min/max to initialize to the channel's data range by default, instead of the full data type range.

## Type of change

Steps to Verify:
----------------
1. Open the current build using this link: https://vole.allencell.org/viewer?manifest=https://dev-aics-dtp-001.int.allencell.org/users/peyton.lee/test-data/vole-bff.json&t=15&cam=pos:0.2566901:-0.2752941:4.805526,tar:0.2566901:-0.2752941:0&c0=ven:1,isv:34847.5,col:ffffff,cpt:4160:0:1:27176:0:1:32690.25:1:1:65296.25:1:1,ram:27176:32690.25&c1=isv:245,col:ffffff,cpt:53:0:1:105.6367:0:1:131.2031:1:1:436.4961:1:1,ram:105.63671875:131.203125&c2=isv:402,col:ffffff,cpt:66:0:1:139.6094:0:1:181.6719:1:1:736.3711:1:1,ram:139.609375:181.671875
2. Open the volume channel settings to view the plot. Play through time on volume. The plot handles will jitter.

1. Repeat the above steps with the preview link. The plot handles will not move: https://allen-cell-animated.github.io/vole-app/pr-preview/pr-462/viewer?manifest=https://dev-aics-dtp-001.int.allencell.org/users/peyton.lee/test-data/vole-bff.json&t=15&cam=pos:0.2566901:-0.2752941:4.805526,tar:0.2566901:-0.2752941:0&c0=ven:1,isv:34847.5,col:ffffff,cpt:4160:0:1:27176:0:1:32690.25:1:1:65296.25:1:1,ram:27176:32690.25&c1=isv:245,col:ffffff,cpt:53:0:1:105.6367:0:1:131.2031:1:1:436.4961:1:1,ram:105.63671875:131.203125&c2=isv:402,col:ffffff,cpt:66:0:1:139.6094:0:1:181.6719:1:1:736.3711:1:1,ram:139.609375:181.671875

Screenshots (optional):
----------------------

fixed-histogram-jitter

### DIFF
--- a/documentation/URL_SPEC.md
+++ b/documentation/URL_SPEC.md
@@ -141,7 +141,7 @@ Examples:
 
 ### Binning
 
-By default, the `lut` (lookup table) is specified in terms of histogram bin
+If the `lut` (lookup table) is specified with unprefixed numbers, they are interpreted as histogram bins. (This is for legacy reasons and backward compatibility.)
 indices for legacy reasons.
 
 When loaded, each channel's raw intensity values are sorted into one of 256


src/aics-image-viewer/shared/utils/viewerChannelSettings.ts
```diff
@@ -35,8 +35,13 @@ export interface ViewerChannelSetting {
    * Defaults to false.
    */
   controlPointsEnabled?: boolean;
+  /**
+   * Control points for the transfer function, as histogram bin values.
+   */
   controlPoints?: ControlPoint[];
   ramp?: [number, number];
+  rawRamp?: [number, number];
+  rawControlPoints?: ControlPoint[];
```
ShrimpCryptid: I've been debating about the best way to handle this. Should I:
1. Keep `controlPoints` and `ramp` as-is, and comment that they are histogram bin indices?
1. Mark `controlPoints` and `ramp` as deprecated, and make a new, separate field that is explicitly for histogram bin indices?

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -160,10 +165,18 @@ const useVolume = (
...
+      setLoadThrewError(false);
+      setChannelVersions(
+        channelVersionsRef.current.map((version) =>
+          // For scenes, reinitialize all channels.
+          Math.min(version, loadType === LoadType.SCENE ? CHANNEL_INITIAL_LOAD : CHANNEL_RELOAD)
+        )
+      );
+    },
+    [channelVersionsRef, setChannelVersions]
+  );
```
ShrimpCryptid: Small behavioral fix that I was going to implement later but rolled into this change instead. 

Previously, multiscene files would reuse the same control points and ramp. Because both were histogram bin indices, they were relative to the range of the data and would look (relatively) correct when new data was loaded.

Because both are now pinned to actual data values, it's possible for them to become really blown out. For example, one scene might have a data range of [0, 50], while the next has a range of [1000, 50000]. Using the same ramp for both would cause the second scene to appear as a solid white box. 

Instead, now multiscene files will reinitialize channels when changing scenes and apply the default LUT behavior, as requested by users (@carolinecch). The next change (keep range pt. 2) will add controls for disabling this behavior and keeping the same ramp/control point values.

src/aics-image-viewer/shared/utils/viewerChannelSettings.ts
ShrimpCryptid: Dan: controlPoints2?
Dan: For readability, would prefer to keep `controlPoints` as the main parsed prop. But that is a breaking change.

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -266,32 +261,37 @@ const App: React.FC<AppProps> = (props) => {
           controlPoints: controlPoints,
           ramp: controlPointsToRamp(ramp),
           // set the default range of the transfer function editor to cover the full range of the data type
-          plotMin: range.min,
-          plotMax: range.max,
+          plotMin: thisChannel.rawMin,
+          plotMax: thisChannel.rawMax,
```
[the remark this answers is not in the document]
frasercl: Off the top of my head the reasons we settled on this were:
- Users may not spot outliers in subsequent timesteps that are out of range of the first one
- A large portion of the histogram may wander out of the initial window between the first loaded timestep and the one the user is looking at when they first open channel settings
- Users may not get a clear sense for what fraction of the representable range of their dataset they're seeing, and may not notice that that range is in fact very small.

Showing a uselessly zoomed-out view of the histogram isn't great, but it has the benefit of being the only failure mode of this default, being obviously wrong, and having a one-click fix.

I don't think this is a strong enough argument to be decisive, and certainly wouldn't be upset if we changed this default &mdash; Thao also recommended it in the designs for #437! In particular I think having more natural controls over histogram range like those just added to AGAVE (in allen-cell-animated/agave#297) might tip the balance here. But for the moment I come down just barely on the side of showing the full dtype range.

src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -21,50 +21,53 @@ const DEFAULT_CONTROL_POINT_COLOR_CODE = "1";
...
+const REGION_REGEX = new RegExp(
+  `^(${FLOAT_REGEX.source}:${FLOAT_REGEX.source})(,${FLOAT_REGEX.source}:${FLOAT_REGEX.source}){2}$`
+);
 
 const HEX_COLOR_REGEX = new RegExp(`(([0-9a-fA-F]{6})|${DEFAULT_CONTROL_POINT_COLOR_CODE})`);
-const NUMERIC_REGEX = /-?[0-9.]*/;
-const CONTROL_POINT_REGEX = new RegExp(`(${NUMERIC_REGEX.source}:${NUMERIC_REGEX.source}:${HEX_COLOR_REGEX.source})`);
+
+/** Represents control points specified by bin indices. */
+const CONTROL_POINT_REGEX = new RegExp(`(${FLOAT_REGEX.source}:${FLOAT_REGEX.source}:${HEX_COLOR_REGEX.source})`);
```
frasercl: `^` for these ones too? (could be no)

src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -334,7 +357,7 @@ const tryDecodeURLList = (url: string, delim: string | RegExp = ","): string[] |
   for (const u of urls) {
     try {
       new URL(u);
-    } catch (_e) {
+    } catch {
```
frasercl: ohhhh... did not know you could do that

frasercl (approved):

src/aics-image-viewer/components/shared/SliderRow/index.tsx
```diff
@@ -1,5 +1,6 @@
-import React from "react";
 import { NouisliderProps } from "nouislider-react";
+import React from "react";
```
ShrimpCryptid: This file  also changed from main

src/aics-image-viewer/shared/utils/urlParsing.ts
ShrimpCryptid: Not for these since multiple of them will be chained together

src/aics-image-viewer/components/ChannelsWidgetRow/index.tsx
```diff
@@ -1,13 +1,16 @@
 import { Channel } from "@aics/vole-core";
-import { Button, Checkbox, List } from "antd";
+import { Button, Checkbox, InputNumber, List } from "antd";
```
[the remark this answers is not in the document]
toloudis: can we merge zustand-4 so that the base branch becomes main again?

documentation/URL_SPEC.md
```diff
@@ -98,45 +98,19 @@ property pair being separated by **commas**.
...
+| -------- | --------------------------------------------------------------------------------- | ---------------------------------------------------------------------- | -------------------------------------- | ----------------------------------- |
+| `ven`    | Whether the volume is enabled.                                                    | `1` (enabled) or `0` (disabled)                                        | `0` (`1` for the first three channels) | `?c0=ven:1`                         |
+| `sen`    | Whether the isosurface is enabled.                                                | `1` (enabled) or `0` (disabled)                                        | `0`                                    | `?c0=iso:1`                         |
+| `isv`    | Isosurface value, the intensity value at which to generate the isosurface.        | Number in the range `[0, 255]`                                         | `128`                                  | `?c0=isv:195`                       |
```
toloudis: should `isv` change the range it names?  it is also a raw intensity now?  or is that going in a different pr?

documentation/URL_SPEC.md
```diff
@@ -164,3 +138,29 @@ Examples:
 | `p10:p95`   | Linear mapping from the 10th percentile to the 95th.       |
 | `m100:m150` | Linear mapping from the median to 1.5 times the median.    |
 | `autoij:0`  | Uses the "auto" algorithm from ImageJ.                     |
+
+### Binning
+
+By default, the `lut` (lookup table) is specified in terms of histogram bin
+indices for legacy reasons.
```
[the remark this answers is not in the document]
toloudis: I wouldnt' call this a "default".  I'd call it "unprefixed values".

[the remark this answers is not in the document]
toloudis: The scary thing is if we ever change the internal length of the LUT (say from 256 to 512)

src/aics-image-viewer/components/App/ChannelUpdater.tsx
```diff
@@ -71,8 +71,14 @@ const ChannelUpdater: React.FC<ChannelUpdaterProps> = ({ index, view3d, image, v
...
       }
+      const histogram = currentImage.getHistogram(index);
       const controlPointsToUse = useControlPoints ? controlPoints : rampToControlPoints(ramp);
-      const gradient = controlPointsToLut(controlPointsToUse);
+      // Convert control points from raw intensity values to histogram bin indices
+      const binIndexedControlPoints = controlPointsToUse.map((cp) => ({
+        ...cp,
+        x: histogram.findFractionalBinOfValue(cp.x),
+      }));
+      const gradient = controlPointsToLut(binIndexedControlPoints);
```
toloudis: am I thinking about this wrong, or should controlPointsToLut accept control points with raw intensity x coordinates?

src/aics-image-viewer/components/App/index.tsx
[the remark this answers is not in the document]
toloudis: I think as a user this feels like a nice default for presentation (data min/max instead of datatype min/max) but I would go to quantitative scientist users like Caroline,Chantelle,Antoine,Leigh to get the real scoop.
I definitely understand not wanting to miss outlier data because the graph was clipped, as they play through time or switch among multiscenes.  However, full data type range can be absurd for many kinds of  images.

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -269,14 +254,14 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
     () => d3.scaleLinear().domain([plotMin, plotMax]).range([0, innerWidth]),
     [innerWidth, plotMin, plotMax]
   );
-  const plotMinU8 = useMemo(() => absoluteToBin(plotMin, histogram), [plotMin, histogram]);
-  const plotMaxU8 = useMemo(() => absoluteToBin(plotMax, histogram), [plotMax, histogram]);
+  const plotMinU8 = useMemo(() => histogram.findFractionalBinOfValue(plotMin), [plotMin, histogram]);
+  const plotMaxU8 = useMemo(() => histogram.findFractionalBinOfValue(plotMax), [plotMax, histogram]);
```
toloudis: why do we still need plotMinU8 and plotMaxU8?

documentation/URL_SPEC.md
toloudis: sounds like a breaking change...

src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -22,50 +22,53 @@ const DEFAULT_CONTROL_POINT_COLOR_CODE = "1";
...
   `^${CONTROL_POINT_REGEX.source}(,${CONTROL_POINT_REGEX.source})*$`
 );
+
 /**
- * Matches a COLON-separated list of control points, where each control point is represented
- * by a triplet of `{x}:{opacity}:{hex color}`.
- * The hex color can be replaced with `w` to represent white (`ffffff`).
+ * Matches a COLON-separated list of control points, where each control point is
+ * represented by a triplet of `{x}:{opacity}:{hex color}`, where `x` is a
+ * histogram bin index or intensity value. The hex color can be replaced with
```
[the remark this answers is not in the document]
ShrimpCryptid: Will do!

src/aics-image-viewer/components/TfEditor/index.tsx
ShrimpCryptid: The histogram is still plotted in a domain relative to bin indices. I believe this gets overhauled in #437 by @frasercl

documentation/URL_SPEC.md
```diff
@@ -164,3 +138,29 @@ Examples:
 | `p10:p95`   | Linear mapping from the 10th percentile to the 95th.       |
 | `m100:m150` | Linear mapping from the median to 1.5 times the median.    |
 | `autoij:0`  | Uses the "auto" algorithm from ImageJ.                     |
+
+### Binning
+
+By default, the `lut` (lookup table) is specified in terms of histogram bin
```
toloudis:
```suggestion
If the `lut` (lookup table) is specified with unprefixed numbers, they are interpreted as histogram bins. (This is for legacy reasons and backward compatibility.)
```

PR #463, opened 2025-11-07
feat: Add option to keep intensity values when switching volumes (keep range pt. 2)

ShrimpCryptid:
Problem
=======
Part 2 of 2 for changes to support https://github.com/allen-cell-animated/cell-feature-explorer/issues/242, ("Keep Vol-E settings when switching between files").

This change adds a toggle button to the UI where users can choose to keep contrast settings for the ramp, control points, and isosurface when switching between volumes. It also fixes some bugs related to isovalue initialization.

Solution
========
- Added a new field to `ChannelState`, `keepIntensityOnNewVolume`. When `true`, the ramp, isovalue, and control point values will not be reinitialized when a new volume is loaded, keeping the existing values the same.
  - Added a related `ViewerChannelSetting` field, `keepIntensityRange`, which is used to initialize the `ChannelState` from props.
- Added serialization/deserialization for a new URL parameter `pin`.
- Updated unit tests.

Misc:
- Fixes a bug where isovalues provided via URL parameters would be overwritten (and incorrectly clamped) during initialization.
- Added a `npm run checks` command to run linting, typechecking, and testing steps for convenience.

## Type of change

Steps to Verify:
----------------
1. Open PR preview link: https://allen-cell-animated.github.io/vole-app/pr-preview/pr-463/viewer?manifest=https://dev-aics-dtp-001.int.allencell.org/users/peyton.lee/test-data/vole-zarr-indexing-bug.json
3. Open the contrast settings panel and switch between scenes. The levels min/max should automatically update to the new defaults.
4. Click the pin icon ("Keep intensity values when switching volumes") to turn on the keep range mode. Switch scenes. The levels min/max values should not change.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/a5a7eb16-87c2-4b31-b992-3b2d6f9ef186

https://github.com/user-attachments/assets/52a43e6c-079b-48e6-8540-11c175240922

image

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -249,7 +249,6 @@ const App: React.FC<AppProps> = (props) => {
         }),
       });
 
-      onImageTitleChange?.(newImage.imageInfo.imageInfo.name);
       view3d.updateActiveChannels(newImage);
       const unsubscribeView = subscribeViewToState(useViewerState, view3d);
       const unsubscribeImage = subscribeImageToState(useViewerState, view3d, newImage);
@@ -260,7 +259,16 @@ const App: React.FC<AppProps> = (props) => {
         removePreviousImage.current = undefined;
       };
     },
-    [props.viewerChannelSettings, view3d, onImageTitleChange, changeChannelSetting]
+    [props.viewerChannelSettings, view3d, changeChannelSetting]
+  );
+
+  const [loadedScene, setLoadedScene] = useState<number | undefined>(undefined);
+  const onChangeScene = useCallback(
+    (image: Volume, sceneIndex: number) => {
+      onImageTitleChange?.(image.imageInfo.imageInfo.name);
+      setLoadedScene(sceneIndex);
+    },
+    [onImageTitleChange]
   );
 
   const onChannelLoaded = useCallback(
@@ -342,6 +350,7 @@ const App: React.FC<AppProps> = (props) => {
     viewerChannelSettings: props.viewerChannelSettings,
     onCreateImage,
     onChannelLoaded,
+    onChangeScene,
     onError,
     maskChannelName,
   });
@@ -379,23 +388,24 @@ const App: React.FC<AppProps> = (props) => {
     });
   }, [view3d]);
 
-  const { metadata, metadataFormatter } = props;
-  const getMetadata = useCallback((): MetadataRecord => {
+  const { metadataFormatter } = props;
+  const metadata = useMemo((): MetadataRecord => {
     let imageMetadata = image?.imageMetadata as MetadataRecord;
     if (imageMetadata && metadataFormatter) {
       imageMetadata = metadataFormatter(imageMetadata);
     }
+    const propsMetadata = props.metadata;
 
     let sceneMeta: MetadataRecord | undefined;
-    if (Array.isArray(metadata)) {
+    if (Array.isArray(propsMetadata)) {
       // If metadata is an array, try to index it by scene
-      if (metadata.length >= numScenes) {
-        sceneMeta = metadata[scene];
+      if (propsMetadata.length >= numScenes && loadedScene !== undefined) {
+        sceneMeta = propsMetadata[loadedScene];
       } else {
-        sceneMeta = metadata[0];
+        sceneMeta = propsMetadata[0];
       }
     } else {
-      sceneMeta = metadata;
+      sceneMeta = propsMetadata;
     }
 
     if (imageMetadata && Object.keys(imageMetadata).length > 0) {
@@ -403,7 +413,7 @@ const App: React.FC<AppProps> = (props) => {
     } else {
       return sceneMeta ?? {};
     }
-  }, [metadata, metadataFormatter, image, numScenes, scene]);
+  }, [props.metadata, metadataFormatter, image, loadedScene, numScenes]);
 
   useEffect((): void => {
     const hasTime = numTimesteps > 1;
@@ -556,7 +566,7 @@ const App: React.FC<AppProps> = (props) => {
             saveIsosurface={saveIsosurface}
             onApplyColorPresets={applyColorPresets}
             viewerChannelSettings={props.viewerChannelSettings}
-            getMetadata={getMetadata}
+            metadata={metadata}
           />
         </Sider>
         <Layout className="cell-viewer-wrapper" style={{ margin: props.canvasMargin }}>
--- a/src/aics-image-viewer/components/TfEditor/index.tsx
+++ b/src/aics-image-viewer/components/TfEditor/index.tsx
@@ -1,12 +1,13 @@
 import { type Channel, type ControlPoint, type Histogram, Lut } from "@aics/vole-core";
 import { PushpinFilled, PushpinOutlined } from "@ant-design/icons";
-import { Button, Checkbox, InputNumber, Tooltip } from "antd";
+import { Button, Checkbox, Dropdown, InputNumber, Tooltip } from "antd";
 import * as d3 from "d3";
 import "nouislider/distribute/nouislider.css";
 import React, { useCallback, useMemo, useRef, useState } from "react";
 import { type ColorResult, SketchPicker } from "react-color";
 
 import { DTYPE_RANGE, LUT_MAX_PERCENTILE, LUT_MIN_PERCENTILE, TFEDITOR_DEFAULT_COLOR } from "../../shared/constants";
+import type { IsosurfaceFormat } from "../../shared/types";
 import {
   type ColorArray,
   colorArrayToObject,
@@ -18,6 +19,7 @@ import { useRefWithSetter } from "../../shared/utils/hooks";
 import type { ChannelState } from "../../state/types";
 
 import SliderRow from "../shared/SliderRow";
+import ViewerIcon from "../shared/ViewerIcon";
 
 import "./styles.css";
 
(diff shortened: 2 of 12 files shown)


src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -718,7 +718,7 @@ export function deserializeViewerChannelSetting(
     match: channelIndex,
     enabled: parseStringBoolean(jsonState[ViewerChannelSettingKeys.VolumeEnabled]),
     surfaceEnabled: parseStringBoolean(jsonState[ViewerChannelSettingKeys.SurfaceEnabled]),
-    isovalue: parseStringFloat(jsonState[ViewerChannelSettingKeys.IsosurfaceValue], 0, 255),
+    isovalue: parseStringFloat(jsonState[ViewerChannelSettingKeys.IsosurfaceValue], -Infinity, Infinity),
```
ShrimpCryptid: Noticed a bug where the isosurface was not being loaded via URL correctly!

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
```diff
@@ -216,16 +216,3 @@ export function rampToControlPoints([min, max]: [number, number]): ControlPoint[
     { x: Math.max(max + 1, TFEDITOR_MAX_BIN), opacity: 1, color: TFEDITOR_DEFAULT_COLOR },
   ];
 }
-
-/** Remaps an array of control points from an old range (as a 2-tuple) to a new one (extracted from a `Channel`) */
-export function remapControlPointsForChannel(
```
ShrimpCryptid: No longer used

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -1,4 +1,5 @@
 import { Channel, ControlPoint, Histogram, Lut } from "@aics/vole-core";
+import { PushpinFilled, PushpinOutlined } from "@ant-design/icons";
```
ShrimpCryptid: Image

Image

src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -126,6 +126,7 @@ export enum ViewerChannelSettingKeys {
   VolumeEnabled = "ven",
   SurfaceEnabled = "sen",
   IsosurfaceValue = "isv",
+  KeepRange = "pin",
```
ShrimpCryptid: Also open to opinions about this, I thought it was nice because it works well with the pushpin icon we're using on the UI

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -185,15 +185,16 @@ const App: React.FC<AppProps> = (props) => {
         return;
       }
 
+      const { channelSettings: channelState } = useViewerState.getState();
```
ShrimpCryptid: I really would love to update `channelSettings` to be `channelState` in the actual store...

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -243,43 +259,58 @@ const App: React.FC<AppProps> = (props) => {
...
+          isovalue,
+        });
+      } else {
+        // Expand the plot min and max to include the current data range as
+        // needed. This keeps the domain visually consistent when replaying
+        // through time or Z slices.
+        //
+        // NOTE: This must use the most up-to-date channel plot range, because
+        // onChannelLoaded can be called multiple times per channel during
+        // loading.
```
ShrimpCryptid: (see above where we're getting channelSettings out of state)

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -132,6 +133,20 @@ const App: React.FC<AppProps> = (props) => {
...
+    for (const key of Object.keys(viewerSettings || {}) as (keyof typeof viewerSettings)[]) {
+      const value = viewerSettings ? viewerSettings[key] : undefined;
+      const lastValue = lastViewerSettingsRef.current ? lastViewerSettingsRef.current[key] : undefined;
+      if (value !== undefined && !isEqual(value, lastValue)) {
+        useViewerState.getState().changeViewerSetting(key, value);
+      }
+    }
+    lastViewerSettingsRef.current = viewerSettings;
+  }, [viewerSettings]);
+
```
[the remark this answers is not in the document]
toloudis: behavior that was supposed to stay? why was it removed? just an oversight or something more intentional?

src/aics-image-viewer/components/ChannelsWidget.tsx
```diff
@@ -86,7 +86,7 @@ const ChannelsWidget: React.FC<ChannelsWidgetProps> = (props: ChannelsWidgetProp
     const channelName = channelNames[channelIndex];
     return channelName ? (
       <ChannelsWidgetRow
-        key={`${channelIndex}_${channelName}_${channelIndex}`}
+        key={channelIndex}
```
toloudis: what does this change do? (or, why was it so much more complicated before?)

src/aics-image-viewer/shared/utils/controlPointsToLut.ts
toloudis: is the vole-core remapControlPoints dead now too?

src/aics-image-viewer/components/ChannelsWidget.tsx
toloudis: Makes sense.  Still, odd that channel index was in there twice, before.

I can vaguely remember considering that if the channel name is different, then there;s a reasonable chance its data range and distribution could be wildly different.  as in a user moving between two unrelated images.

src/aics-image-viewer/components/useVolume.ts
```diff
@@ -232,24 +231,11 @@ const useVolume = (
...
-      if (allNamesAreEqual) {
-        const newChannelSettings = channelNames.map((channel, index) => {
-          return { ...channelSettings[index], name: channel };
-        });
-        initChannelSettings(newChannelSettings);
-        return newChannelSettings;
-      }
-
-      const newChannelSettings = channelNames.map((channel, index) => {
+      const newChannelSettings = channelNames.map((name, index) => {
```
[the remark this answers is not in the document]
ShrimpCryptid: I believe it was keeping the settings the same when switching between similar volumes (same # of channels + same names) and resetting all the channels if they differed? This also handled initializing new channels, which the changed code also handles.

This will be a breaking behavior change then, I think. Before, if you had two similar volumes, it would keep the ramp and control points at the same bin indices. Now, they will be re-initialized unless specified by the pin/lock button, and instead of keeping the same bin indices, they'll use the same raw intensity values.

src/aics-image-viewer/components/App/index.tsx
ShrimpCryptid: It's getting passed as a generic callback into `sceneLoader.createVolume()`, but is also being used as a callback for specific loading functionality (`view3d.setTime` and `sceneLoader.loadScene`).

src/aics-image-viewer/shared/utils/urlParsing.ts
frasercl: hmm... I like it

src/aics-image-viewer/components/App/index.tsx
frasercl: I say go for it

src/aics-image-viewer/components/App/index.tsx
```diff
@@ -248,38 +265,53 @@ const App: React.FC<AppProps> = (props) => {
...
         ? getDefaultViewerChannelSettings()
         : props.viewerChannelSettings;
       const thisChannel = image.getChannel(channelIndex);
-      const noLut = !thisChannelSettings || !thisChannelSettings.controlPoints || !thisChannelSettings.ramp;
+      const noLut = !channelState || !channelState.controlPoints || !channelState.ramp;
 
-      if (isInitialLoad || noLut || channelsToResetOnLoad.includes(channelIndex)) {
+      const hasOldRange = channelRangesRef.current[channelIndex] !== undefined;
+      const initializeToExistingRange = channelState && channelState.keepIntensityOnNewVolume && hasOldRange;
+      const initializeToDefaults = isInitialLoad && !initializeToExistingRange;
```
frasercl: Logic here is a bit hard to parse, could (optionally) use maybe one more comment

frasercl (commented):
Most pressing concern is the element hierarchy in `TFEditor`, but looks good!

src/aics-image-viewer/components/TfEditor/index.tsx
```diff
@@ -552,29 +554,44 @@ const TfEditor: React.FC<TfEditorProps> = (props) => {
...
+            style={{ height: 24, padding: "0 4px" }}
+            type={props.keepIntensityOnNewVolume ? "default" : "text"}
+          >
+            <span style={{ margin: 0 }}>
+              {props.keepIntensityOnNewVolume ? <PushpinFilled /> : <PushpinOutlined />}
+            </span>
+          </Button>
+        </Tooltip>
+        {!props.useControlPoints && (
+          <div className="tf-editor-control-row">
```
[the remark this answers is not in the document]
ShrimpCryptid: Ah, yeah, I was reusing the class for convenience sake. Let me make a new CSS class.

src/aics-image-viewer/components/App/index.tsx
ShrimpCryptid: Going to split this off into a separate PR I think just in case anything breaks

PR #465, opened 2025-11-11
Fix/import type lint

frasercl:
Review time: smallish (~10min)

Adds and applies new eslint rules for enforcing the use of `import type` wherever possible, to cut down on useless code in the built module:
- https://typescript-eslint.io/rules/consistent-type-imports
- https://typescript-eslint.io/rules/no-import-type-side-effects

Inspired by allen-cell-animated/timelapse-colorizer#786.

### DIFF
--- a/src/aics-image-viewer/components/App/ChannelUpdater.tsx
+++ b/src/aics-image-viewer/components/App/ChannelUpdater.tsx
@@ -1,9 +1,10 @@
-import { Lut, View3d, Volume } from "@aics/vole-core";
-import React, { useEffect } from "react";
+import { Lut, type View3d, type Volume } from "@aics/vole-core";
+import type React from "react";
+import { useEffect } from "react";
 
 import { controlPointsToLut, rampToControlPoints } from "../../shared/utils/controlPointsToLut";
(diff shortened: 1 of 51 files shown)


.eslintrc
```diff
@@ -32,16 +32,16 @@
     "no-unused-vars": 0, // Rely on typescript-eslint rule above
     "@typescript-eslint/explicit-function-return-type": ["error", { "allowExpressions": true }],
     "@typescript-eslint/typedef": ["error", { "parameter": true }],
+    "@typescript-eslint/consistent-type-imports": "warn",
```
ShrimpCryptid: In the original TFE PR, I ended up using the "inline-type-imports" fix style so types and imports stay on the same line. Did you decide against that stylistically?

https://github.com/allen-cell-animated/timelapse-colorizer/pull/786/files#diff-e7e195d3d8fa5b82fa51ed952f3a179d429cbc46c0432ce1f7b357c54c162822

toloudis: I would feel good about `import` and `import type` always being on separate lines for fast readability. but not a big deal either way

toloudis (approved):
LGTM, love this change for code cleanup

Issue #412, opened 2025-08-18
Metadata panel does not update when switching scenes


ShrimpCryptid:
## Description

When the metadata panel is open and the user switches to another scene, the metadata does not update unless the panel is closed and reopened.

## Expected Behavior

The panel should update when the scene does.

## Reproduction

1. Open https://allen-cell-animated.github.io/vole-app/pr-preview/pr-411/viewer?manifest=https://dev-aics-dtp-001.int.allencell.org/users/peyton.lee/test-data/vole-bff.json
2. Go to the metadata panel.
3. Switch to the next scene. The metadata will not update.
4. Close the metadata panel.
5. Reopen the metadata panel. The contents are now different.

## Environment

Windows 11, Chrome

jennifermccarty:
Dan - This is a very important one.

PR #467, opened 2025-11-29
Receive multiscenes and metadata from BFF

### LINKED ISSUE #461: Receive multiscenes with metadata from BFF
## Use Case

Scientists would like to open a selection of images as a multi-scene collection in Vol-E, with metadata.

## Solution

BFF pops up a new tab with `window.open`, then `postMessage`s the relevant data into that tab. (Interesting stretch: save history of selections to return to?)

We initially tried implementing this via `postMessage` and local storage, but ran into browser security issues (link PRs). Those issues may now have been resolved, or other approaches with more tradeoffs may still work.

frasercl:
Review time: medium-large (35-40min)

Resolves #461: receives large multi-scene collections and image metadata from BFF.

## Background

We previously tried to do this in #379 (along with the corresponding PR in BFF: AllenInstitute/biofile-finder#516). That approach attempted to use a trick to write data to Vol-E's local storage *before* opening a new window, which worked okay in testing but ran afoul of cross-origin isolation in production. This implementation receives data from BFF *after* the target Vol-E window has opened, which should avoid that issue.

The new approach works like this:
- BFF opens a Vol-E URL with two additional query parameters: `storageid` uniquely identifies the message it wants to send, and `msgorigin` is the origin of the app that opened Vol-E (the slice of the URL from the beginning through the port, i.e. https://bff.allencell.org), which we use to keep the browser's cross-origin policy happy when sending messages.
- Vol-E takes the presence of *both* these query parameters to mean that the opening window has a message it wants to send. It sends the value of `storageid` back to the opening window to signal it's loaded and ready for the message, and in the meantime falls back to loading from the `url` query parameter in case the rest of the transaction fails.
- When a message arrives back from BFF, it stores the contents in local storage using `storageid` as a key, replaces the image URLs and metadata in `App` props with those from the message, and removes `msgorigin` from the URL. If the user reloads or opens a new window with the same URL, Vol-E will see that *only* `storageid` is present and will attempt to load from local storage.

AllenInstitute/biofile-finder#621 implements the BFF side of this transaction and has more details.

## Storage

A large part of this PR is a new system for managing local storage, which manages entries like a least-recently-used cache, evicting entries when needed to keep its size under control. Some more details on how this works:
- Currently we recognize two types of entry: `scenes`, keyed by the value of `storageid` used to access them; and `meta`, keyed by image URL.
- Browsers have a quota for the maximum size of local storage (roughly 5 MiB, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage)). When inserting an entry would push us over the quota, we evict the oldest entries until space becomes available.
- Additionally, we enforce a maximum number of entries per type: 25 for `scenes`, and 300 for `meta`. Those numbers are pretty arbitrary.

In the future we could use this tracking to implement a kind of simple history browser for multi-scene selections that the user has opened from BFF!

## Testing

- start this app: `npm start`
- checkout `feature/vole-open-local` in biofile-finder
- in [.../useOpenWithMenuItems/index.tsx line 350](https://github.com/AllenInstitute/biofile-finder/pull/621/files#diff-80216214fcb8b17280859a5fca4de94214b43f92d6793c07efb3a397f1a0aa7aR350), change `VOLE_BASE_URL` to `http://localhost:9020/viewer`.
- start a dev build of the BFF web app: `npm --prefix packages/web run start`
- open a dataset and make a large selection of image files
- click "OPEN WITH" -> "Vol-E"
- **Expectation:** subject to the caveats and issues below, all the selected images should load, the "Scene" slider should begin on the image that was "focused" in BFF, and the metadata for the opened image should be visible in the "Metadata" panel.

## Remaining issues

Not everything about this system is ironed out as of this initial PR! Future fixes and polish should include:
- [x] Image titles and metadata do not update when switching scenes (tracked in #412).
- [ ] Currently, Vol-E considers the "fallback" load from the `url` query param to be the "initial load," and so doesn't automatically apply the "Default" transfer function preset to the "real" load that begins after a message arrives from BFF. Since that "fallback" load generally doesn't finish in time, all channels of the image the user eventually sees start with a default ramp from 0-255.
- [ ] Testing unfinished versions of this feature revealed that we could have much better error handling for cases where `App` tries to load an `undefined` image url, either because the `imageUrl` prop is undefined or because the `scene` prop is out-of-bounds.
- [ ] We could use a bit more validation that the message from BFF has all the expected parts. Nothing would stop a 3rd-party application from trying to use the same path but messing up the message type somehow!
- [x] (sigh) Even though `LocalStorage` will be annoying to mock, the storage manager in this PR could use some tests.
- [ ] The sharing dialog currently does not take into account the possibility that Vol-E could load from a large collection of URLs, and will dutifully generate exactly the kind of monstrously long viewer URL this change seeks to avoid.

### DIFF
--- a/website/components/AppWrapper.tsx
+++ b/website/components/AppWrapper.tsx
@@ -144,7 +144,10 @@ export default function AppWrapper(props: AppWrapperProps): ReactElement {
       window.addEventListener("message", receiveMessage);
       window.setTimeout(() => window.removeEventListener("message", receiveMessage), 60000);
       // Sending a message back lets the opening window know we're ready to receive more data
-      (window.opener as Window | null)?.postMessage(undefined, msgorigin);
+      const msg = {
+        appInfo: { name: "Vol-E", version: VOLEAPP_VERSION, coreVersion: VOLECORE_VERSION },
+      };
+      (window.opener as Window | null)?.postMessage(msg, msgorigin);
     }
 
     getViewerStateFromSearchParams();


src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -0,0 +1,162 @@
+import { mapValues } from "lodash";
+
+import type { MetadataRecord } from "../types";
+
+const QUEUE_KEY = "GLOBAL_ENTRY_QUEUE";
+
+const enum StorageEntryType {
+  Scenes = "scenes",
+  Meta = "meta",
```
ShrimpCryptid: Nit: Enum values conventionally all caps

src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -0,0 +1,162 @@
...
+ * handle exceeding the storage quota by removing items off the back of the `queue` until space is available.
+ *
+ * Also, unlike `setItem`, this function can accept `value`s of type `string[]`. This allows `queue` to also be passed
+ * into `value`, so that it can be written to storage *after* picking up any evictions that are made during the write.
+ */
+function setStorageItem(key: string, value: string | string[], queue: string[]): void {
+  let success = false;
+  while (!success) {
+    try {
+      window.localStorage.setItem(key, Array.isArray(value) ? value.join(",") : value);
```
ShrimpCryptid: Should values be escaped in case they already contain `,` characters?

src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -0,0 +1,162 @@
...
+      success = true;
+    } catch (e) {
+      if (e instanceof DOMException && e.name === "QuotaExceededError") {
+        const evictKey = queue.shift();
+        if (evictKey === undefined) {
+          const err = new Error("Tried to insert a single entry into local storage that was larger than the quota.");
```
ShrimpCryptid: How will users be notified about this? Is there an onscreen error message?

website/components/AppWrapper.tsx
```diff
@@ -39,24 +44,75 @@ export default function AppWrapper(props: AppWrapperProps): ReactElement {
...
   useEffect(() => {
     // On load, fetch parameters from the URL and location state, then merge.
     const locationArgs = location.state as AppDataProps;
-    parseViewerUrlParams(searchParams, props.firestore).then(
-      ({ args: urlArgs, viewerSettings: urlViewerSettings }) => {
-        setViewerSettings({ ...urlViewerSettings, ...locationArgs?.viewerSettings });
-        setViewerProps({ ...DEFAULT_APP_PROPS, ...urlArgs, ...locationArgs });
-      },
-      (reason) => {
+    let ignore = false;
```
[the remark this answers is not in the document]
ShrimpCryptid: Oh, is this to make a one-time useEffect that makes the linter happy about deps?

src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    index += 1;
+    return true;
+  });
+
+  // push new keys onto the end
+  for (const key of Object.keys(entries)) {
+    if (key.indexOf(",") !== -1) {
+      throw new Error(`Cannot write to local storage at key "${key}": keys may not contain commas.`);
```
ShrimpCryptid: Document in docstring for method comment?

src/aics-image-viewer/shared/utils/storage.ts
frasercl: Seeing both styles in this repo, but mainly this one

src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -272,14 +268,31 @@ class DataParams {
...
+  msgorigin?: string = undefined;
   /**
-   * The ID of a cell within the loaded dataset. Used with `dataset`.
+   * If this param is present, the opening window wants to send more scene URLs via `postMessage`.
+   *
+   * The value of this param should be the total number of scenes the opening window wants to send, so we can set up
+   * the app with the correct number of scenes before knowing what all the scenes are.
    */
-  id?: string = undefined;
+  msgscenes?: string = undefined;
```
toloudis: I thought we didn't actually need this.  Can you add to the comment about why this has to be sent separately from the actual data?

src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const enum StorageEntryType {
+  Scenes = "scenes",
+  Meta = "meta",
+}
+
+const MAX_ENTRIES: { [K in StorageEntryType]: number } = {
+  scenes: 25,
+  meta: 1000,
+};
```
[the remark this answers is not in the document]
toloudis: why not just bundle them all together?

src/aics-image-viewer/shared/utils/storage.ts
toloudis: For users it needs to be worded differently.  The wording is good for an engineer trying to debug

src/aics-image-viewer/shared/utils/storage.ts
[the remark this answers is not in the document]
frasercl: My thought there was to avoid duplication. Say the user keeps closing and reopening the same selection over and over. They'll get a different key for the multiscene they're opening every time, but because metadata is keyed by the URL, that entry will just get overwritten rather than taking up redundant space in storage.

documentation/EXTERNAL_APP_MESSAGE_SPEC.md
```diff
@@ -0,0 +1,115 @@
...
+When it receives this message, Vol-E will remove the `msgorigin` parameter from
+its URL and save the transferred data in the browser's local storage. Image
+metadata is stored based on the URL it describes, meaning that if the user
+later opens a different session that includes a scene for which Vol-E
+previously received metadata, that metadata record will be used in the new
+session. Scene URLs are saved by the value of `storageid`, meaning that the
+user can return to the same URL later and restore the same multi-scene
+collection. (If the external application does not send `scenes` as part of its
+message, Vol-E will also remove the `storageid` parameter from its URL, since
+it doesn't need to be kept around to identify a stored multi-scene collection.)
```
toloudis: if `scenes` is missing, doesn't that mean there is effectively nothing to load?   what does VolE do if there's a message passed that doesn't contain `scenes`?

documentation/EXTERNAL_APP_MESSAGE_SPEC.md
```diff
@@ -0,0 +1,115 @@
...
+  url.searchParams.append("url", scenes[startingScene]);
+
+  const handle = window.open(url);
+
+  // STEP 2: set up a handler for the "I'm ready!" message from Vol-E
+  const loadHandler = (event: MessageEvent) => {
+    // check that this is the message we're waiting for
+    if (event.origin !== url.origin || event.data !== storageid) {
+      return;
+    }
```
toloudis: consider `if event.source !== handle ...` 
Would that work?

src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -272,14 +268,31 @@ class DataParams {
    * See `ManifestJson` for the type definition.
    */
   manifest?: string = undefined;
+  /** The name of a dataset in the Cell Feature Explorer database. Used with `id`. */
+  dataset?: string = undefined;
+  /** The ID of a cell within the loaded dataset. Used with `dataset`. */
+  id?: string = undefined;
+  /** The key of a collection of scenes stored in local storage. Overrides `url` unless `msgorigin` is present. */
+  storageid?: string = undefined;
```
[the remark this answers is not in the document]
toloudis: oh man, I just commented something similar to Peyton.   I think the MessageEvent has a `.source` property that uniquely identifies the sender of the message.  I would say the version where VolE generates the storage key makes it simpler for client apps to pass data, and just a bit cleaner.  

I didn't quite follow the overwriting collections scenario.  How would that happen?  Maybe we can discuss.

documentation/EXTERNAL_APP_MESSAGE_SPEC.md
```diff
@@ -0,0 +1,115 @@
...
+returned from [`window.open`] for the following steps.
+
+## Step 2: Wait for Vol-E to load
+
+On load, if both the above query parameters are present, Vol-E will send a
+message back to the opening application at `msgorigin` containing the value
+of `storageid`. The external application code should wait for this message
+before moving to the next step, to ensure that the Vol-E window is open and
+ready to receive data. Ideally, it should also check that the message value
+matches `storageid`, to ensure it's sending data to the right Vol-E window (in
```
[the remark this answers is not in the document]
frasercl: Ran a test and confirmed this absolutely works to differentiate messages &mdash; meaning it's safe to send messages from BFF without having it generate a special key to identify each message. I'll see about refactoring to have Vol-E generate `storageid`s (`collectionid`s?) instead.

Minor question: any specific thoughts on what the content of the "ready" message that Vol-E sends back to BFF ought to be? In #379 it was just the string `"SUCCESS"`. Could just be `undefined`? Am I overthinking this? (Almost definitely)

documentation/EXTERNAL_APP_MESSAGE_SPEC.md
```diff
@@ -0,0 +1,105 @@
...
+
+The external application code will want to hold onto the [`WindowProxy`] object
+returned from [`window.open`] for the following steps.
+
+## Step 2: Wait for Vol-E to load
+
+On load, if the `msgorigin` query parameter is present, Vol-E will send a
+message back to the opening application. The external application code should
+wait for this message before moving to the next step, to ensure that the Vol-E
+window is open and ready to receive data. Ideally, it should also check that
```
toloudis: consider removing the word "Ideally" and just saying the calling app must ensure that the message came from the object created in step 1.   Also, I think it is ok to use the word "SUCCESS" or some other basic handshake acknowledgement as the message contents.  Is there any interesting _additional_ info that VolE could send in its message? A VolE version number perhaps?  Or something like 
`{ handshake:"SUCCESS", appInfo: {name: "Vol-E", version: "4.1.27"} }`

frasercl: Would it make any sense to send the generated `storageid` back as well, so the app can send the user to the same collection again later?

PR #473, opened 2026-01-08
Smarter URL prop encoding and decoding

frasercl:
Review time: small-medium (~20min) (most of the diff is package-lock.json!)

Makes the app smarter about encoding and decoding the `url` query parameter.

## Background

In its simplest form, this query parameter (`vole.allencell.org/viewer?url=...`) contains a single URL to the single image Vol-E should open. But there are multiple ways the user can "open" multiple image URLs at the same time, with corresponding URL param syntax:

- The user can open multiple *scenes*, which they can switch between in a single session, by concatenating multiple URLs with `+`: `example.com/image1.zarr+example.com/image2.zarr`
- Each scene can have multiple *sources*, whose data is combined into a single image, by concatenating multiple URLs with `,`: `example.com/source1.zarr,example.com/source2.zarr+example.com/scene2.zarr`

For a bunch of good reasons these URLs are generally percent-encoded to get rid of special characters. (For example, `http://` becomes `http%3A%2F%2F`.) But it's not clear whether the URLs ought to be encoded individually *before* concatenation or all at once *after* concatenation, or that any given app will be smart enough not to encode it multiple times. The most complex combination I've seen in the wild is scenes that were encoded, concatenated with `+`, then encoded *again*. This *happens* to work currently, but it implies potentially wilder combinations of encodings aren't outside the realm of possibility, and we ought to be prepared for those combinations (regardless of whether they're technically "correct").

## Solution

1. Overhaul how we decode nested lists of URLs in the `url` query parameter, to try to accept arbitrary combinations of encoding and concatenation
2. Add a unit test for 1.
3. Make Vol-E a little bit smarter about not attempting to encode a `url` param that's already encoded
4. This PR also includes a number of version bumps for dev packages, from me trying to get the environment to support the newish `URL.canParse` API for checking URL validity.

### DIFF
--- a/src/aics-image-viewer/shared/utils/urlParsing.ts
+++ b/src/aics-image-viewer/shared/utils/urlParsing.ts
@@ -315,15 +315,15 @@ class DataParams {
   dataset?: string = undefined;
   /** The ID of a cell within the loaded dataset. Used with `dataset`. */
   id?: string = undefined;
-  /** The key of a collection of scenes stored in local storage. Overrides `url` unless `msgorigin` is present. */
-  storageid?: string = undefined;
+  /** The key of a collection of scenes stored in local storage. Overrides `url`. */
+  collectionid?: string = undefined;
   /**
    * The origin of an opening window that wants to send a message to this window.
    *
-   * The presence of both this param and `storageid` implies that this window has just been opened by another app, and
-   * that the opening app has more data to send. Until that message is received, we fall back to `url`. Once that
-   * message arrives, the scenes to open are written to local storage at key `storageid` and `msgorigin` is removed,
-   * allowing the window to switch to reading local storage.
+   * The presence of this param implies that this window has just been opened by another app, and the opening app has
+   * more data to send. Until that message is received, we fall back to `url`. Once that message arrives, the scenes to
+   * open are written to local storage at a new `collectionid` and `msgorigin` is removed, allowing the window to
+   * switch to reading local storage.
    *
    * All this happens independently of URL parsing, so the only meaningful thing this parsing code does with this param
    * is check whether it is present.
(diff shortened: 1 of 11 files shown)


src/aics-image-viewer/shared/utils/urlParsing.ts
```diff
@@ -1027,25 +1050,24 @@ export async function loadFromManifest(
...
+ * @param search The query string to parse, which must be valid in the `URLSearchParams constructor
+ * @param firestore Optional Firestore instance. If provided, the function can load data from a
+ * Firestore dataset if the `dataset` and `id` parameters are provided.
  * @returns An object containing:
  * - `args`: Partial AppProps object.
  * - `viewerSettings`: Partial ViewerState object.
  *
- * `args` can be passed as props to the `ImageViewerApp`, and `viewerSettings`
- * can be passed to `ViewerStateProvider`.
+ * `args` can be passed as props to the `ImageViewerApp`, and `viewerSettings`  can be passed to `ViewerStateProvider`.
```
rugeli: tiny nit: extra space between `viewerSettings` and `can`

src/aics-image-viewer/shared/utils/test/urlParsing.test.ts
```diff
@@ -789,18 +781,62 @@ describe("parseViewerUrlParams", () => {
...
+    // GOAL: no matter how, where, or how many times URL encoding is applied when concatenating
+    // these scene URLs into one param, they should decode back to the same URLs.
+    const scenes = [
+      "https://example.com/image1.ome.zarr",
+      ["https://example.com/image2.ome.zarr", "https://example.com/image3.ome.zarr?foo=1%2C2%2C3"],
+    ];
+
+    // Recursively generate combinations of encoded and unencoded URLs
+    interface NestedArray<T> extends Array<T | NestedArray<T>> {}
+    const allEncodingCombindations = (urls: string | NestedArray<string>, delims: string[]): string[] => {
```
rugeli: nit: did you mean to call this constant `allEncodingCombinations`? (without the `d`)

PR #475, opened 2026-01-22
fix: Enable OIDC tokens for npm package publishing

ShrimpCryptid:
Problem
=======
Fixes failing NPM publishing action by switching to the new [OIDC (OpenID Connect) standard](https://docs.github.com/en/actions/concepts/security/openid-connect). This will also require us to update the NPM package as well to use [Trusted Publishers](https://docs.npmjs.com/trusted-publishers).

Solution
========
- Adds token permissions to the `publish.yml` action.

## Type of change

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,6 +16,7 @@ permissions:
   # See https://docs.github.com/en/actions/concepts/security/openid-connect
   # and https://docs.npmjs.com/trusted-publishers.
   id-token: write
+  contents: read
 
 jobs:
   lint:


.github/workflows/publish.yml
```diff
@@ -68,8 +80,7 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm publish
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
ShrimpCryptid: NPM tokens were deprecated by GitHub: https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/

PR #481, opened 2026-02-14
Refactor: create shared `ControlPanelRow` component

frasercl:
Review time: medium (20min)

Refactors the `ChannelsWidgetRow` and `CustomizeWidget` components to be based around a shared `ControlPanelRow` component. I had a few motivations for doing this:

- Currently, `ChannelsWidgetRow` is based on antd's [`List`](https://ant.design/components/list) component, which is (for some reason) deprecated in the latest major version.
- Even if `List` weren't deprecated, there's less and less technical reason to use it. As our visual requirements for channel settings have grown, the baseline styling that comes with the antd component has shifted from being a useful leg up to an annoying barrier to get around. #479 recently took another step in this direction.
- Channel settings and background/bounding box color customization share a similar layout (a row with a color picker, a title, and a bottom border), so they might as well share code too.
- Channel settings styling doesn't really have a consistent home at the moment &mdash; it's spread between the stylesheets for `ChannelsWidgetRow` and `ControlPanel`, the latter of which had some selectors that reached pretty far down into the former's territory. This change gave me an opportunity to establish a bit of additional clarity. Styles that pertain specifically to a single channel settings element should now be confined to `ControlPanelRow`'s stylesheet, those that apply to the whole control panel are in `ControlPanel`'s, and `ChannelsWidgetRow`'s stylesheet is deleted entirely.
- There's precedent for doing this in `SliderRow`, which is shared between `GlobalVolumeControls` and `TfEditor`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@aics/vole-app",
-  "version": "3.1.1",
+  "version": "3.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@aics/vole-app",
-      "version": "3.1.1",
+      "version": "3.2.1",
       "license": "ISC",
       "dependencies": {
         "@aics/vole-core": "^4.6.4",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aics/vole-app",
-  "version": "3.1.1",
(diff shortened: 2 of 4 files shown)


frasercl:
> looks good! removing `idx` from ColorPicker feels like a cleaner pattern

I truly don't know why I did that in the first place 😅

PR #482, opened 2026-02-17
Fix reading image metadata from storage

frasercl:
Review time: tiny (<5min).

Fixes a bug that prevents reading and displaying scene metadata from storage.

The function that reads scene metadata from storage (introduced in #467) accepts a `(string | string[])[]`, representing an array of *scenes*, each of which may be either a single image or an array of multiple data sources. Because it's not clear what we should do with metadata concatenated together from multiple sources, that function just yields `undefined` for any scenes of type `string[]`.

But in the feedback stage of #473, I simplified the function that *produces* those scene descriptions to only return `string[]` by representing single-source scenes as arrays of length 1. Now no metadata is ever read under any circumstances! So this changes the default behavior to just show the metadata of the first source, rather than giving up entirely.

### DIFF
--- a/src/aics-image-viewer/shared/utils/storage.ts
+++ b/src/aics-image-viewer/shared/utils/storage.ts
@@ -158,11 +158,14 @@ export function readStoredMetadata(
   skipCacheUpdate: boolean = false
 ): (MetadataRecord | undefined)[] {
   const keySet = new Set<string>();
-  const result = scenes.map((s) => {
+  const result = scenes.map((scene) => {
     // can't handle multi-source scenes (yet)
-    const scene = Array.isArray(s) ? s[0] : s;
+    const firstScene = Array.isArray(scene) ? scene[0] : scene;
+    if (firstScene === undefined) {
+      return undefined;
+    }
 
-    const globalKey = `${StorageEntryType.Meta}@${sanitizeStorageKey(scene)}`;
+    const globalKey = `${StorageEntryType.Meta}@${sanitizeStorageKey(firstScene)}`;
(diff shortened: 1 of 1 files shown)


src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -158,11 +158,9 @@ export function readStoredMetadata(
...
 ): (MetadataRecord | undefined)[] {
   const keySet = new Set<string>();
-  const result = scenes.map((scene) => {
-    if (Array.isArray(scene)) {
-      // can't handle multi-source scenes (yet)
-      return undefined;
-    }
+  const result = scenes.map((s) => {
+    // can't handle multi-source scenes (yet)
+    const scene = Array.isArray(s) ? s[0] : s;
```
ShrimpCryptid: This is probably an impossible case, but will `s[0]` ever return `undefined`?

ShrimpCryptid (approved):

src/aics-image-viewer/shared/utils/storage.ts
```diff
@@ -158,11 +158,9 @@ export function readStoredMetadata(
   skipCacheUpdate: boolean = false
 ): (MetadataRecord | undefined)[] {
   const keySet = new Set<string>();
-  const result = scenes.map((scene) => {
-    if (Array.isArray(scene)) {
-      // can't handle multi-source scenes (yet)
-      return undefined;
-    }
+  const result = scenes.map((s) => {
```
toloudis: nit: for readability I'd consider renaming `s` back to `scene`, and on line 163 rename scene to `firstSceneSource`

PR #499, opened 2026-03-27
Toggle sharing single scene or entire collection

frasercl:
Review time: smallish (~10-15min)

Loosens up the share modal's restrictions on trying to share very large scene collections and image metadata from BFF. This behavior was introduced in #488.

There are two types of data that can be passed to Vol-E from BFF which cause problems for shareable links:
1. Large sets of images, which can make the URL arbitrarily long if every single image's path is included in the URL.
2. Extra image metadata, which we have no means of including in the URL at all.

In #488 I added a restriction around 1. and introduced a warning message when either was present. I talked to UX about this, and we agreed to some revisions:
- Messages should be styled like info, not a warning
- Users should always be given the option to fit the entire scene collection into a URL, no matter how long the URL might become

## Screenshots

Before
image

After
image

### DIFF
--- a/website/components/Modals/ShareModal.tsx
+++ b/website/components/Modals/ShareModal.tsx
@@ -158,28 +158,35 @@ const ShareModal: React.FC<ShareModalProps> = (props: ShareModalProps) => {
         footer={null}
       >
         {urls.length > 0 && (
-          <FlexRow style={{ justifyContent: "space-between" }}>
-            <Radio.Group
-              value={showCurrentScene}
-              onChange={(e) => setShowCurrentScene(e.target.value)}
-              options={[
-                { value: false, label: "All scenes" },
-                {
-                  value: true,
-                  label: props.imageTitle !== undefined ? `Current scene (${props.imageTitle})` : "Current scene",
-                },
-              ]}
-            />
-            {hasTooManyScenes && (
-              <Tooltip
-                title={`Vol-E currently has ${urls.length} scenes open. Sharing a large scene collection requires a very long URL. We don't recommend including more than 4 scenes in a sharing link.`}
-                overlayStyle={{ zIndex: 10000 }}
-                placement="left"
-              >
-                <ExclamationCircleOutlined style={{ color: "var(--color-message-warning-text)", cursor: "pointer" }} />
-              </Tooltip>
-            )}
-          </FlexRow>
+          <Radio.Group
+            value={showCurrentScene}
+            onChange={(e) => setShowCurrentScene(e.target.value)}
+            options={[
+              {
+                value: true,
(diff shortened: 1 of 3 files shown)


website/components/Modals/ShareModal.tsx
```diff
@@ -18,38 +18,26 @@ import { FlexRow } from "../LandingPage/utils";
 
 type ShareModalProps = {
   appProps: AppDataProps;
+  imageTitle?: string;
   // Used to retrieve the current camera position information
   view3dRef?: React.RefObject<View3d | null>;
 };
 
+const MAX_URL_CHARACTERS = 2000;
```
toloudis, line 26: just curious, what was the rationale for choosing this number?  A comment would be helpful.  Because why not 4000 or 32767?

website/components/Modals/ShareModal.tsx
```diff
@@ -58,67 +46,40 @@ const encodeSceneUrl = (scene: string | string[]): string => {
...
-    hasStoredMetadata = readStoredMetadata(urls).some((meta) => meta !== undefined);
-
-    if (hasTooManyScenes) {
-      paramProps.scene = 0;
-    }
 
-    const serializedUrl = hasTooManyScenes
-      ? encodeSceneUrl(urls[viewerSettings.scene])
-      : urls.map(encodeSceneUrl).join("+");
+  const serializedUrl = showAllScenes ? urls.map(encodeSceneUrl).join("+") : encodeSceneUrl(urls[viewerSettings.scene]);
```
toloudis, line 80: I feel like `"+"` could be a symbolic constant called something like `SCENE_DELIMITER` since it probably shows up in a couple different places, in other parts of the code.

website/components/Modals/ShareModal.tsx
```diff
@@ -130,6 +91,9 @@ const ShareModal: React.FC<ShareModalProps> = (props: ShareModalProps) => {
     urlParams.push(viewerParamString);
   }
 
+  // location.pathname will include up to `.../viewer`
+  const baseUrl = location.protocol + "//" + location.host + location.pathname;
```
toloudis, line 95: surprised there isn't a common function to do this!

toloudis (approved):
A couple of recommendations (a comment and a symbolic constant) but overall LGTM.

website/components/Modals/ShareModal.tsx
frasercl: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers#answer-417184

I can put this in a comment!

ascibisz (approved):

PR #501, opened 2026-04-03
Feature/scale level controls 2

### LINKED ISSUE #401: Select OME-Zarr Resolution Level UX DESIGN/RESEARCH
## Use Case

Internal and external scientist users would like more information about, and control over, which scale level of their multiscale OME-Zarr images is loaded in Vol-E. Currently, we pick scale levels automatically based on a set memory limit, then in some circumstances we offset to a higher level (2D modes, where we can use the fact that a smaller range of data is required to load at higher resolution) or a lower level (playback, when we want to prioritize load speed over quality).

## Solution

Design new UI to give users the ability to control which multiscale level to load.

Questions to consider:
- Does this appear as a dialog before an image is opened (like AGAVE)? In the main app UI, allowing levels to be switched on-the-fly? Both?
- How much control over scale level offset behavior (described above) should we offer to users?
- How much metadata should we show users about the selected resolution level? Can/should we include an estimate of the memory required to load a scale level, like AGAVE?

I made this sketch of a possible design for the "on-the-fly" case, which might live as a new category in the "advanced settings" panel. I think a few of the design ideas in it are actually bad (some are good!), but they serve to give a sense of the design space:

Image

frasercl:
Review time: smallish (~10-20min)

Resolves #401: implements controls to explicitly pick multiscale levels in OME-Zarr images!

Following discussion with @thao-do, the control appears in the toolbar. Users can pick either "auto" or "manual" control. In "manual" mode, users can pick scale dimensions from an adjacent dropdown; in "auto" mode, that dropdown is disabled.

image

### Steps to verify

- `npm start`
- Pick any OME-Zarr image from the test landing page
- Try out the new controls!


ShrimpCryptid (commented):
Two quick pieces of feedback:
1. I'd like the dropdown to have a disabled styling and not change colors when hovered.
2. In the Auto mode, I'd like the dropdown to show the currently rendered scale level.

vole-scale-level

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -1,3 +1,5 @@
+// TODO this really ought to be exported at the top level...
+import type { VolumeDims } from "@aics/vole-core/es/types/VolumeDims";
```
toloudis, line 2: ouch, yes.  Or make it an exported member from some class that is already exported like Volume

toloudis (commented):
agree w/Peyton that the Auto mode should actually show the true resolution level. 
 
Also (not behavior related) I wonder if the "dropdown" chevron should vanish in Auto mode. Understood that styling layout is potentially reserved for another PR.

meganrm (approved):

src/aics-image-viewer/components/Toolbar/index.tsx
```diff
@@ -242,6 +248,34 @@ const Toolbar: React.FC<ToolbarProps> = (props) => {
...
+              >
+                {props.multiscaleDims.map((dims, idx) => {
+                  const [_t, _c, z, y, x] = dims.shape;
+                  return (
+                    <Select.Option key={idx} value={idx}>
+                      {x} x {y} x {z}
+                    </Select.Option>
+                  );
+                })}
+              </Select>
```
ShrimpCryptid, line 277: Consider setting a minWidth property on this? Currently the entire layout of the top toolbar shifts based on the  selected option (or when the selection is enabled/disabled).

src/aics-image-viewer/components/AxisClipSliders/index.tsx
```diff
@@ -310,7 +310,10 @@ const AxisClipSliders: React.FC<AxisClipSlidersProps> = (props) => {
                 label={""}
                 vals={[props.scene]}
                 max={props.numScenes}
-                onChange={([scene]) => props.changeViewerSetting("scene", scene)}
+                onChange={([scene]) => {
+                  props.changeViewerSetting("scene", scene);
+                  props.changeViewerSetting("useExactScaleLevel", false);
```
ShrimpCryptid, line 315: Should this remain the same if the scale levels are comparable?

frasercl: Would be nice, though we don't know at this point what the dimensions of the new scale level will be.

ShrimpCryptid (approved):

PR #505, opened 2026-04-17
Show cause in error alert

### LINKED ISSUE #323: Error.cause is never shown in our error banner, even when "more info" is clicked
## Description

We have volume loading errors and the `cause` field can sometimes have very specific data that identifies the type of error.  But that field is never shown in the front end error banner.  Seeing this info can greatly decrease debugging time.

## Expected Behavior

When users click more info, we should show this detail too, so they can use it in a bug report for us.  

## Reproduction

## Environment

frasercl:
Review time: small (5-10min)

Resolves #323: shows an error's `cause` field in the error alert. Also, adds a new error description for out-of-memory errors that occur when the user has manually selected a specific scale level.

image

### DIFF
--- a/src/aics-image-viewer/components/ErrorAlert/index.tsx
+++ b/src/aics-image-viewer/components/ErrorAlert/index.tsx
@@ -30,8 +30,7 @@ const UNKNOWN_ERROR_DESCRIPTION: React.ReactNode = (
   </>
 );
 
-type DescribedErrorTypes = Exclude<VolumeLoadErrorType, VolumeLoadErrorType.TOO_LARGE>;
-const ERROR_TYPE_DESCRIPTIONS: { [T in DescribedErrorTypes]: React.ReactNode } = {
+const ERROR_TYPE_DESCRIPTIONS: { [T in VolumeLoadErrorType]: React.ReactNode } = {
   [VolumeLoadErrorType.UNKNOWN]: (
     <>
       An unknown error occurred while loading volume data. Check the browser console (F12) for more details. If this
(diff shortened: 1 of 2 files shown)


src/aics-image-viewer/components/ErrorAlert/styles.css
```diff
@@ -23,8 +24,14 @@
...
-    font-weight: 600;
+  .ant-alert-message > div {
+    &:first-child {
+      font-weight: 600;
+    }
+
+    &:last-child {
+      font-style: italic;
+      color: var(--color-text-body);
+    }
```
ShrimpCryptid, line 35: Nit: Would it be better to use explicit classnames here?

ShrimpCryptid: e.g. Could there ever be a case where there is only one child, and both styles are applied?

ShrimpCryptid (approved):
LGTM! Would like to see an example with the preview build.

src/aics-image-viewer/components/ErrorAlert/styles.css
frasercl: Maybe! I think e.g. that an error without a `cause` would currently cause the error description to be styled like the cause. You're right, it's best to use classes here.

toloudis (approved):
My take is that this is a bit inelegant way to special case error messages that require dynamic data.  BUT I don't have any quick/better ideas, and the result is what we want!

src/aics-image-viewer/components/ErrorAlert/index.tsx
```diff
@@ -29,7 +30,8 @@ const UNKNOWN_ERROR_DESCRIPTION: React.ReactNode = (
   </>
 );
 
-const ERROR_TYPE_DESCRIPTIONS: { [T in VolumeLoadErrorType]: React.ReactNode } = {
+type DescribedErrorTypes = Exclude<VolumeLoadErrorType, VolumeLoadErrorType.TOO_LARGE>;
```
toloudis, line 33: maybe a slightly more future-proofed way to do this would be to maintain a separate list of the excluded error types.  I am assuming they are excluded only because they require dynamic data to produce their error message?

PR #506, opened 2026-04-17
Only prefetch priority directions during playback

pgarrison:
## Problem

Playback in time is slower than it could be. 

## Solution

During playback, only prefetch along the playing axis.

## Analysis

I estimate that this change increases Vol-E's framerate by about 22% (N=11, p=0.02). Average framerate increased from ~4.2 FPS to ~5.1 FPS in my benchmark reading real AICS data from S3.

Screenshot 2026-04-17 at 1 55 28 PM

This change should not meaningfully affect lag; in my test, lag increased 2% on average, but this was far from statistically significant (p=0.75).

## Type of change

- Performance improvement

## Steps to Verify

### DIFF
--- a/src/aics-image-viewer/shared/utils/sceneStore.ts
+++ b/src/aics-image-viewer/shared/utils/sceneStore.ts
@@ -5,6 +5,7 @@ import type {
   RawArrayLoaderOptions,
   Volume,
   VolumeLoaderContext,
+  ZarrLoaderFetchOptions,
 } from "@aics/vole-core";
 import { createDefaultMetadata, VolumeFileFormat } from "@aics/vole-core";
 import type { ThreadableVolumeLoader } from "@aics/vole-core/es/types/loaders/IVolumeLoader";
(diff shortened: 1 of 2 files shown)


src/aics-image-viewer/shared/utils/sceneStore.ts
```diff
@@ -96,4 +97,11 @@ export default class SceneStore {
       currentLoader.setPrefetchPriority(priority);
     }
   }
+
+  public updateFetchOptions(fetchOptions: Partial<ZarrLoaderFetchOptions>): void {
```
toloudis, line 101: I wonder if the function above this one (setPrefetchPriority) could also be part of the ZarrLoaderFetchOptions and thus combined with this function.  Just a thought.

toloudis (approved):
Love it.  Can't wait to try it out.

toloudis:
Incidentally, FPS can be slightly deceptive because it's fundamentally nonlinear in time.  Its inverse, SPF (or just "frame time") is a more linear measure of how long time is spent and thus how much time is being saved.

interim17 (approved):

src/aics-image-viewer/shared/utils/sceneStore.ts
pgarrison: I saw that `syncMultichannelLoading` and `setPrefetchPriority` were already separate but similar functions, so I continued that pattern. I could see merging all three.

pgarrison:
I had a measurement error (duplicate gpuFuse calls for the same frame) and can no longer demonstrate a statistically significant difference with vs. without this change.

I'm inclined to merge anyway since it seems like a good prefetch strategy.

Re-requesting your review @toloudis

pgarrison:
I have grouped a few small improvements together in #507 to create an improvement big enough to be statistically significant at my relatively small sample sizes (N=20, in that case).

PR #507, opened 2026-04-20
Prefetch performance improvements

pgarrison:
Problem
=======
Playback in time is slower than it could be.

This PR includes the changes from #506 
This PR has a [companion in vole-core](https://github.com/allen-cell-animated/vole-core/pull/396): please review that too.

Solution
========
* During playback, only pre-fetch along the playing axis. (Reviewed already in #506)
* Allow 6 pre-fetch requests in flight at a time (this aligns with browser limitation of 6 requests per domain)
* Increase the maximum number of timesteps that can be pre-fetched to 10 ([companion PR](https://github.com/allen-cell-animated/vole-core/pull/396))

Analysis
======
I estimate this change improves framerate almost to the maximum achievable within Vol-E, while increasing lag by about 16%. I am planning a future PR to improve lag.

In the comparison between main and the changes in this PR, median time per frame decreased from 375ms to 314ms. This difference was statistically significant (p=0.03, N=20).

The error bars below correspond to one standard deviation (it would be more appropriate to center the error bars on the mean than the median, but I think median performance is probably most relevant for our testing).

Screenshot 2026-04-20 at 4 31 19 PM

Above, I also compared Vol-E with the changes in this PR to a demo web app that makes the same network requests but doesn't render anything. In this test, the median time per frame was about 2% from optimal (301ms vs. 295ms). This difference was not statistically significant (p=0.12, N=26).

Below, I plot all the data from both tests, including lag and FPS. The comparison between main and this PR has a noticeable outlier in favor of this PR. With that data point removed, the improvement is still statistically significant (p=0.04).

Screenshot 2026-04-20 at 4 43 04 PM

I did not record lag metrics from the benchmark demo app.

## Type of change
Performance improvement

Steps to Verify:
----------------

### DIFF
--- a/src/aics-image-viewer/components/useVolume.ts
+++ b/src/aics-image-viewer/components/useVolume.ts
@@ -122,6 +122,7 @@ const useVolume = (
       setPlayingAxis(axis);
       // prioritize prefetching along the playing axis
       sceneLoader.setPrefetchPriority(axis ? [AXIS_TO_LOADER_PRIORITY[axis]] : []);
+      sceneLoader.updateFetchOptions({ onlyPriorityDirections: isPlaying });
       // sync multichannel loading so we don't show loaded channels one at a time
       sceneLoader.syncMultichannelLoading(isPlaying);
       if (image) {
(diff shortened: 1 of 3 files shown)


ShrimpCryptid (approved):

toloudis (approved):
LGTM

Issue #397, opened 2025-07-01
surfaces are too dim and they can stop working


toloudis:
## Description

surfaces are visible but too dark relative to their channel color settings.
And turning them off and on is broken.

## Expected Behavior

surfaces should be brighter.  But more importantly, the Surface checkbox can get into a bad state where you can't see the surface anymore.

## Reproduction

- load https://volumeviewer.allencell.org/viewer?url=https://animatedcell-test-data.s3.us-west-2.amazonaws.com/variance/3578.zarr/
- uncheck all volume channels
- check surface for: seg_struct.
- surface shows up
- then check surface for con_memb
- change isovalue to 1
- then check surface for con_dna
- change isovalue to 1

surfaces are dark but visible.

- uncheck surface for con_dna
- check it again
- SURFACE DOES NOT REAPPEAR

## Environment

chrome on macos, apple silicon

PR #508, opened 2026-04-23
Present each frame for at least 125ms

pgarrison:
Problem
=======
Playback in time is laggier than it could be.

This PR has a [companion in vole-core](https://github.com/allen-cell-animated/vole-core/pull/397): please review that too.

Solution
========
* Require that each frame be presented at least 125ms after the previous
* Increase the maximum number of timesteps that can be pre-fetched to 12 ([companion PR](https://github.com/allen-cell-animated/vole-core/pull/397))
  - Pre-fetching improves total framerate at the cost of lag. The 125ms minimum decreases total framerate. This change offsets that framerate drop.
* Not directly important for performance, but nice to have: remove duplicate `gpuFuse` calls (App/index.tsx)

Analysis
======
I estimate this change decreases lag by about 24% for AICS zarrs on S3, with minimal to no impact on aggregate framerate.

I measure lag as the coefficient of variation of the time between frames. For example, without this change, three frames might be presented for 25ms, 35ms, and 480ms (CV = 1.44). After this change, those three frames would ideally be presented for 125ms, 125ms, and 290ms (CV = 0.53).

In the comparison between my previous PR #507 and this one, median lag decreased from 1.30 to 0.96. This difference was statistically significant (p=0.003, N=9). Median framerate nominally improved, but is absolutely not statistically significant (p=0.98, N=9).

Screenshot 2026-04-21 at 12 32 51 AM

Since you might be wondering why this change was necessary, here's a comparison of the time each frame was presented between two runs, one with this PR and one without. Note that without this PR, many frames are presented for less than 125ms. This is because without this PR, frames are requested every 125ms, but there can be a long gap between the request and the presentation.

Screenshot 2026-04-23 at 2 42 30 PM

## Type of change
Performance improvement

Steps to Verify:
----------------

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -327,14 +327,13 @@ const App: React.FC<AppProps> = (props) => {
       // save the channel's new range for remapping next time
       channelRangesRef.current[channelIndex] = [thisChannel.rawMin, thisChannel.rawMax];
 
-      view3d.updateLuts(image);
+      // Note: updateLuts and updateActiveChannels are not called here because the
+      // channelVersion bump (via onChannelDataLoaded) will trigger ChannelUpdater
+      // effects that apply LUT/color settings and call updateLuts with the final state.
       view3d.onVolumeData(image, [channelIndex]);
 
       if (image.channelNames[channelIndex] === maskChannelName) {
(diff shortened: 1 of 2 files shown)


toloudis (approved):
cool, I like removing any calls to updateLuts and updateActiveChannels

ShrimpCryptid (approved):
LGTM!

PR #510, opened 2026-05-01
Styling tweaks for scale level controls

frasercl:
Review time: small (~10-15min)

Implements a number of small styling and behavior improvements requested by @thao-do in a feedback session following #501 merging, or which I noticed in the same session.

- Text in the scale level dropdown should be left-aligned and should be light gray, even while the dropdown is disabled
- A small strip of the scale level selection dropdown triggered incorrect styling when hovered
- Per the [Figma design](https://www.figma.com/design/EJCCnElvRvObznBhN519sl/Vol-E---CFE), tooltips should be dark gray, not black
- The word "Resolution" should be light gray, not white
- The styling of options in the scale level dropdown did not exactly match the adjacent render mode dropdown
- When the toolbar is in "scroll mode" to accommodate narrow screens, the "scroll right" button did not always disappear when the toolbar was scrolled all the way to the right
- Resolution controls appear in the toolbar only after image metadata has been loaded and the image is known to contain multiple scale levels; the toolbar didn't notice when this caused its contents to overflow its horizontal space, so it didn't enter "scroll mode."

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -54,6 +54,7 @@ const defaultVisibleControls: ControlVisibilityFlags = {
   showAxesButton: true,
   showBoundingBoxButton: true,
   metadataViewer: true,
+  scaleLevelControls: true,
 };
 
 const defaultProps: AppProps = {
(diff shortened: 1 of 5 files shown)


rugeli (approved):
lgtm with one question!

interim17 (approved):

PR #511, opened 2026-05-01
Tweak error alert to better surface new errors

frasercl:
Review time: smallish (15-20min)

Since #501, it is now possible to trigger an error alert in the relatively normal course of using the app, by selecting a scale level which cannot fit in memory. This PR applies some feedback from @thao-do for how to better surface errors now that this is the case:
- When an error was caused by selecting a specific scale level, the error message includes the dimensions of that scale level
- When an error appears while the alert is still open, the alert briefly flashes to get the user's attention and jumps to that error's message, rather than hiding the new error behind a "next error" button and staying on the last error that hasn't been dismissed

I also aligned the styling of the error banner more closely with the [Figma design](https://www.figma.com/design/EJCCnElvRvObznBhN519sl/Vol-E---CFE?node-id=1834-4812&t=LEZ7b7IdwAei3ehl-0), which features larger icons and uses an outlined "caution" sign rather than a filled "X."

### Screenshots

image

image

image

This link can be used to test out the new behaviors using the "Resolution" dropdown in the top toolbar &mdash; the largest scale levels of the image are too large: https://allen-cell-animated.github.io/vole-app/pr-preview/pr-511/viewer?url=https%3A%2F%2Fuk1s3.embassy.ebi.ac.uk%2Fidr%2Fzarr%2Fv0.4%2Fidr0048A%2F9846152.zarr

### DIFF
--- a/src/aics-image-viewer/components/App/index.tsx
+++ b/src/aics-image-viewer/components/App/index.tsx
@@ -340,7 +340,7 @@ const App: React.FC<AppProps> = (props) => {
   );
 
   const onError = useCallback(
-    (image: Volume | undefined, error: unknown) => {
+    (error: unknown, image?: Volume) => {
       showError(error, image);
       onImageTitleChange?.(undefined);
     },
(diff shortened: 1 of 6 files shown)


interim17 (approved):
LGTM, I like the new flash behavior

rugeli (approved):

Issue #490, opened 2026-03-06
Rotation gizmo


frasercl:
## Use Case

Requested by @antoineborensztejn: would like to be able to rotate in only one axis at a time

## Solution

Add a rotation gizmo with handles to grab and rotate in one axis at a time

## Alternatives

Some control outside the viewer canvas?

PR #512, opened 2026-05-06
Precise rotation controls

### LINKED ISSUE #497: precise on-axis views in 3d mode
## Use Case

For precise viewing, users in 3d mode want not only to be able to constrain rotation along a single axis, but they also would like "quick" side views along +/- X, Y, and Z axes. (left, right, top, bottom, front and back views)

## Solution

Tool bar buttons?  (see AGAVE?)  or something better.  perhaps clickable axis widget (like double clicking on one of the axes will switch the view)

## Alternatives

frasercl:
Review time: medium-large

Resolves #497 and #490: adds controls for rotating the volume in one axis at a time, and for jumping the camera to different fixed perspectives.

After much deliberation, these controls have been added to a new "tab" in the bottom drawer, which previously only held clipping settings:

image

All the linear algebra for calculating the new position of the camera is done up at this level rather than in `vole-core`, and is therefore done without `three`'s help. A large part of the line count here is a tiny bespoke linalg library for this purpose, in shared/utils/camera.ts.


toloudis (approved):
I tested it and it's looking and feeling great.  The only clunky thing to me is the weird tabs on the bottom drawer.

meganrm (approved):
I am generally always in favor of using a library if possible, so I agree with dan's comment. otherwise seems good

ShrimpCryptid (approved):

PR #524, opened 2026-05-20
Fix channel settings not appearing when channel name is the empty string

frasercl:
Review time: tiny (>5min)

Fixes a bug where a channel settings row does not render for channels whose name in image metadata is the empty string.

### Steps to verify

- Go to https://vole.allencell.org and open https://ome-zarr-scivis.s3.us-east-1.amazonaws.com/v0.5/96x2/pig_heart.ome.zarr/scale4/pig_heart. There should be no channel settings!
- Open the same image at the PR preview link below. The channel settings should appear properly, apparently missing a channel name.


ShrimpCryptid (approved):

toloudis (approved):
LGTM, thanks for the quick fix!

PR #525, opened 2026-05-22
Refactor: move custom styling for `Select` and `Radio` components into `StyleProvider`

frasercl:
Review time: medium (~25min)

Currently, the `Select` (dropdown) and `Radio` (radio button group) components from `antd` are only meaningfully used in the toolbar. As a result, we've ended up with quite a lot of styling for these components down in `Toolbar`'s stylesheet which in principle should apply to these components no matter where they appear in the app.

Future work will add instances of both components to the control panel. So rather than have to redo all of `Toolbar`'s bespoke styling in the control panel, I took the opportunity to update, polish, and reorganize a bunch of styling up in `StyleProvider`, where app-wide styling lives.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "3.3.0",
       "license": "ISC",
       "dependencies": {
-        "@aics/vole-core": "^4.9.0",
+        "@aics/vole-core": "^4.9.1",
         "@ant-design/icons": "^5.2.5",
         "@fortawesome/fontawesome-svg-core": "^6.5.2",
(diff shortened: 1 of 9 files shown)


ShrimpCryptid (commented):
Image

Outside the scope of this PR, but might need to tweak style overrides on the cookie banner

ShrimpCryptid (approved):
Clicked around in the preview and I love that all our dropdowns have consistent styling now. Left a few comments but overall love that we're deleting a bunch of CSS!

toloudis (approved):
Nice clean change set.  Please check the copilot comments and see if they are valid, it seems a couple of them might be.

Issue #222, opened 2024-04-19
Changing ROI slider causes others to reset


ShrimpCryptid:
## Description

https://github.com/allen-cell-animated/website-3d-cell-viewer/assets/30200665/df726bc3-06e3-453e-84f0-c6bb3fe4947a

Changing one of the ROI sliders causes the others to reset to near the endpoints of the range. Is this intentional behavior?

## Expected Behavior

The ROI slider should not change unless the user is interacting with it.

## Reproduction

1. Open https://allen-cell-animated.github.io/website-3d-cell-viewer-release/
2. Change one slider
3. Change another slider

## Environment

Chrome on Windows 10

PR #529, opened 2026-06-02
polish tooltip styles and fix control panel accessibility

lynwilhelm:
Purpose
=======
Improve tooltip visual polish across the viewer and fix accessibility gaps on buttons that switch the panel views

Problem
=======
[Issue ticket](https://github.com/allen-cell-animated/vole-app/issues/528)

Solution
========
Improved styling and adding tooltips to control panel buttons

Change summary:
---------------
**ControlPanel**
- Added tooltip ("Show panel" / "Hide panel") and `aria-label` to the collapse/expand button
- Added `aria-label` to all icon-only tab buttons
- Fixed tab tooltips being incorrectly suppressed when the panel was expanded

**StyleProvider**
- Lightened tooltip background (`#222` → `#4b4b4b`), reduced font size to 12px and weight to 300, darkened box shadow
- Fixed multi-line tooltip layout using `flex-direction: column`
- Added `GlobalTooltipStyle` for portal-rendered tooltip overrides

## Type of change

Architectural decisions
=======
Tooltip styling is split across two mechanisms by necessity:
- ConfigProvider component tokens (colorBgSpotlight, fontSize) — used where antd's token system supports the property directly
- GlobalTooltipStyle (createGlobalStyle) — used for font-weight, box-shadow, and layout, because antd renders tooltips into portals on document.body, outside the scoped CssProvider div. A comment in the ConfigProvider block cross-references GlobalTooltipStyle so the two halves stay linked for future engineers.

Tradeoffs
=======
- !important on box-shadow is required to override antd's CSS-in-JS as well as centering the text, which wins on specificity for portal-rendered elements. This is documented inline.
- GlobalTooltipStyle applies uniformly to all tooltips in the app. Any future tooltip needing divergent styling will need to explicitly override these rules.

Testing performed
=======
- Tooltips appear on tab icons in both collapsed and expanded panel states
- Collapse/expand button tooltip and aria-label update correctly on toggle
- Single channel mode tooltip renders its two lines stacked correctly
- Box shadow, font size, font weight, and background color verified across right- and bottom-placement tooltips
- Ran all prompts Sean gives in [these notes](https://docs.google.com/document/d/1XBQpF_5bisdpwZjJn8at7bNajJf7m-dZw9qj-JGwcoE/edit?tab=t.0) for best practices

Steps to Verify:
----------------
Hover over anything that has a tooltip, and check far left panel buttons.

Screenshots (optional):
-----------------------
Before:
Screenshot 2026-06-02 at 1 49 31 PM
After:
Screenshot 2026-06-02 at 1 49 57 PM
Screenshot 2026-06-02 at 1 50 12 PM
Screenshot 2026-06-02 at 1 50 18 PM

-----------------------
2 files (see below)

Thanks for contributing!


ShrimpCryptid (approved):
Definitely an improvement! Left an optional comment about font weight/sizing.

toloudis (approved):

Issue #354, opened 2025-01-18
copy settings and paste into new window


toloudis:
## Use Case

Antoine has two windows open with different volumes.  He wants to copy the settings from one window and paste them all into the second window.

@antoineborensztejn 

## Solution

unknown, but some kind of clipboard interaction would make this very seamless.
example: a tool button, perhaps even in the Share dialog, to copy all settings, and then a button to allow pasting in the saved settings.  kbd shortcuts would be interesting too.

jennifermccarty:
Dan - scoping.

ShrimpCryptid:
Note that this may also require #414 to be implemented, since otherwise we would be pasting bin values instead of raw data values.

PR #534, opened 2026-06-17
Reorganize state (de)serialization

frasercl:
Review time: large or small, depending on how in-depth you feel like going on a PR that mostly just pushes code around.

## Problem

Solves two distinct problems with the current state serialization code:
1. Currently, all functions and types related to viewer/channel state (de)serialization and URL (un)parsing is packed into one ~1.2k file, at shared/utils/urlParsing.ts. This file is tough to browse and understand.
2. urlParsing.ts is oriented specifically around (de)serializing to/from a URL. With #354 on the horizon, we're set to need (de)serialization to/from other contexts.

Fortunately, urlParsing is built around some intermediate types, e.g. `ViewerStateParams`, which contain the same information as `viewerState` but with condensed keys and safely stringified values. Parsing a URL is done in two steps: first, URL parameters are parsed into instances of these types, then these types are validated and converted into the actual state record. This suggests a possible improvement.

## Solution

Split the code currently contained in urlParsing.ts between four files:
1. The serialized parameter types from `urlParsing` move into the existing state/types.ts file
2. Code for serializing to/deserializing from these parameter types move into state/serialize.ts and state/deserialize.ts, respectively
3. urlParsing.ts retains code related specifically to reading and writing URL parameters.

urlParsing's large unit tests module is split in three to follow the three of the four above modules that actually contain functional code.

The underlying idea here is: `ViewerStateParams` and its family of types are a *format-agnostic alternate representation of `ViewerState`/`ChannelState`/etc.*, and therefore should live in the directory responsible for managing viewer state. These types are primarily optimized for packing into a URL, but can be used in other contexts where the primary need is compact, consistent, and reliable state transfer.

## More problems and possible future work

There's at least one major point of friction with this new organization: channel settings currently deserialize into a `ViewerChannelSetting` (the type used to describe the initial settings for channels in `App` props), not a `ChannelState` (the type that's actually stored in state). This breaks the ideal contract I described above: currently, the channel params type is not *actually* a conveniently convertible alternate representation of a `ChannelState`. Plus, `ViewerChannelSetting` requires a channel index, which inserts a bit of unwanted context-dependency where it's not always useful (e.g. serializing the state of a single channel).

In a future PR, I'd like to make the default deserialization of channel states convert right to a `ChannelState` object, with an optional mapping from there to a `ViewerChannelSetting` for use cases where that's preferable. Building to that could be a great jumping-off point to wrangling the scattered fragments of code that handle `ViewerChannelSettings` into a consistent place.

There's also some other state-related code that hasn't migrated into the state/ directory yet, notably the default viewer state factory function.


ShrimpCryptid (approved):
LGTM! I wish GitHub had an easier tool for checking that whole blocks of code were copied unchanged between files (or to diff a section between files...)

meganrm (approved):

PR #535, opened 2026-07-06
Deserialize state to `ChannelSetting`

frasercl:
Review time: small (~15min)

Part 2 of the state serialization/deserialization refactor begun in #534. That first PR mostly pushed existing serialization code into new modules. In its description I said:

> There's at least one major point of friction with this new organization: channel settings currently deserialize into a `ViewerChannelSetting` (the type used to describe the initial settings for channels in App props), not a `ChannelState` (the type that's actually stored in state).
> ...
> In a future PR, I'd like to make the default deserialization of channel states convert right to a `ChannelState` object, with an optional mapping from there to a `ViewerChannelSetting` for use cases where that's preferable.

This follow-up **adds the ability to read a serialized `ChannelState` directly back into a `ChannelState`**, though it doesn't quite get all the way to treating that as the default.


rugeli (approved):
Overall looks good to me! left a few nits and non-blocking questions

ShrimpCryptid (approved):

PR #540, opened 2026-07-14
Import/export channel settings to/from file (copy & paste 2)

### LINKED ISSUE #536: Copy and paste all channel settings between windows
Implementation ticket for #354. Add ability to copy and paste (to/from clipboard) and import and export (to/from a file) *all* channel settings in a volume. Channel settings are matched up by name.

Design here: https://www.figma.com/design/EJCCnElvRvObznBhN519sl/Vol-E---CFE?node-id=2031-8295&t=nSIJKqcSGR0z2GEC-0

frasercl:
Review time: medium (~25-30min)

Implements the remaining features to close #536:

- Export channel settings to a .json file
- Import those settings back into Vol-E through a special modal
- Option to exclude color when copying or exporting

## Screenshots

image
image
image

### DIFF
--- a/src/aics-image-viewer/components/ControlPanel/CopySettingsButton.tsx
+++ b/src/aics-image-viewer/components/ControlPanel/CopySettingsButton.tsx
@@ -258,8 +258,13 @@ const CopySettingsButton: React.FC<CopySettingsButtonProps> = (props) => {
         overlayStyle={{ minWidth: 100 }}
         getPopupContainer={getDropdownContainer}
         open={dropdownOpen}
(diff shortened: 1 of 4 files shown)


toloudis (approved):
my comments are mostly about trying to understand the code. 
1. if the clipboard format is the same as the import/export format, then I would consider not using the word clipboard to describe it.  if it's not really the same thing, then my comments around this are irrelevant
2. i think some of the code in CopySettingsButton could be moved out into a more functional module such as parseClipboard.ts

meganrm (approved):

PR #541, opened 2026-07-15
Upgrade @types/three to 0.184.0 and pin three exactly

pgarrison:
Problem
=======
When building `vole-app` with a linked local `vole-core`, there is a mismatch in the versions of `@types/three` that results in the following error upon an `npm install` from `vole-app`:

```
src/aics-image-viewer/components/useVolume.ts(298,9): error TS2739: Type 'Box3' is missing the following properties from type 'Box3': toJSON, fromJSON
```

Solution
========
Update `@types/three` to `0.184.0` from `0.171.0`

## Type of change

Change summary:
---------------
In addition, I made some updates how the three.js dependency is handled.
* `three` is now an explicit dependency of `vole-app`.
  * It was previously not listed in `package.json`, despite `useVolume.ts` doing an explicit import from `three`.
  * `vole-core` lists `three` as a peer dependency, meaning that the consumer (`vole-app`) should provide it.
* `three` and `@types/three` are pinned to an exact version, instead of with `^`.
  * In order to keep the `package-lock.json` files in sync between `vole-app` and `vole-core`, I recommend pinning these dependencies exactly. `0.184.1` has been released, but `vole-app` shouldn't update to it yet without `vole-core`.

Steps to Verify:
----------------
1. From `vole-core`, do `npm i` and `npm link`
2. In `vole-app`, do `npm link vole-core`
3. In `vole-app`, do `npm i`: it should not error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "react-color": "^2.19.3",
         "react-router-dom": "^6.26.2",
         "styled-components": "^6.1.13",
+        "three": "0.184.0",
         "usehooks-ts": "^3.1.0",
         "uuid": "^13.0.0",
         "zustand": "^5.0.6"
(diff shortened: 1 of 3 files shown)


frasercl (approved):
This should no longer be necessary once allen-cell-animated/vole-core#405 lands on main, but it's waiting on a couple of other breaking changes to catch up and might still be a bit out.

ShrimpCryptid (approved):

PR #556, opened 2026-08-07
Reorg enum types

frasercl:
Makes a few reorgs to types that are small but touch a lot of files:
- Remove shared/enums.ts. These are all types that are closely associated with state, so they're moved to state/types.ts.
- Converts state parameter types to types from classes. These types are never instantiated with their defaults (`undefined` for all properties) and we can still generate a list of their keys from the enums.
- Renames the type `PerAxis` to `XYZ`.


src/aics-image-viewer/shared/utils/parseUrl.ts
```diff
@@ -87,7 +88,7 @@ type ViewerMessage = {
 };
 
 const allowedParamKeys: Array<keyof AppParams> = [
-  ...Object.keys(new ViewerStateParams()),
+  ...Object.values(ViewerStateKeys),
```
ShrimpCryptid: This use case prevents the enums from being inlined as `const`, right? I remember we had had a discussion a while back about `const` vs non-`const` enums in a TFE PR.

ShrimpCryptid (approved):
LGTM!

src/aics-image-viewer/shared/utils/parseUrl.ts
frasercl: Yeah, I'm still not a huge fan of the default non-`const` TypeScript enum, but this feels like a case where it's useful to take advantage of making the enum keys into actual items in the built code.

